PR #15, opened 2019-08-31
Do not print url, before make public

zenoengine:


lanctot:
Hi @zeroengine, apologies this has taken so long! It's the next one I plan to merge, but we have made a number of changes that affect the go's playthrough (and I have another one coming) so you will need to pull and regenerate the playthough again. I'll follow-up when this other commit is made.

lanctot:
Ok @zeroengine, can you pull and generate the go playthough again?

zenoengine:
> Ok @zenoengine, can you pull and generate the go playthough again?

Yes, There's no problem. I create new PR, please check #52 .

Issue #25, opened 2019-09-04
terminal state detection connect_four not working correctly


danielwillemsen:
It seems that the terminal state detection of connect_four does not detect all winning states. 
See the following example for a minimal reproduction:
```
import pyspiel
game = pyspiel.load_game("connect_four")
state = game.new_initial_state()
state.apply_action(3)
state.apply_action(2)
state.apply_action(3)
state.apply_action(2)
state.apply_action(3)
state.apply_action(2)
state.apply_action(3)
print(state)
print(state.is_terminal())
```

results in:
```
.......
.......
...x...
..ox...
..ox...
..ox...

False

```

This should be a winning state for x and thus print True, right? Or do I just understand the functionalities of state and is_terminal() incorrectly?

OpenSpiel:
Fixed thanks to your PR! (https://github.com/deepmind/open_spiel/pull/26)

Thanks again, it is really helpful.

PR #26, opened 2019-09-04
fixed terminal condition in connect_four

danielwillemsen:
There was an error in the terminal condition check in the C++ implementation of connect four. 

Short explanation: You only need to check 3 cells to the side of your initial checking point, to reach a connected length of 4, as the initial cell is also part of the connecting chain. This was not correctly implemented in the out-of-range checking before the actual connection check. Resulting behaviour was missing terminal state along edges.

This solves #25


danielwillemsen:
@googlebot I signed it!

PR #31, opened 2019-09-05
fix install osx

kousun12:
Fresh install fails, seems like `brew install gcc@7` was unintentionally removed from the install script.

```
$ ./open_spiel/scripts/build_and_run_tests.sh

+ CXX=g++
+ [[ darwin18 == \d\a\r\w\i\n* ]]
+ CXX=/usr/local/bin/g++-7
++ sysctl -n hw.physicalcpu
CMake Error at CMakeLists.txt:4 (project):
  The CMAKE_CXX_COMPILER:

    /usr/local/bin/g++-7

  is not a full path to an existing compiler tool.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,9 @@ build/
 open_spiel/abseil-cpp/
 open_spiel/games/bridge/double_dummy_solver/
 pybind11/
+
+# Install artifacts
+get-pip.py
+
+# IDE
+.idea/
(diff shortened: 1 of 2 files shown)


kousun12:
@googlebot I signed it!

OpenSpiel:
Thanks. If you have the courage, it should be possible to also have a Travis Test for MacOS, which could check reliably that it compiles and run under a fresh MacOs install.

PR #33, opened 2019-09-06
Fix intro.md

zenoengine:
:smile:


lanctot:
Hi @zenoengine , thanks. This fix will be better to fix internally because we have to remap the directory structures for the internal copy and external copy. So but we will make reference to this PR when we fix it. Thanks again!

PR #38, opened 2019-09-09
Add dependencies as submodules.

jblespiau:


lanctot:
Nice!

jblespiau:
Rerun travis

jblespiau:
Because of the technicalities involved in exporting the code from within Google/DeepMind to Github (in particular the fact the Cl version of sub-modules is hidden into the Git metadata), we will be continuing to use the git-clone commands. The cost involved of maintaining submodules is too high compared to the very small usability gains.

PR #40, opened 2019-09-09
added information state

Gregory-Eales:


Gregory-Eales:
@googlebot I signed it!

lanctot:
Hi Gregory, thanks!

The InformationState should be perfect recall. In its case we should return HistoryString (inherited from open_spiel::State). See https://github.com/deepmind/open_spiel/blob/26349aeb2c09cbe8f876678007f9543bf51aa28a/open_spiel/spiel.h#L325

Also we need to modify the game type. 

Can you change your implementation in the following way:
- keep InformationState but return HistoryString() instead
- add Observation and return ToString() in it (as you have now in InformationState)
- modify the GameType at the top of go.cc. We need to set provides_information_state and provides_observation both to true?

lanctot:
Awesome, thanks! This is great. 

I am still catching up from being on holiday, so I don't know yet how to merge the PRs, but I will ~delegate~ suggest it to some of my colleagues to be merged until I'm up to speed on the process.

@jblespiau @locked-deepmind

Gregory-Eales:
It looks like I had to change the go.txt playthrough file in order for it to pass the test

lanctot:
Hi Gregory, the playthrough test is failing because it needs to be updated. Can you generate a new one? (See step 9 on https://github.com/deepmind/open_spiel/blob/master/docs/developer_guide.md)

You can verify that it worked by running the tests with `ctest -j$(nproc)` after a successful build.

PR #43, opened 2019-09-10
Add gradient clipping to the Policy Gradient agent

dvoram:
Without gradient clipping the PolicyGradient agent was unstable, occasionally resulting in NaN probabilities.

I added max_gradient_norm argument to the __init__ function of the PolicyGradient agent with the default value of 10.0. If set to None, gradient clipping is disabled.

The gradient clipping is implemented using [tf.clip_by_norm](https://www.tensorflow.org/api_docs/python/tf/clip_by_norm) operation.

### DIFF
--- a/open_spiel/python/algorithms/policy_gradient.py
+++ b/open_spiel/python/algorithms/policy_gradient.py
@@ -104,8 +104,7 @@ class PolicyGradient(rl_agent.AbstractAgent):
                entropy_cost=0.01,
                num_critic_before_pi=8,
                additional_discount_factor=1.0,
-               max_gradient_norm=10.0
-               ):
+               max_global_gradient_norm=None):
     """Initialize the PolicyGradient agent.
 
     Args:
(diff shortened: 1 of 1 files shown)


dvoram:
@googlebot I signed it!

fzvinicius:
Thanks for the PR, @dvoram. A few remarks:

- Some parameter settings are expected to induce bad or Nan gradients. Which game and params did you use? 
- Note that grad clipping was not needed in the reference papers (e.g. RPG paper) which policy gradient was used. That being said, your contribution is welcome, but please set the default to not use grad clip.
- Please use [clip_by_global_norm](https://www.tensorflow.org/api_docs/python/tf/clip_by_global_norm) instead. More details on the paper [On the difficulty of training Recurrent Neural Networks
](https://arxiv.org/abs/1211.5063).

lanctot:
@dvoram , did you see that last message?

dvoram:
Thank you for the comments. Sorry for that delay, I missed the message.

>     * Some parameter settings are expected to induce bad or Nan gradients. Which game and params did you use?

I actually used my own environment implementing a variation of the [Guessing game](https://gym.openai.com/envs/GuessingGame-v0/). The NaNs were produced even with the default parameters, got worse with learning rates increased.

>     * Note that grad clipping was not needed in the reference papers (e.g. RPG paper) which policy gradient was used. That being said, your contribution is welcome, but please set the default to not use grad clip.

This definitely makes sense. I will set the parameter to None by default.

>     * Please use [clip_by_global_norm](https://www.tensorflow.org/api_docs/python/tf/clip_by_global_norm) instead. More details on the paper [On the difficulty of training Recurrent Neural Networks
>       ](https://arxiv.org/abs/1211.5063).

Ok.

Issue #46, opened 2019-09-11
Suggestion: Player instead of int


michalsustr:
For similar reasons why there is `typedef int64_t Action` I'd like to suggest to use `typedef int Player` as well.

lanctot:
We actually already have it as an enum (which makes it compatible with int): https://github.com/deepmind/open_spiel/blob/626c762e7811ba0f9e2c6a45aaf6f9436aaafae6/open_spiel/spiel.h#L36

.... but's not consistently used across the code base. It probably should be; at least some of the core API functions should use it instead of int. It's a rather big change to make, though. I'll bring it up with the other team members to see what they think, but overall I think this is a good idea.

locked-deepmind:
Yes, the type exists, but its barely used. I agree it would make things clearer to use it more consistently - PRs welcome!

PR #48, opened 2019-09-11
Use type 'Player' instead of 'int' in as many places as possible.

michalsustr:
Refactor based on #46 

Touches a lot of code. :-)

I found it useful to use a struct for finding out the cases:
```
struct Player {
    int id;

    inline bool operator==(const Player &rhs) const { return id == rhs; }
    inline bool operator==(const int &rhs) const { return id == rhs; }
    inline bool operator!=(const Player &rhs) const { return !(rhs == *this); }
    inline bool operator<(const Player &rhs) const { return id < rhs; }
    inline bool operator<=(const Player &rhs) const { return id <= rhs; }
    inline bool operator>=(const Player &rhs) const { return id >= rhs; }
    inline bool operator<(const int &rhs) const { return id < rhs; }
    inline bool operator<=(const int &rhs) const { return id <= rhs; }
    inline bool operator>=(const int &rhs) const { return id >= rhs; }
    inline bool operator+(const Player &rhs) const { return id + rhs; }
    inline bool operator-(const Player &rhs) const { return id - rhs; }
    inline void operator++(int) { id++; }
    inline void operator++() { id++; }
    friend std::ostream &operator<<(std::ostream &os, const Player &player) {
        os << player;
        return os;
    }
};
```
since the compiler interchanges `Player` and `int` freely ... (unlike for example in Haskell)

### DIFF
--- a/open_spiel/algorithms/deterministic_policy.cc
+++ b/open_spiel/algorithms/deterministic_policy.cc
@@ -20,7 +20,7 @@ namespace open_spiel {
 namespace algorithms {
 
 DeterministicTabularPolicy::DeterministicTabularPolicy(
-    const Game& game, int player,
+    const Game& game, Player player,
     const std::unordered_map<std::string, Action> policy)
     : table_(), player_(player) {
   CreateTable(game, player);
@@ -32,7 +32,7 @@ DeterministicTabularPolicy::DeterministicTabularPolicy(
 }
 
 DeterministicTabularPolicy::DeterministicTabularPolicy(const Game& game,
-                                                       int player)
+                                                       Player player)
     : table_(), player_(player) {
   CreateTable(game, player);
 }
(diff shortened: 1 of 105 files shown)


michalsustr:
@googlebot I fixed it.

lanctot:
Wow, this is amazing, thanks!

Just a quick heads-up: our travis testing mechanism is currently broken and we're in the process of fixing it. We will take a look once that's done (give it a day or two, though).

lanctot:
Quick question since it is difficult to find on my phone: did you remove that custom Player struct?

I would guess not because the code shows Player rather than PlayerId. I would rather keep using the enum PlayerId since it is already an int already (and we don't need a custom type and overriding methods) and we are already using the PlayerId values in other places.

lanctot:
No, thanks. I prefer the enum for simplicity. Kind of the same argument as the other thread. I do not like introducing custom types for things that are simple (plain old data), just makes everything unnecessarily complicated... the benefits are not worth what you pay IMO.

michalsustr:
Unfortunately we can't have `LegalActions(PlayerId player)`, because a call `LegalActions(0)` will not compile: https://repl.it/repls/SupportiveMonstrousCoins , so there needs to be a `Player` typedef.

lanctot:
Ok, instead of typedef, can you add `using Player = int;` here: https://github.com/deepmind/open_spiel/blob/9a0848880400eb26e481ddfdc558b1b3d075e154/open_spiel/spiel_utils.h#L85 ?

lanctot:
Ok @michalsustr , I've started to merge this one. Should be available in our next push to github. Thanks!

michalsustr:
Oh, ok :) I wanted to do the Player{x} -> x sometime in the evening, but that should be a quick regex search & replace. The using is there already.

PR #53, opened 2019-09-15
Swift TicTacToe maximum number of players is 2

jackpal:
This matches the Swift implementation limits, and also matches the C++ version of TicTacToe's GameInfo.


lanctot:
Thanks, looks good! @danzheng @saeta?

Issue #47, opened 2019-09-11
`kStandard` strategy averaging incorrect in `ExternalSamplingMCCFRSolver`


dmorrill10:
https://github.com/deepmind/open_spiel/blob/ee3b6d906e982e3e62b11bf239a98340b172b9a8/open_spiel/algorithms/external_sampling_mccfr.cc#L108

The behavioral policy is being added to the cumulative policy when the player's sequence probability (the probability that the player plays to the current information state and plays action `aidx`) should be added instead. The probability that the node was sampled will also have to be taken into account by dividing by that probability.

It doesn't look like there's a trivial fix for this because the reach and sampling probabilities aren't being passed down the tree, but adding and keeping track of these values would make the fix easy.

lanctot:
The sequence probs cancel when you do regret updates on the opponent nodes in 2-player games, leaving only the policy probability for the (infostate, action pair) and most (all?) pseudo-code thst I know presents it this way, which is why I called it standard.

I guess you are referring to the (n>2) case? I kept it as is to reflect what Noam did in Pluribus, and that's why I provided the kFull average type to do it correctly (also explained in the header comments).

Make sense?

dmorrill10:
Thanks, @lanctot, you're right. Sorry about that. I forgot that the weighting canceled with the sampling probability and that the chance probabilities can be ignored since they're constant. When there are more than two players, since that extra player's policy might be changing, I think you should be dividing by the probability that player sampled the actions leading to the information state, no? This is a different issue from the one pointed out in the header.

lanctot:
Yes, true, that's also broken. We should add that to the comments in the header as well and maybe kStandard is not a good word choice (maybe kSimple is better). 

I also should have mention the matching to Pluribus in the n-player case for that averaging type.

Really the only "right" one is kFull. In my chats with Rich Gibson he told me that is what he always did.

I put standard/simple because Noam said it worked fine/better and in the few quick tests I ran it seemed to be better than full.

Feel like submitting a PR? If not I will make these changes.

dmorrill10:
You could maybe call it kPluribus and later a player sampling probability fixed version could be added kStochastic. But kSimple is fine too.

Okay, that makes sense, thanks.

You can go ahead and make the changes, thanks.

PR #56, opened 2019-09-15
Laser tag

alexminnaar:
I have added the laser tag game that was mentioned in the contributing doc.  It is implemented similar to how the markov soccer was implemented.  The thing I am unsure of is how observations should be handled here.  Right now there aren't any.  Let me know if that should change.


locked-deepmind:
This is great - thank you!
Could you add a playthrough for the game?

alexminnaar:
@locked-deepmind oops, completely missed that.  Yes I will add that.

alexminnaar:
@locked-deepmind added the playthrough

lanctot:
This is great, thanks!

Regarding the observation: I saw that you had InformationState but that should be Observation (and you could leave out InformationState). Both for the string and normalized vectors.

I know you copied it from MarkovSoccer and we will have to fix that eventually too.

So can you rename InformationState to Observation? You will need to change the GameType values and regenerate the playthrough.

alexminnaar:
@lanctot it looks like there is a problem with  Test #47: turn_based_simultaneous_game_test.

lanctot:
Hmm, Travis is cutting off the traces at cooperative box-pushing, probably because every game has a lot of simulations and is printing them all. Can you run `games/turn_based_simultaneous_test` manually? Does it fail on your machine? If so, is it clear why? If it fails and is not clear, can you copy & paste the last 10-20 lines of the test here?

Note that we just added a requirement to the API that actions need to be sorted in the LegalActions() a few days ago. If your branch is not up-to-date it could be why it might fail in Travis but not for you locally. If so, you may have to sort the actions, as chess:ChessState::LegalActions does (see here: https://github.com/deepmind/open_spiel/blob/316a71f083afdc59fabb2609faa1e6f3c542ed4b/open_spiel/games/chess.cc#L109)

On a different note, can you tell me which description of the Laser Tag environment you based this on? Was it the one from the [original SSD paper](https://arxiv.org/abs/1702.03037)? The header still shows MarkovSoccer, so we should change that. But also: that environment was not zero-sum, so we should probably support both the original one as described in the paper, and also a zero-sum one that you have implemented (we can make this a game parameter).

alexminnaar:
@lanctot ah I didn't realize that ```turn_based_simultaneous_game_test``` was calling ```laser_tag```.  It looks like the problem was that I had renamed the ```InformationState``` method to ```Observation``` however it inherits ```SimMoveState``` so it needs to implement ```InformationState```.  So I changed that back, but then it created an error generating playthroughs because ```laser_tag``` had ```/*provides_observation=*/true```.  So I set ```/*provides_observation=*/false``` so the tests would pass.  So I'm not sure if those ```GameType``` fields are correct, right now they are

```
/*provides_information_state=*/false,
/*provides_information_state_as_normalized_vector=*/false,
/*provides_observation=*/false,
/*provides_observation_as_normalized_vector=*/true,
```

For the other note, I tried basing the laser tag environment off of [this paper](https://arxiv.org/pdf/1711.00832.pdf).  Also, would the other option be kGeneralSum in addition to kZeroSum?

lanctot:
Oh.. I had not thought of the SimMoveState requiring InformationState; that is a bit of a problem due to some old assumptions that no longer hold. We should exclude the games that don't provide InformationState from that test. 

Ok right, that's the PSRO paper -- we used the same environment as in Leibo et al., but it was not zero-sum. You're right that the game type would need to be set appropriately in each case. 

At this point, there's probably a number of changes that might be easier to do on our end. If you're happy with its current state, I can import it, and do a few minor changes, so that we can include it in our next update. What do you think?

By the way, this is our first external game contribution! And, it was something someone asked me for on the day of the release. So thanks again for doing this, I'll also add it to games.md and in the next paper update too.

alexminnaar:
@lanctot I'm fine with that plan.  No problem, happy to contribute :)

lanctot:
Ok, I've imported this one and fixed the problem with the tests. I also added a few capabilities to more closely match the envs from the paper (option for nonzero-sum returns and for extended episodes of 1000 steps with multiple tags). Should go in on the next update, and it'll be quite nice to update the paper to add this our list of games!

PR #57, opened 2019-09-15
fix build directory name

alexminnaar:
I think the build script actually appends the python version to the build directory name here https://github.com/deepmind/open_spiel/blob/9a0848880400eb26e481ddfdc558b1b3d075e154/open_spiel/scripts/build_and_run_tests.sh#L44 so if we are assuming python 3 I think the build directory would actually be called "build_python_3".  This is what happened when I built it.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -32,7 +32,7 @@
     # For the python modules in open_spiel.
     export PYTHONPATH=$PYTHONPATH:/<path_to_open_spiel>
     # For the Python bindings of Pyspiel
-    export PYTHONPATH=$PYTHONPATH:/<path_to_open_spiel>/build/python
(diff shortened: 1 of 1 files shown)


lanctot:
Hi @alexminnaar , yeah - I think we tagged on the name because we were supporting Python 2 and 3 at some point. I agree -- we should probably just remove this. @jblespiau @locked-deepmind, can you think if this will break anything and let me know what you think?

locked-deepmind:
Seems safe to me. There are probably some other py2 vestigial pieces too.

jblespiau:
Thanks. I made the change of the directory name to be build/ internally, and it will be pushed with the next release. I am thus closing this PR, as it will fix the directory name in install, which will now match the directory name in the instructions.

PR #60, opened 2019-09-17
Add import numpy as np in concept.md

srijanreddy98:
In line 87 of concepts.md np is used but not imported. This just adds that import statement.
```
action = np.random.choice(action_list, p=prob_list)
```

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -75,6 +75,7 @@ Here are for example the Python code to play one trajectory:
 ```python
 import random
 import pyspiel
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


lanctot:
Good catch!

lanctot:
We have integrated this PR internally, so this will go out on the next push to github. Thanks!

srijanreddy98:
Happy to help.

lanctot:
Oh there was need to close it. Our push to github does that and associates the commits with it, so I like to keep it up until we do the push.

PR #66, opened 2019-09-22
Backgammon narrative

jamesdfrost:
Changes to use Backgammon standard narrative formats for human readable notation, as follows:

- Always show the numbering going from Bar->24->0->Off, irrespective of which player is moving.
- Show the start position followed by end position, instead of start position, number of moves. E.g. 24/22 10/8
- Show hits with an asterisk, e.g. 9/7*.
- Order the moves by highest number first, e.g. 22/7 10/8 not 10/8 22/7. Not an official requirement, but seems to be standard convention.
- Show duplicate moves as 10/8(2) instead of 10/8 10/8.
- Show moves on a single piece as 10/8/5 not 10/8 8/5

Also introduced tests to check for the above scenarios.

Files changed - backgammon.cc, backgammon_test.cc and backgammon.txt (for the playthrough tests)


lanctot:
Hi @jamesdfrost , this sounds great from the description. Unfortunately, though, the PR seems to have similiar issues that you were having in the other one. It's showing 6 commits, including windows.md. I'd like to keep the PRs separate so that the diffs represent only the changes that were made. 

I think this is because you're probably re-using the same original branch you made for that first PR (or the new branch has the same commits.) For now, rather than figuring out how to do the branching properly on the same repos, could you try entirely from scratch: for each PR, re-clone the repos (to a different directly), create the PR, and switch to that branch only in that directory. Then copy over the backgammon.cc and backgammon_test.cc that you have and create just the one commit on that branch. 

This way you could have several different branches checked out and not interacting with each other (because of them is cloned to a different directory)

PR #72, opened 2019-09-24
Backgammon Notation

jamesdfrost:
Changes to use Backgammon standard narrative formats for human readable notation, as follows:

- Always show the numbering going from Bar->24->0->Off, irrespective of which player is moving.
- Show the start position followed by end position, instead of start position, number of moves. E.g. 24/22 10/8
- Show hits with an asterisk, e.g. 9/7*.
- Order the moves by highest number first, e.g. 22/7 10/8 not 10/8 22/7. Not an official requirement, but seems to be standard convention.
- Show duplicate moves as 10/8(2) instead of 10/8 10/8.
- Show moves on a single piece as 10/8/5 not 10/8 8/5
- Also introduced tests to check for the above scenarios.

Files changed - backgammon.cc, backgammon_test.cc and backgammon.txt (for the playthrough tests)


lanctot:
Ok, @jamesdfrost , I've started to import it. Should go in on the next update.

PR #77, opened 2019-09-26
bugfix in connect_four

danielwillemsen:
If I understand the meaning of NumDistinctActions correctly, this should be the number of columns for connect four.


lanctot:
Good catch! Thanks, we will merge it in the next update.

PR #87, opened 2019-10-03
use C++17 for external linkage of inline functions

juliendehos:
`GameParametersToString` is declared inline in `open_spiel/game_parameters.cc`. This doesn't compile with C++11 and requires C++17.

### DIFF
--- a/open_spiel/CMakeLists.txt
+++ b/open_spiel/CMakeLists.txt
@@ -8,7 +8,7 @@ Any Python library found with CMake modules FindPython2/FindPython3 that \
 agrees  with the major version and is at least as high for minor version and \
 patch number is accepted. If the variable is not set, we use the FindPython \
 module which favours Python 3 over Python 2 if both are available.")
-set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD 17)
 if(APPLE)
   set (CMAKE_CXX_FLAGS "-Werror -undefined dynamic_lookup")
 else()


lanctot:
Hi @juliendehos , we do not currently support C++17. But this is great timing, as we just talked about this recently, and we plan to be supporting it as of next week. :) 

It's a fairly big change, have to update docs, check everything still works on the major platforms, etc. we have to do this manually for MacOS since we don't have Travis setup.

So, your contribution is welcome -- but we'll have to add a fair bit to it, and sweep through the code to move some of the absl stuff to the standard C++17 ones.

lanctot:
> `GameParametersToString` is declared inline in `open_spiel/game_parameters.cc`. This doesn't compile with C++11 and requires C++17.

Odd.. seems to work on our end. And the travis builds are passing. But ok, we'll keep this in mind.

juliendehos:
Ok.
Open_spiel seems to compile with gcc>=8 but with gcc7, I had to set the C++17 option or to remove the `inline` keyword in `open_spiel/game_parameters.cc` https://github.com/deepmind/open_spiel/blob/6b33c5a3aec8251a238e1e994b0c3fa72a3b780e/open_spiel/game_parameters.cc#L63

lanctot:
Ok, thanks for following-up! Should be fixed next week.

lanctot:
Ok, we have C++17 working internally, will be available on our next update (Monday).

Thanks @finbarrtimbers @jblespiau

lanctot:
C++17 support is in!

juliendehos:
Thanks.
It now compiles successfully with gcc 7, 8 and 9, on my machine.

PR #91, opened 2019-10-09
Laser tag arbitrary grids

alexminnaar:
I wanted to try to implement the second TODO for the laser tag game which is to extend it to arbitrary grids.  My idea was to add 3 new GameParameters

- `grid`:  A string that represents the grid in row-major format.  '.' for empty spaces and '*' for obstacles.
- `num_rows`:  number of rows in the grid.
- `num_cols`: number of columns in the grid.

(technically only one of the last 2 parameters should be required, maybe I should change that.)
In the `LaserTagGame` constructor the obstacle locations are extracted from the grid as well as the grid dimensions.  Also, since the grid is arbitrary, the spawn points can no longer be hardcoded so I implemented a heuristic where 4 equidistant open-space indices are used as the spawn points (Maybe in the future the spawn points should be input variables as well?).  The grid dimensions, obstacles and spawn points are then used within `LaserTagState` via the `parent_game` variable.  I also made some changes to the test to account for the new inputs and also the slight change to the spawn points.


lanctot:
Hey @alexminnaar , this is really cool! Unfortunately, we are a bit hammered at the moment, so it might take a bit longer than the last few PRs before we can merge it.

BTW, there will be an update on the games implentations pushed tomorrow. Should be a relatively easy merge if you pull from master, but may not even be required if there's no conflict.

alexminnaar:
Looks like there were some merge issues, I'll check up on this tonight

lanctot:
Ah, right -- almost certainly caused by this commit: https://github.com/deepmind/open_spiel/commit/9e8d77e1ab184d489d22c790bb457e3655a95d1f#diff-ce3aff990562bd01c87b4f85d1219dbe

If you choose any game in there (i.e. markov_soccer) and look at the diff, I think it should be mostly  clear how to resolve the conflict, but don't hesitate to ask if it's not obvious.

alexminnaar:
Yeah it looks like `parent_game` isn't there anymore but the game is still supplied through the `LaserTagState` constructor so I assume I can just get it from there

lanctot:
You can add it back if you need the LaserTagGame specifically. I did that in a number of games (to avoid casting it every time), see for example here: https://github.com/deepmind/open_spiel/blob/b1b321a57754df76309ed9d7ac3c9b449ed34901/open_spiel/games/negotiation.cc#L255

But: if you only need the general methods that are part of the Game superclass, you can now use the inherited shared_ptr<const Game> game_ field from State.

alexminnaar:
Yep looks like that worked

PR #94, opened 2019-10-16
Issue about the "action_probabilities" in Python implementation of Deep CFR

mzktbyjc2016:
The action probabilities used to get the NashConv/Exploitability should be defined on legal actions. Current version only seems to work well in kuhn poker and the NashConv of deep CFR will report a runtime error when switching to leduc poker.


mzktbyjc2016:
> @googlebot I signed it!

lanctot:
Thank you, we will take a look!

Issue #5, opened 2019-08-28
Cool work. Do you plan to support large games, such as Texas Hold'em?


lihuiknight:
FYI

lihuiknight:
wow, that's great. I plan to use this toolbox and do some experiments.

Writing Texas Hold'em solver in pure C/C++ needs a lot of effort. Looking forward.

lihuiknight:
Yes, Noam Brown implemented Texas Hold'em solver in pure C/C++.  I meet him in IJCAI 2019.

It can imagine there are a lot of codes in his solver.  Reimplementation of such difficult game needs a lot time.

Kailianghu:
Cool work ！The ACPC code ([http://www.computerpokercompetition.org/downloads/code/competition_server/project_acpc_server_v1.0.42.tar.bz2]) is pretty good for poker engine

saeta:
Hey all!

Thank you very much for your interest in OpenSpiel. I lead the [Swift for TensorFlow](https://github.com/tensorflow/swift) project in Google Brain and collaborated with @lanctot and many others across both Brain and DeepMind to help make OpenSpiel. When I saw the recent research results coming out using Texas Hold'em, I built a game implementation for OpenSpiel. Unfortunately, I had some edge cases in the game engine that we discovered late, so we held it back from the initial release. Earlier this week, I fixed it up and so @lanctot has now open sourced it at https://github.com/deepmind/open_spiel/blob/master/swift/Sources/OpenSpiel/TexasHoldem.swift (I also spent some time last night and [optimized the game engine a bit](https://github.com/deepmind/open_spiel/commit/6a99c8bbb523a129c4739c25042be5673f2653b9), dropping the time to play a round from 8.4 usec down to 4.7 usec.) :-) 

I and others will be discussing the design of the Swift OpenSpiel port at the [Swift for TensorFlow Open Design Review](https://twitter.com/bsaeta/status/1166382453897871360) _this Friday (tomorrow)_. Please do join if you're able and ask us questions! (A recording will be made and posted later.) To get the invite, join [`swift@tensorflow.org`](https://groups.google.com/a/tensorflow.org/forum/#!forum/swift).

All the best,
-Brennan

dennisjay:
Hi all!

I have an implementation of a Poker gameengine based on the ACPC-Server with a very similar API to open_spiel.

I think I'll be able to contribute it as a pull request.

Are there any style guidelines and rules for contribution? How can I contact you in case of questions?

[...]

PR #97, opened 2019-10-21
Universal Poker Implementation based on ACPC Server

dennisjay:
Dear OpenSpiel Team,

I read in this comment [https://github.com/deepmind/open_spiel/issues/5] that there is the need of a Poker Implementation that does support large poker games as well. Here it is! A full configurable poker implementation based on the ACPC Server (annual poker competition clib).

Here are some advantages of this lib:
- Fully customizable (players, deck size, boardcard count, holecard count, blinds, ...)
- ACPC Headers are fully abstracted so they do not appear in this lib header files (which otherwise could break namespaces)
- Fast (20% faster than Leduc-Poker Impl for a similar Game) 
- Support for bigger Graphs Like Teaxas Holdem unlimited

Some things that should be cared about (Just as far I know)
- Clearance of Copyright and Licence (sorry don't know much about it).
- API and C++ Docs of Methods
- more automated Tests
- Limit Games don't work right now as good
- Configurable Betting abstraction (now hardcoded to All-In, Bet-Pot, Call/Check and Fold)

Please give me some feedback about the code as I'm very interested how I did.

Best Regards
DJ


dennisjay:
I added the ACPC Licence im my branch

dennisjay:
@googlebot I signed it!

kuzmenkoY:
@googlebot I signed it!

lanctot:
Hi @dennisjay, wow this is incredible!

It will take us a bit of time to look at this. We have to check that licenses are compatible, then look over all the code.

In the meantime, do you know why the checks are failing (including the cla)? Travis CI also seems to show a build error.

kuzmenkoY:
@googlebot I consent.

dennisjay:
@lanctot CLAs are done. Some python unit test are still failing. Will look at it tonight in the evening. Perhaps I can ask somebody to write some c++ docs and automated tests if required.

lanctot:
Hi @dennisjay , how much modification did you have to make to the actual ACPC code? If it's only small changes, then what do you think of forking ACPC adding your changes, and then just have OpenSpiel depend on it? 

This could alleviate any licensing issues because the code would not get bundled with OpenSpiel, and would be downloaded separately. It would also be slightly cleaner and less code to review :)

dennisjay:
@lanctot Yeah we can do that I did only one modification to ACPC to solve some pointer issues when creating the game. But that can be handled another way so that we can use the original source only. Should I download it from this rep then [https://github.com/ethansbrown/acpc] or the tar.gz file from [http://www.computerpokercompetition.org/downloads/code/competition_server/project_acpc_server_v1.0.42.tar.bz2] ?

dennisjay:
Actually found a solution for this:

1. I forked https://github.com/ethansbrown/acpc and compared it with the version provided under http://www.computerpokercompetition.org/downloads/code/competition_server/project_acpc_server_v1.0.42.tar.bz2

2. There were some small differences regarding these two versions and the one on server was newer as I suspect

3. So I commited the new version now to my fork and used this (master) in install.sh 

4. So the source is completly untouched now and i patched acpc_cpp to use the original version

jblespiau:
Thanks for contributing. This has been integrated and will be available on the next push to Github.

I reformatted it, so that it uses the same style as the rest (https://google.github.io/styleguide/cppguide.html).
We also added an implementation for `UniversalPokerState::ActionToString`. We did not check it thoroughly though, given its size.

We made it a conditional dependency, i.e. by default it's not downloaded and linked. One need to set a global variable to ON (in scripts/global_variables.sh). However, it's tested on Travis CI.

I hope this will be useful to the community and that people will step forward to continue to improve and maintain it.

Thank you again.

lanctot:
JB, can we leave these open until the update? They get closed automatically and linked with the commits.

lanctot:
> JB, can we leave these open until the update? They get closed automatically and linked with the commits.

I guess it must not have auto-closed on the update because I re-opening it or because there were conflicts, so I'm closing this now. The code is in.. thanks @jblespiau !

dennisjay:
Thanks @lanctot und @jblespiau for the review and to give me that opportunity. :-)

lanctot:
Thank _you_ for this, it was one of our largest (if not *the* largest) PR to date and a non-trivial amount of work. Thanks for putting up with the back-and-forth. This is a really solid addition to the code and makes it possible to apply these RL algorithm to standard poker games which have been used in the literature!

jblespiau:
@dennisjay Hi. I am trying to use the code, and I had a few questions (it may be easier by email?).

In particular, I was wondering what was wrong in:

uint32_t BettingTree::GetMaxBettingActions() const {
  return IsLimitGame() ? 3 : 4;
}

given that:

ALL_ACTIONS[5] = {ACTION_DEAL, ACTION_FOLD,
                                                ACTION_CHECK_CALL,
                                                ACTION_BET_POT, ACTION_ALL_IN};

Should we remove the ACTION_ALL_IN from there, given it will never be played?

PR #104, opened 2019-10-26
Args fix

alexminnaar:
I think that `nargs='?'` is required here for "tic_tac_toe" to be the default game when no positional arguments are given i.e. `python value_iteration.py`  https://docs.python.org/3/library/argparse.html#default


lanctot:
Nice catch! But, I believe we've switched most of them over to absl flags, so I'll do it for this one too! (Will still integrate it via this PR so no need to close it.)

PR #107, opened 2019-10-31
ConvNet DQN example for ConnectFour

jamesdfrost:
I've implemented a simple ConvNet example which is based on the DQN Algorithm. 

The algorithm is called DQN_Conv.py and has the following differences from the fully connected example:

instead of taking just one array of hidden_layers_sizes it takes the following:
- convolutional_layers_sizes      (e.g. [64,32] )
- kernel_shape                           (e.g. [3,3] )
- fully_connected_layers_sizes   (e.g. [64,16] )

it also takes the state_representation_shape instead of the state_representation_size to ensure the inputs are formatted correctly. (this necessitated a change to rl_environment to return this.

At the moment I've stuck with Sonnet for defining the network, but that seems to have a limitation in the construction of convolutional layers, which must number 2 currently. Will look at this in a future revision.

Have implemented both a feedforward and convolutional version of connect_four so the two networks can be compared. To run, type:

```bash
python3 open_spiel/python/examples/connect_four_dqn_conv.py
python3 open_spiel/python/examples/connect_four_dqn.py
```

### DIFF
--- a/open_spiel/python/algorithms/dqn_conv.py
+++ b/open_spiel/python/algorithms/dqn_conv.py
@@ -1,0 +1,444 @@
+# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""DQN agent implemented in TensorFlow for simple convolutional network."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import random
+import numpy as np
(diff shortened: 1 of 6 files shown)


lanctot:
Great!! We will take a look.

jamesdfrost:
Have added a test to this as well now, dqn_conv_test.py, but not sure how to get this added to the master tests list.

lanctot:
Hi @jamesdfrost , if I understand correctly, you have a completely separate DQN implementation for convolutional networks (`dqn_conv.py`).

Is there a reason to prefer this over just supporting conv nets in the original DQN code? The code should not be _that_ different, right... just how the network is invoked?

lanctot:
Likewise, couldn't we also merge connect_four_dqn.py and connect_four_dqn_conv.py?

jamesdfrost:
@lanctot - makes sense. I'll change the DQN code so you can pass it details of either a convolutional or a feed forward network. The differences are just in the network itself and also how the data is reshaped, otherwise basically identical.

lanctot:
Great! Can I ask that you do the same for the example as well (merge `connect_four_dqn.py` and `connect_four_dqn_conv.py`)? Thanks -- I'll integrate it shortly afterward!

jamesdfrost:
I'm going to resubmit this with the changes above over the Christmas break - have been struggling for time due to work commitments over the last month.

lanctot:
No rush, of course! Thanks for making those changes and your continued contributions. Happy holidays@

PR #109, opened 2019-11-07
removing "fixed grid"

alexminnaar:
I think at this point coin_game, laser_tag, and markov_soccer all have grid/grid-size game parameters so we can remove this part.


jblespiau:
Thanks. I copied the change and it will be in, in the next release. Closing now as the tool will not close this one automatically (as I copy-pasted it and did not applied the PR directly).

lanctot:
Hey @jblespiau: re-opening until we update (btw it does autoclose-- tgst is what I have been doing... I think it uses the commit number(s)).

I plan to do an update tonight or tomorrow.

lanctot:
Nevermind, I am wrong because JB didn't import it the same way we import the rest. So I'm closing this one manually.

PR #110, opened 2019-11-10
Building mac version on travis

dennisjay:
Dear open_spiel team,

I saw you have this issue:
https://github.com/deepmind/open_spiel/issues/84

I have a mac and tested some stuff. The problem is a bug of gcc@7 with xcode11:
https://github.com/Homebrew/homebrew-core/issues/44579

One option is to switch to gcc@9 where the bug is solved. But then I tought maybe it is an option to use clang++ and voila build is working and unit tests are running (so far). So why not using the native clang++?


jblespiau:
Thank you so much!

I was just writing the doc to make sure to document what is supported or not:

****************

### Building and running tests

Make sure that the virtual environment is still activated.

By default, g++ is used on Linux and clang++ is used on MacOs.

#### Using g++ on Linux (default)

Build and run tests (Python 3):

```bash
mkdir build
cd build
CXX=g++ cmake -DPython_TARGET_VERSION=3.6 -DCMAKE_CXX_COMPILER=${CXX} ../open_spiel
make -j$(nproc)
ctest -j$(nproc)
```

The CMake variable `Python_TARGET_VERSION` is used to specify a Python version.
Any Python library found with CMake modules FindPython2/FindPython3 that agrees
with the major version and is at least as high for minor version and patch
number is accepted. If the variable is not set, the FindPython module is used:
it builds for Python 3 if both Python 2 and Python 3 are available. In the two
examples above, CMake will search for Python 2 and accept any version >= 2.7 or
search for Python 3 and accept any version >= 3.6.

One can run an example of a game running (in the `build/` folder):

```bash
./examples/example --game=tic_tac_toe
```

#### Using g++ on MacOs

Uncomment the 2 lines in `build_and_run_tests.sh` and `install.sh`

#### Using clang++ on Linux

This is not actively supported or tested. Feel free to contribute.


****************

and then had a few questions:
- Can you confirm this is correct?
- What do you think of not using the same compiler on Both platforms? Do you think it should be, by default, the same compiler to prevent strange errors? But we a possibility to change the compiler?
- When looking at `build_and_run_tests.sh` I see (after an edit of mine of the commented line:

```
CXX=g++
NPROC=nproc
if [[ "$OSTYPE" == "darwin"* ]]; then  # Mac OSX
  NPROC="sysctl -n hw.physicalcpu"
  # Uncomment the following line if you want to use g++. See also install.sh.
  # CXX=/usr/local/bin/g++-7
fi
```

which let me think that "CXX" is still equal to g++. Are you really sure this is compiling with clang and not g++? (at the beginning of the cmake command, it should list exactly which compiler is being used).

So I am really eager to push this, but I would like you feedback on these few questions, as a subtle bug could be hidden there (if CXX is still equal to g++).

Thanks!

dennisjay:
@jblespiau yeah the code is behaving like described in the documentation.

To your second question: Each plattform has it's default compiler, so good ANSI c++ code - and this repo has a very good quality, will compile with each compiler as good. For example if at some point of time windows should be supported there will araise this question too because vc++ will be the default compiler there.  Additionally you have the unittests, which will detect a logical error directly. So the risk of creating logical error by changing the compiler is IMHO very low. Of course there can be compiler errors (as we have seen) but with CI you will prevent them very good.

To your third question. When installing the MAC CLI Tools they link the gcc command to clang automatically. Of course there is an option to detelct the full path of clang compiler and use this explicit. I can take a look at it if you want.

jblespiau:
Thanks. It is being reviewed and submitted (and will be visible when we export it back to github, probably early next week).

The build was failing on my machine with clang version 8.0.1-3 (in particular because abseil does not compile), so I have kept gcc@7 for now. Given we will have a MacOs Travis CI configuration, it will be easier to review any change toward clang on that platform.

PR #113, opened 2019-11-20
[Swift] Fix latent `inout` argument differentiation bugs.

dan-zheng:
`mutating` functions like `Array.append` and `ActionValueCalculator.loss`
cannot be differentiated because differentiation does not yet support
function applications with `inout` arguments.

These latent bugs were uncovered in https://github.com/apple/swift/pull/28352,
which correctly diagnoses function applications with `inout` arguments.

https://bugs.swift.org/browse/TF-129 tracks `inout` argument support.
In the meantime, disable affected code and tests.

---

Manually tested with [Swift for TensorFlow nightly development toolchains](https://github.com/tensorflow/swift/blob/master/Installation.md#development-snapshots).
Unblocks Swift for TensorFlow CI.

### DIFF
--- a/swift/Sources/OpenSpiel/ExploitabilityDescent.swift
+++ b/swift/Sources/OpenSpiel/ExploitabilityDescent.swift
@@ -215,7 +215,10 @@ public struct ActionValueCalculator<Game: GameProtocol> {
   }
 
   /// The exploitability descent loss for a policy.
-  @differentiable(wrt: policy)
+  // @differentiable(wrt: policy)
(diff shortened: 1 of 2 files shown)


marcrasi:
Why does AD diagnose on these? It looks to me like the inout arguments aren't active so differentiation should work even without inout argument support.

dan-zheng:
> Why does AD diagnose on these? It looks to me like the inout arguments aren't active so differentiation should work even without inout argument support.

Thanks for the catch! It's likely this patch wasn't correct, I'll investigate further.

PR #114, opened 2019-11-21
Fix RCFR reservoir buffer

dmorrill10:
The probability that each candidate is included in the buffer is currently incorrect and this patch corrects the sampling algorithm to fix this problem.


lanctot:
Thanks! I've imported it, should come into the next update. (Please don't close the PR. It will be close automatically by the update on Monday.)

PR #116, opened 2019-11-22
Add a Julia wrapper

findmyway:
Hi everyone,

I want to add a Julia wrapper for OpenSpiel in this PR and would like to hear your thoughts.

Now we already have the Python and the Swift bindings. It would be great to add another Julia wrapper given that there's a fast-growing machine learning community in Julia. Just like Swift, with this wrapper, we can use *a single programming language for the entire OpenSpiel environment, from game implementations to algorithms and deep learning models*.

Here I use the [CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl), which is similar to pybind11, to write the Julia wrapper. This draft PR works well according to my test. Before moving on, I'd like to hear your thoughts.

<hr>

TODO:

- [x] Test on multiple Julia version (Tested on Julia@v1.1, v1.2, v1.3)
- [x] README.md
- [ ] ~~Register OpenSpiel.jl~~ (CxxWrap.jl@v0.9.0 is not registered yet. Watch https://github.com/JuliaInterop/CxxWrap.jl/issues/175)


lanctot:
Wow, this is really neat!!!

I would be very happy to support a Julia API. That said, I have zero experience with Julia and I'm not sure how much knowledge there is about it on the team. So I can't comment much on the code or approach, but it looks quite clean to me (and yes, looks quite analogous to pyspiel.cc). I have pointed the team to the PR and asked if there's anybody who knows Julia who can comment.

Would it be possible to add a Julia example (that simply executes runs a uniform random policy and prints out the states and actions similar to [examples/example.cc](https://github.com/deepmind/open_spiel/blob/master/open_spiel/examples/example.cc) and [python/examples/example.py](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/examples/example.py))? It would be neat to see what that looks like in Julia and confirm that the API actually works.

A note about longer-term support: if none of us know Julia, we would need someone to support any contributions in Julia. Is this something that you could help us with, even if just initially?

lanctot:
Hi @findmyway, I noticed you have a wget with something that installs in /usr/local. Does Ubuntu/Debian not have a julia-dev that could be installed via apt-get?

lanctot:
https://packages.debian.org/buster/libjulia-dev

findmyway:
Hi @lanctot, Thanks for providing the link (To be honest I have never use `apt-get` to install Julia before...).

The problem is that according to my test the latest CxxWrap still has some problem with Julia 1.0.x. Hopefully, it will be addressed in the next release. So here I download the Julia v1.2 and use it directly.

Generally speaking the modifications of files except for `spieljl.cc` and `CMakeFiles.txt` are only exists for Travis CI. I'll create another independent registered package later, then people just need to run 

```julia
] add OpenSpiel
```

to install and use this package.

<hr>

> Would it be possible to add a Julia example (that simply executes runs a uniform random policy and prints out the states and actions similar to examples/example.cc and python/examples/example.py)? 

I just added an `example.jl` file. And you can see the result [here](https://travis-ci.org/findmyway/open_spiel/jobs/615582008#L2990-L3082).

>  Is this something that you could help us with, even if just initially?

Yes, I would be glad to like to help.

<hr>

Although the example works well now, some API and more serious tests are not done yet. Once I finished those parts I'll turn this PR into **Ready for Review**.

Just let me know if you have any concern :smile:

lanctot:
I see, ok, but installing arbitrary code in root-privileged directories is bad practice, we should avoid it. Seems like a prefix of $HOME work would still work, I'd prefer that. What do you think? (Only as a temporary measure until Julia 1.2 is in the repos.) Then an expert can override this and install in in /usr/local if they want to, but it's not done by default.

The other thing is: I think you're handing the optional add-ons differently than the ones we have imported recently (e.g. Hanabi and universal_poker). If so: we should make this consistent. Once that's done, could you add a `README.md` in the `julia/` subdirectory explaining how to enable it (similar to `games/hanabi` and `games/universal_poker`)?

Ultimately we'll need to do more standardizing of how we handle optional add-ons like this, because there will be more on the way. E.g. wget is only needed for the Julia download so we should isolate it to the install of Julia deps (ideally a script per add-on called 'install_deps.sh' that gets called by the main script), but these are changes we can do after we've merged this. For now, it'd be good just to be consistent with our other optional dependencies like Hanabi and ACPC (universal poker).

findmyway:
&gt; Seems like a prefix of $HOME work would still work, I'd prefer that.&nbsp;


Sure. That would be much better.


&gt;&nbsp;I think you're handing the optional add-ons differently than the ones we have imported recently (e.g. Hanabi and universal_poker).&nbsp;
Yes. That is intentional. The reason is that, the julia wrapper is independent with the Python/Swift wrapper and it would be better to place it in the same level.


&gt;&nbsp;Ultimately we'll need to do more standardizing of how we handle optional add-ons like this, because there will be more on the way.


Agree. In fact, currently `install.sh` and `build_and_test.sh` are written arround the the Python wrapper. But for Julia wrapper, many packages are unnecessary. Like you said, we can do it after this PR get merged.


&gt;&nbsp;Once that's done, could you add a&nbsp;README.md&nbsp;in the&nbsp;julia/&nbsp;subdirectory explaining how to enable it&nbsp;


Sure. I will explain how to use the Julia wrapper for both users and developers in detail with README.md. Once I registered the OpenSpiel.jl (usually need three days),&nbsp;I'll simplify this PR further.


By the way, do you have any plan to make a release?


Thamks,
Jun

------------------ Original ------------------

lanctot:
Ok, this all sounds great.

> By the way, do you have any plan to make a release? 

Actually, yes, we have been talking about it lately. We've gotten a lot of additions since initial release and we have a few more planned. I'd personally love to get Tensorflow 2.0 support officially, if we can swing it. We have a Go API almost ready, thanks to some amazing work by Julien Perolat.

We also have a number of engineers who'll be adding things to the repos in December.

I am away a good portion of December + January and there are a few AI conference deadlines which might keep us very busy.  So I think maybe mid February would be a good time for a release.

It would be great if we could have the Julia API in by then, fully tested, and maybe with a few example algorithms, too. It'll make a great feature to list in the release notes! Thanks for doing this.

lanctot:
One follow-up: I think we should treat the optional add-ons in all the same way. Not sure what you meant by your argument above about why this should be done separately. We plan to do the same for the Go API.

findmyway:
After merging the latest master branch, I think now I understand what you mean. Will keep the add-ons as consistent as possible later.

lanctot:
Hi @findmyway , thanks for your continued work on this. I just wanted to let you know that Monday we will do an update that might break your changes. It's a minor re-wording change to several API functions, so it should be an easy merge via a pull from master, but you will have to apply several changes to your API, so I wanted to let you know ahead of time. Also Travis will probably show as broken after the change until you make those fixes.

findmyway:
Hi @lanctot , thanks for letting me know in advance. I'll make changes correspondingly later.

lanctot:
Hi @dennisjay : ok, I've just done the update now. We renamed a few core API functions, so you'll need to change those too. 

My best advice is to look at the changes in `pyspiel.cc` in this commit: https://github.com/deepmind/open_spiel/commit/ad858de0d69d2398b1d429eb1ab57d58b4684243#diff-a98df0e7ba0e07e097211a9455ba8685

findmyway:
Oops, the Hanabi environment just made a big change last night https://github.com/deepmind/hanabi-learning-environment/commit/5df6a73fd885618ba998f64efb728d493060e09a , which breaks the Travis CI.

lanctot:
Ah yes @jblespiau can you disable the Hanabi in Travis as well until it is fixed?

lanctot:
(We saw this yesterday and disabled Hanabi internally until we can fix it, but to the best of our knowledge it seemed to be disabled in Travis... I guess we missed something!)

findmyway:
Hi @lanctot , I think this PR is ready for review.

findmyway:
I notice that James Bradbury is one of the co-authors of *OpenSpiel: A Framework for Reinforcement Learning in Games*. And I think I had met him in the last JuliaCon. Not sure if he has an interest in giving a review. (cc @jekbradbury)

lanctot:
Ok great. I will take a look and also try to find someone who knows Julia.

We will also do an update soon (in an hour or so) that fixes the broken Hanabi build. Could you merge and rerun travis after it? I will reply when it is done.

lanctot:
> We will also do an update soon (in an hour or so) that fixes the broken Hanabi build. Could you merge and rerun travis after it? I will reply when it is done.

Ok, I just updated. Can you merge from master again and rerun Travis?

findmyway:
Updated

lanctot:
Ah, yes, of course! I had forgotten he had experience with Julia. I will try him :) Thanks!

jekbradbury:
Happy to review this! Will get to it sometime in the next couple days.

.travis.yml
```diff
@@ -37,6 +37,7 @@ matrix:
     env:
       # Unfortunately, the build fails on MacOs for universal poker.
       - BUILD_WITH_ACPC="OFF"
+      - BUILD_WITH_JULIA="OFF" # untested on mac yet
```
jekbradbury: Would you like me to try building the Julia wrapper on Mac?

docs/julia.md
```diff
@@ -0,0 +1,86 @@
...
+    ```
+
+1. Install
+    ```julia
+    ] dev ./open_spiel/julia  # run in Julia REPL
+    ```
+
+## Example
+
+Here we demonstrate how to use the Julia API to play one trajector:
```
jekbradbury:
```suggestion
Here we demonstrate how to use the Julia API to play one game:
```

docs/julia.md
```diff
@@ -0,0 +1,86 @@
...
+    while !is_terminal(state)
+        if is_chance_node(state)
+            outcomes_with_probs = chance_outcomes(state)
+            println("Chance node, got $(length(outcomes_with_probs)) outcomes")
+            actions, probs = zip(outcomes_with_probs...)
+            action = actions[sample(weights(collect(probs)))]
+            println("Sampled outcome: $(action_to_string(state, action))")
+            apply_action(state, action)
+        elseif is_simultaneous_node(state)
+            chosen_actions = [rand(legal_actions(state, pid-1)) for pid in 1:num_players(game)]  # in julia, index starts with 1
```
jekbradbury:
```suggestion
            chosen_actions = [rand(legal_actions(state, pid-1)) for pid in 1:num_players(game)]  # in Julia, indices start at 1
```

docs/julia.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Here we need the StatsBase package for weighted sampling
+using Pkg
+Pkg.add("StatsBase")
+using StatsBase
+
+function run_once(name)
+    game = load_game(name)
+    state = new_initial_state(game)
+    println("initial state of game[$(name)] is:\n$(state)")
```
jekbradbury:
```suggestion
    println("Initial state of game[$(name)] is:\n$(state)")
```

docs/julia.md
```diff
@@ -0,0 +1,86 @@
...
+run_once("tic_tac_toe")
+run_once("kuhn_poker")
+run_once("goofspiel(imp_info=True,num_cards=4,points_order=descending)")
+```
+
+## Q&A
+
+1. Why is this package named `OpenSpiel_jll` but not `OpenSpiel`?
+
+    The reason is that we plan to use [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl) for the building process once the dependencies and APIs are stable. So by convention, this package is named `OpenSpiel_jll`. Another package named `OpenSpiel` will be registered later.
```
jekbradbury: Maybe something like:
```suggestion
    This package is a low-level wrapper for the OpenSpiel C++ codebase, designed to use the [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl) infrastructure for building and distributing binary dependencies for Julia packages. Following the Julia community convention, that makes it a "JLL". Another package named `OpenSpiel` will be registered later, with a dependency on `OpenSpiel_jll` and containing pure-Julia games and algorithms.
```
For my own edification: do you know if anyone has registered a jll that uses CxxWrap yet?

docs/julia.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. What is `StdVector`?
+
+    `StdVector` is introduced in [CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl) recently. It is a wrapper of `std::vector` in the C++ side. Since that it is a subtype of `AbstractVector`, most functions should just work out of the box.
+
+1. `0-based` or `1-based`?
+
+    Considering that this package is very low level, most APIs are `0-based` to keep the code concise and consistent. Especially take care of the `Player` id which starts from `0`. And it wouldn't surprise you that some types from the Julia side, like `StdVector`, are `1-based`.
```
jekbradbury:
```suggestion
    As this package is a low-level wrapper of OpenSpiel C++, most APIs are zero-based: for instance, the `Player` id starts from zero. But note that some bridge types, like `StdVector`, implicitly convert between indexing conventions, so APIs that use `StdVector` are one-based.
```

open_spiel/julia/src/patch.jl
```diff
@@ -0,0 +1,43 @@
...
+    ps_pairs = ["$k => $v" for (k, v) in zip(keys(ps), values(ps))]
+    println(io, "StdMap{$K,$V} with $(length(ps_pairs)) entries:")
+    for s in ps_pairs
+        println(io, "  $s")
+    end
+end
+
+load_game(s::Union{String, CxxWrap.StdLib.StdStringAllocated}; kw...) = length(kw) == 0 ? _load_game(s) : _load_game(s, StdMap{StdString, GameParameter}([StdString(string(k)) => v for (k,v) in kw]))
+
+load_game_as_turn_based(s::Union{String, CxxWrap.StdLib.StdStringAllocated}; kw...) = length(kw) == 0 ? _load_game_as_turn_based(s) : _load_game_as_turn_based(s, StdMap{StdString, GameParameter}([StdString(string(k)) => v for (k,v) in kw]))
```
jekbradbury: These two lines are probably a little too long 🙂

open_spiel/julia/test/games_api.jl
```diff
@@ -0,0 +1,148 @@
+@testset "Games" begin
+
+@testset "check registered names" begin
+    game_names = registered_names()
+
+    expected = Set([
```
jekbradbury: I'm not sure this is a good idea; it adds another place that has to be updated when anyone adds a game, but without adding a check that could signal meaningful failure.

.travis.yml
findmyway: That would be great! I don't have a Mac computer at hand so it's not very convenient for me to test it. I guess the `install.sh` needs some modification correspondingly.

docs/julia.md
findmyway: As far as I know, no one has registered a jll that uses CxxWrap yet. Because CxxWrap@v0.8.2 has some problems with Julia v1.0.3 (only fixed recently) and libcxxwrap itself is not registered yet.

open_spiel/julia/test/games_api.jl
findmyway: I agree. However, this test case is directly copied from the Python side. I'm not sure how to improve this test case further. @lanctot any idea?

open_spiel/julia/src/patch.jl
findmyway: Thanks for pointing out. I break them into separate lines.

open_spiel/julia/test/games_api.jl
lanctot: Sorry I missed this. I agree with James (and we should remove that test from the Python ones too).

lanctot:
BTW I changed C++ compiler to clang. This might affect this code. Can you try pulling form master and checking that everything still works?

open_spiel/julia/test/games_api.jl
findmyway: I just removed this test case here.

.travis.yml
findmyway: It seems that CxxWrap has a bug on Mac. https://github.com/JuliaInterop/CxxWrap.jl/issues/182

findmyway:
I just merged the latest master branch. It seems everything still works fine on Ubuntu. But with mac os, there seems to be a bug with CxxWrap (https://github.com/JuliaInterop/CxxWrap.jl/issues/182).

lanctot:
Thanks @jekbradbury! Thank you @findmyway for your patience on this. I was waiting for a few days to see if the Clang / MacOS issue had been taken care of and looks like you found a way (username checks out!) 

I'm quite busy at the moment but I expect this is good to pull in now, I will try to see if someone else from the team will import it, so it can be merged soon. 

Very nice contribution, thank you!

lanctot:
Oh can I ask you to fix that one last conflict?

findmyway:
It seems the newly added `skat` game is using unicode to show the state. I'll investigate the test error later today.

findmyway:
After investigating, I think the error with showing the state of `skat()` game comes from the different way of treating `std::string`. For Python, from [the doc of pybind11](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html):

> When a C++ function returns a std::string or char* to a Python caller, pybind11 will assume that the string is valid UTF-8 and will decode it to a native Python str, using the same API as Python uses to perform bytes.decode('utf-8'). If this implicit conversion fails, pybind11 will raise a UnicodeDecodeError.

But for CxxWrap, each element in `std:string` is treated as an independent unit (aka [`Char`](https://docs.julialang.org/en/v1/base/strings/#Core.Char) in Julia). 

Personally, I would use `std:wstring` to show those [card symbols](https://github.com/deepmind/open_spiel/blame/master/open_spiel/games/skat.cc#L60-L64). But that means all the `ToString` methods need to be changed, which is not worth to do so. I've opened an issue in [CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl/issues/185) and let's wait and see the package author's thoughts. If it wouldn't be fixed from the CxxWrap side, then I'll write some patch code to make it work.

<hr>

Besides, the result of the travis build is quite strange to me. It seems [the build on the Mac](https://travis-ci.org/deepmind/open_spiel/jobs/634060904?utm_medium=notification&utm_source=github_status) is entering https://github.com/deepmind/open_spiel/blob/master/open_spiel/scripts/build_and_run_tests.sh#L125-L137 . However, [the build on Ubuntu](https://travis-ci.org/deepmind/open_spiel/jobs/634060903?utm_medium=notification&utm_source=github_status) is entering https://github.com/deepmind/open_spiel/blob/master/open_spiel/scripts/build_and_run_tests.sh#L138-L162 I'm not sure if I missed something here. Do you have any idea?

locked-deepmind:
Thanks for the thorough investigation. If it is a major pain, we could
switch the Skat game to use ASCII.

lanctot:
Does pybind11 support std::wstring? If so, we could add a `ToWideString` or `ToUnicodeString` to the API, which returns ToString converted to wide characters by default, but can be overridden for nicer state string representations. Then move the current skat string to override this function and provide a simpler ASCII one in addition.

What do you think @locked-deepmind ?

jekbradbury:
@findmyway I think this is a bug in CxxWrap.jl, in that everything else in Julia works just like pybind11 and assumes any memory typed as a string contains Unicode code points encoded with UTF-8 (while supporting invalid UTF-8 for pass-through). Either CxxWrap's `StdString` should behave like this, or there should be a way of using CxxWrap on `std::string`s that behaves this way; you shouldn't have to change the C++ implementation to use wstrings.

findmyway:
Hi @lanctot , @locked-deepmind 

Thanks to @jekbradbury 's suggestion, I add {the following line](https://github.com/deepmind/open_spiel/pull/116/files#diff-249d595fcdaf07842d8bfbd6afa0f40fR1) to make `std:string` behaves just like the one in pybind11.

```julia
Base.print(io::IO, s::CxxWrap.StdLib.StdStringAllocated) = write(io, [reinterpret(UInt8, s[i]) for i in 1:length(s)])
```

And now all the tests pass again.

PR #119, opened 2019-11-24
Call `GetStatePolicy` with `State` instead of `std::string`.

dmorrill10:
This allows one to compute best responses for `Policy`s that do not implement `GetStatePolicy` with an info state string, e.g., non-tabular policies that choose action distributions accoring to state features. Since the default version of `GetStatePolicy` given a `State` defers to the string version, this does not require changes in any current `Policy` classes.


jblespiau:
It's already submitted internally, and I am pushing it to github.

lanctot:
@jbledpiau: Are you doing something different with the imports? Why did this not get closed with the update

jblespiau:
I did nothing differently than usual. I do not know why this PR was not closed automatically.

lanctot:
Ok, makes sense. Thanks!

PR #123, opened 2019-12-03
Update evaluate_bots.py

findmyway:
`bot.step(state)` is assumed to return an `Action` here


jblespiau:
Thanks. I have merged it manually (i.e. I did the change) internally, and will be available at the next release.

Issue #124, opened 2019-12-04
Using Swift OpenSpiel in XCode


sbodenstein:
Running into a problem including OpenSpiel in a Swift XCode 11 project (obviously using the S4TF compiler toolchain): I want to use it in my own project via `import OpenSpiel`, but if I try to include it as a dependency ('Swift Packages' -> 'Add Package Dependency'), it won't compile as it complains that 'Packages are not supported using the legacy build system'. And without using the legacy build system, S4TF doesn't work. 

Has anyone managed to get this to work? 

@taliesinb, @saeta

saeta:
Great question @sbodenstein. I've for now been doing development on Linux... I just tried an alternate variation of your steps where I instead create a project on the command line (`swift project init ... && $EDITOR Package.swift`) and then run `swift package generate-xcodeproj` and then finally open the resulting `XXX.xcodeproj` file with XCode, but that didn't work for me either (can't import `OpenSpiel` when trying to build within XCode). [For posterity, my `Package.swift` is below.] It builds & runs just fine on the command line, however.

@dan-zheng do you have a guess as to the right incantation / sequence of steps to get this working with XCode?

```
saeta-macbookpro2:foo saeta$ cat Package.swift 
// swift-tools-version:5.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "foo",
    dependencies: [
        // Dependencies declare other packages that this package depends on.
        // .package(url: /* package url */, from: "1.0.0"),
        .package(url: "https://github.com/deepmind/open_spiel.git", .branch("master")),
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "foo",
            dependencies: ["OpenSpiel"]),
        .testTarget(
            name: "fooTests",
            dependencies: ["foo"]),
    ]
)
```

dan-zheng:
> Has anyone managed to get this to work?

Unforunately, Xcode is known to not work well with Swift for TensorFlow toolchains. Xcode's SwiftPM integration in particular is known to not work, because Swift for TensorFlow toolchains require the legacy build system, as you encountered.

Swift for TensorFlow toolchains do work with SwiftPM though! I use the same steps as @saeta shared above for creating a SwiftPM project:
```console
$ mkdir foo
$ cd foo
$ swift package init
[...]

PR #125, opened 2019-12-05
Add instructions for using Swift OpenSpiel in XCode

sbodenstein:
This adds information for developers to workaround the issue #124. This workaround was suggested by @dan-zheng and @saeta. Also a simple example of using Swift OpenSpiel is added.


sbodenstein:
@googlebot I signed it!

PR #131, opened 2020-01-07
Fix invalid function call in benchmark_game.

jhtschultz:
benchmark_game throws an error on games with chance nodes and simultaneous move games because InformationStateTensor and ObservationTensor are called with kChancePlayerId and kSimultaneousPlayerId. Here's a simple fix.


lanctot:
Nice catch, thanks! Will import it and it should go into the code on the next update (please don't close the issue, as it will be closed when merged.)

PR #143, opened 2020-02-04
Add `InformationStateString` to Connect Four

davidcotton:
Add `InformationStateString` to Connect Four to as it is required to run Fictitious Play


lanctot:
Thanks! Will import it now, should go in on the next update.

PR #144, opened 2020-02-05
Add `InformationStateString` to Chess

davidcotton:
Similar to #143 add `InformationStateString` to Chess. Also fixes minor typo in FP.


lanctot:
Cool, thanks again. Imported it, and will go out on next update (likely Monday)

PR #150, opened 2020-02-20
Increase Required Pip Version

SethKitchen:
Tensorflow and jaxlib installs will fail unless latest pip version is installed -- default Windows Ubuntu pip version is very low.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 # The single source of truth for the OpenSpiel pip dependencies.
+pip >= 20.0.2
 absl-py == 0.9.0
 tensorflow < 2.0, >= 1.15.1
 dm-sonnet == 1.32


lanctot:
Nice! 

> Tensorflow and jaxlib installs will fail unless latest pip version is installed -- default Windows Ubuntu pip version is very low

Right, we currently have the pip upgrade in the installation instructions. Hmm... I didn't realize you could actually put the pip version in `requirements.txt`. I am not super knowledgeable on the Python dependency system; I wonder if this has any problematic consequences?

I will ask someone who knows a bit more to come comment, but I guess it should be ok.

lanctot:
I have gotten a few reassurances that this should be fine. I will import it today so it should go out in Monday's update. Thanks!

Issue #149, opened 2020-02-20
Persisting checkpoints of CFR solvers


inejc:
I can't seem to find any support for persisting the state of C++ CFR solver implementations (apologize if it exists and this issue is redundant). I did find pickling support for game and state objects though.

I believe it would be useful to be able to specify where and how often checkpoints (from which self-play can continue at some later time) are saved to disk. I can think of two reasons:
- this would make training more robust (e.g. when it takes longer times)
- it would allow for usage of interruptible cloud VMs (reduce the cost of training significantly)

I haven't spent much time on this but maybe there are other solvers where this would be useful?

elkhrt:
Yes, I agree this could be generally useful to deal with preemption. I
don't entirely understand the parallelism you're proposing in your second
point though.

Contributions welcome!

inejc:
Just saying that this feature would enable training on virtual machines that might be terminated at any time but are offered at much lower prices (e.g. GCP preemptible instances or EC2 spot instances).

In any case, I will look into this and use this thread for potential questions.

inejc:
A couple of follow up questions:

1. I see there is also no support for the serialization of `Policy` for later inference. Is persistence in C++ handled somehow differently, e.g. with external dependencies?

2. Why do comments above the `AveragePolicy` methods for CFR solvers say that the returned policy instance should only be used during the lifetime of the solver object? As far as I can see a `CFRAveragePolicy` object only contains `info_states_` which is passed by value and `uniform_policy_` which is passed as a shared pointer, so both instances should exist after the solver object is destroyed?

3. `ExternalSamplingMCCFRSolver` and `OutcomeSamplingMCCFRSolver` keep track of an explicit, tabular `uniform_policy_` for the default policy which seems like a lot of overhead for large games and even making the implementation infeasible for games with sampled stochastic chance nodes. It would be nice if `CFRAveragePolicy`'s `default_policy_` would be a `shared_ptr<Policy>` and the above CFR solvers would have an additional constructor where an arbitrary `Policy uniform_policy_` would be passed in (e.g. a `UniformPolicy`).

4. Are there any other obvious things I should look into when designing and planning a large game to be used with `ExternalSamplingMCCFRSolver`?

If the answer to 1. is no, I think the first step for me to explore should be the serialization of `Policy`. This also seems like the first step towards the ability to save checkpoints as some of the CFR solvers contain explicit, default policy objects.

lanctot:
> A couple of follow up questions:
> 
[...]

PR #154, opened 2020-02-27
Remove necessity to keep explicit default policy for MCCFR solvers

inejc:
These changes are based on discussions in https://github.com/deepmind/open_spiel/issues/149. The idea is to remove the need to keep an explicit default policy within MCCFR solvers and thus make them feasible to run on large games where memory constraints are tighter.

The proposed changes allow for usage of:
- `TabularPolicy` which is the default behavior and the current state
- `UniformPolicy` for cases where the inferred average CFR policy will only ever be queried with a `State` instance
- `nullptr` which will enable to also query the inferred policy with an `info state string` but move the handling of info state lookup fails to the external caller (empty policy would be returned in that case)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,5 +36,6 @@ open_spiel/scripts/jill.sh
 
 # IDE
 .idea/
+.vscode/
 
 open_spiel/cmake-build-debug/
--- a/open_spiel/algorithms/cfr.cc
+++ b/open_spiel/algorithms/cfr.cc
@@ -24,8 +24,22 @@ namespace algorithms {
 
 CFRAveragePolicy::CFRAveragePolicy(
     const CFRInfoStateValuesTable& info_states,
-    std::shared_ptr<TabularPolicy> default_policy)
+    std::shared_ptr<Policy> default_policy)
     : info_states_(info_states), default_policy_(default_policy) {}
+
+ActionsAndProbs CFRAveragePolicy::GetStatePolicy(const State& state) const {
+  ActionsAndProbs actions_and_probs;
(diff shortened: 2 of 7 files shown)


inejc:
@googlebot I signed it!

lanctot:
This is great, thanks! 

I'm making a few (mostly) cosmetic changes. Turns out we don't have to store a tabular policy after all (we have a UniformPolicy object and the best response code does state-based queries already.

Should get merged in Monday's update. (Please don't close the PR.)

lanctot:
Nevermind, I spoke too soon. They do indeed need to be tabular policies due to assumptions in the best response code (which we should probably tweak eventually...)

inejc:
> Nevermind, I spoke too soon. They do indeed need to be tabular policies due to assumptions in the best response code (which we should probably tweak eventually...)

I see... I suppose this isn't a blocker for the proposed changes?

One minor thing: we should probably rename `uniform_policy_` in solvers to `default_policy_` since this could, in theory, be something different to uniform now. Please LMK whether I can add commits to this PR or you want to potentially change that later?

lanctot:
> I see... I suppose this isn't a blocker for the proposed changes?

Correct! It's already under internal review, which should be easy to get done today, so will almost surely get merged in Monday's update.
 
> One minor thing: we should probably rename `uniform_policy_` in solvers to `default_policy_` since this could, in theory, be something different to uniform now. Please LMK whether I can add commits to this PR or you want to potentially change that later?

Yep, that was one of the changes I made :) (renaming to default_policy_) Please don't add commits at this point because I've already imported the PR and the changes would clash with mine.

PR #158, opened 2020-03-03
feat: add Dockerfile

yarncraft:
**Example usage:**
```bash
docker build -t openspiel .
docker run openspiel python3 python/examples/matrix_game_example.py
````

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM ubuntu:18.04
+RUN apt update
+RUN dpkg --add-architecture i386 && apt update
+RUN apt-get -y install \
+    clang \
+    curl \
+    git \
+    python3 \
+    python3-dev \
+    python3-pip \
+    python3-setuptools \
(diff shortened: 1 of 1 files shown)


yarncraft:
@lanctot I've tested it on the Python examples and everything seems to run fine. You could use Github Actions to build the docker image and push it onto the Github Package Registry of your organisation so that people simply can _docker pull_ the image. An example workflow is demonstrated here: https://github.com/yarncraft/dockercicd!

You just need to configure the secrets and you're good to go 👍

lanctot:
Hmm, I'm having issues importing this. Tryng to figure out where the Dockerfile lives. Am I understanding correctly that it's at the top-level (root) of the repos?

lanctot:
> Hmm, I'm having issues importing this. Tryng to figure out where the Dockerfile lives. Am I understanding correctly that it's at the top-level (root) of the repos?

Scratch that -- I think I've figured it out.

Issue #159, opened 2020-03-06
Issue with Julia API and passing shared_ptr<Evaluator> as an argument


lanctot:
Hi @findmyway ,

We recently had to change `MCTSBot` to use a `shared_ptr<Evaluator>`. However, when I passed it directly via the constructor, I got a strange problem with the Julia API test that I couldn't explain (see below). 

The only thing that fixed it was to change the constructor to take a `const std::shared_ptr<Evaluator>&` instead, which is a bit weird: see https://github.com/deepmind/open_spiel/blob/9f325f3d39ece108c0d073eb63d67974ecdec3da/open_spiel/algorithms/mcts.h#L147

Do you have any idea why the error occurs below and how we can fix it? Here's the error when we try to use `std::shared_ptr<Evaluator>` directly (and changing the signature of MCTSBot in the wrapper `spieljl.cc`):

```
can play tic_tac_toe: Error During Test at /usr/local/google/home/lanctot/tmp/open_spiel/open_spiel/julia/test/bots.jl:8
  Got exception outside of a @test
  MethodError: convert(::Type{CxxWrap.StdLib.SharedPtrAllocated{Evaluator}}, ::CxxWrap.StdLib.SharedPtrAllocated{Evaluator}) is ambiguous. Candidates:
    convert(::Type{T}, p::T) where T<:CxxWrap.SmartPointer in CxxWrap at /usr/local/google/home/lanctot/.julia/packages/CxxWrap/TZf0M/src/CxxWrap.jl:108
    convert(::Type{T1}, p::T2) where {T, T1<:CxxWrap.SmartPointer{T}, T2<:CxxWrap.SmartPointer{T}} in CxxWrap at /usr/local/google/home/lanctot/.julia/packages/CxxWrap/TZf0M/src/CxxWrap.jl:112
  Possible fix, define
    convert(::Union{Type{T1<:CxxWrap.SmartPointer{T}}, Type{T<:CxxWrap.SmartPointer{T}}}, ::Union{T2<:T1<:CxxWrap.SmartPointer{T}, T<:CxxWrap.SmartPointer{T}})
  Stacktrace:
   [1] cconvert(::Type, ::CxxWrap.StdLib.SharedPtrAllocated{Evaluator}) at ./essentials.jl:387
[...]

PR #160, opened 2020-03-06
fix share ptr convert error

findmyway:
fix https://github.com/deepmind/open_spiel/issues/159

### DIFF
--- a/open_spiel/examples/gtp.cc
+++ b/open_spiel/examples/gtp.cc
@@ -40,7 +40,7 @@ std::unique_ptr<open_spiel::Bot> MakeBot(
     const open_spiel::Game& game,
     std::shared_ptr<open_spiel::algorithms::Evaluator> evaluator) {
   return std::make_unique<open_spiel::algorithms::MCTSBot>(
-      game, std::move(evaluator), /*uct_c=*/2, /*max_simulations=*/1000,
+      game, evaluator, /*uct_c=*/2, /*max_simulations=*/1000,
       /*max_memory_mb=*/0, /*solve=*/true, /*seed=*/0, /*verbose=*/false);
 }
 
--- a/open_spiel/algorithms/mcts.cc
+++ b/open_spiel/algorithms/mcts.cc
@@ -191,7 +191,7 @@ std::vector<double> dirichlet_noise(int count, double alpha,
   return noise;
 }
 
-MCTSBot::MCTSBot(const Game& game, const std::shared_ptr<Evaluator>& evaluator,
+MCTSBot::MCTSBot(const Game& game, std::shared_ptr<Evaluator> evaluator,
                  double uct_c, int max_simulations, int64_t max_memory_mb,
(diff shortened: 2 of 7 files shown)


findmyway:
Hi @lanctot ,

This is just a quick fix. I report the error here https://github.com/JuliaInterop/CxxWrap.jl/issues/212 

By the way, do you have any timeline to tag a release? I have registered `OpenSpiel` and `OpenSpiel_jll` based on a commit several weeks ago in Julia, so that users can simply run `] add OpenSpiel` in Julia REPL to use this package (but it only works in Linux for now). Could you kindly ping me a week ahead so that I can update the doc in time (and hopefully make it also work on Mac 
 at that time)

lanctot:
Wow, amazing response time, thanks! Though that fix is really not obvious! It's good to have the example in case we run into it again.

Tagging a release is a good idea. We have talked about it a lot but not done it. Let me discuss it with the team but I expect we wil do it in the coming months. I will give you a heads up when we do.

findmyway:
> It's good to have the example in case we run into it again.

Yeah, [the following line](https://github.com/deepmind/open_spiel/pull/160/files#diff-249d595fcdaf07842d8bfbd6afa0f40fR1) is the only relevant change:

```julia
Base.convert(::Type{CxxWrap.StdLib.SharedPtrAllocated{Evaluator}}, ::CxxWrap.StdLib.SharedPtrAllocated{Evaluator}) = p
```

Just replace `Evaluator` to some other type the next time. There should be a more elegant approach. But I just can't figure it out right now. Let's see if the author of CxxWrap.jl has any thoughts.

>  I will give you a heads up when we do.

Thanks!

lanctot:
I have imported this PR, should be merged on Monday's update.

open_spiel/examples/gtp.cc
```diff
@@ -40,7 +40,7 @@ std::unique_ptr<open_spiel::Bot> MakeBot(
     const open_spiel::Game& game,
     std::shared_ptr<open_spiel::algorithms::Evaluator> evaluator) {
   return std::make_unique<open_spiel::algorithms::MCTSBot>(
-      game, std::move(evaluator), /*uct_c=*/2, /*max_simulations=*/1000,
```
tkoeppe: This is a strict pessimisation, could you kindly explain a bit why this is necessary?

findmyway: Hi @tkoeppe ,

I think that I brought in this change by accident. During development, there was a build error without this change. But I just checked it again and found that it was unnecessary here.

Thanks for pointing this out! Do I need to make another PR to revert this change, or could you just modify it directly?

tkoeppe: Let's leave that to @lanctot to decide, thanks!

lanctot:
Let's do it from our side. Thanks @tkoeppe, I learned something about `shared_ptr` today!

PR #164, opened 2020-03-07
Wrapping docker build in compose file to put configuration in code

margoparmentier:
In line 19 and 20 of your Dockerfile, you expect the 'install.sh' file to be in the top directory of you docker build context.
In your install docs, you tell the user to build the Docker file using the default build context (which is the directory that the Dockerfile is located in), but this location does not have a 'install.sh' file available.

To solve this issue without changing the Dockerfile, I've created a docker-compose file in the top directory. This compose file has the correct and expected context (which includes the 'install.sh' file) and points to the location of the Dockerfile in the subdirectory.

I've also changed the install documentation accordingly, which accidentally also reduces it to a oneliner ;)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,8 @@
+version: "3.7"
+
+services:
+  openspiel:
+    build:
+      context: .
+      dockerfile: open_spiel/Dockerfile
+    image: openspiel
--- a/docs/install.md
+++ b/docs/install.md
@@ -92,8 +92,7 @@ Linux versions).
 In the top-level directory:
 
 ```bash
-docker build -t openspiel .
-docker run openspiel python3 python/examples/matrix_game_example.py
+docker-compose run openspiel python3 python/examples/matrix_game_example.py
 ```
 
 ## Running the first examples


lanctot:
Great, thanks! I have imported it and should get merged soon (hopefully Monday)

Travis-CI seems to be failing on the last test, but seems like an odd network error (unrelated to your changes). I've restarted it twice. Hopefully it will pass this time. Any ideas?

lanctot:
Now they're all failing. I guess it must be a network issue on Travis's side.

yarncraft:
Oops, the Dockerfile has to be on the root directory indeed, in order to get it working I can just simply be moved

lanctot:
> Oops, the Dockerfile has to be on the root directory indeed, in order to get it working it can just simply be moved

Any opinion on which option is better, just moving the Dockerfile or adding the docker-compose.yml (and why)? @margoparmentier @yarncraft

yarncraft:
> > Oops, the Dockerfile has to be on the root directory indeed, in order to get it working it can just simply be moved
> 
> Any opinion on which option is better, just moving the Dockerfile or adding the docker-compose.yml (and why)? @margoparmentier @yarncraft

Well the Docker Compose should IMO solely be used when multiple services are to be deployed. In my original pull request the Dockerfile was correctly placed in the root folder, I would just move it up one dir.

Also, the new CICD pipeline relies on the Dockerfile sitting in the root directory.

lanctot:
> Well the Docker Compose should IMO solely be used when multiple services are to be deployed. In my original pull request the Dockerfile was correctly placed in the root folder, I would just move it up one dir.
> 
> Also, the new CICD pipeline relies on the Dockerfile sitting in the root directory.

Oh that may have been my mistake on the import, oops!

I need you to be more specific. I thought I had imported the Dockerfile exactly where you had it in your PR (in the directory with `spiel.h`). When you say "root directory", do you mean the one with `CONTRIBUTING.md`?

If we move it up there, is the current installation instructions for Docker correct? https://github.com/deepmind/open_spiel/blob/master/docs/install.md

yarncraft:
I corrected the location in #165

yarncraft:
The documentation is still correct!

lanctot:
Ok I will fix this internall by moving the Dockerfile up one directory.

yarncraft:
> Ok I will fix this internall by moving the Dockerfile up one directory.

I've done this in #165 👍

lanctot:
Thanks for the contribution @margoparmentier. It turns out I messed up the original PR import, and when fixed the Dockerfile is in the place it was intended to be, so we don't need docker-compose.

margoparmentier:
OK, as long as it's fixed, I'm happy :)

PR #165, opened 2020-03-07
Move Dockerfile to root directory

yarncraft:


lanctot:
Hi @yarncraft, thanks for the PR. It's easier to fix this on my end, though, because I have to modify some internal mappings file (that's where the mistake was made, I believe.) So I'll close this and the other PR. However, can you please confirm that the current installation instructions are correct for this new location?

yarncraft:
> Hi @yarncraft, thanks for the PR. It's easier to fix this on my end, though, because I have to modify some internal mappings file (that's where the mistake was made, I believe.) So I'll close this and the other PR. However, can you please confirm that the current installation instructions are correct for this new location?

Ah I see! Ok that's fine, I will guide you through #161 afterwards

PR #171, opened 2020-03-14
Enhanced Dockerfile with Multi Staged Build

yarncraft:
Option to choose final runtime:
- Base (3.13GB) =`docker build --target base -t openspiel . --rm`
- Slim (2.26GB) = `docker build --target python-slim -t openspiel . --rm`

The second image serves for development in Python.

You can navigate through the runtime of the container (after the build step) with:
`docker run -it --entrypoint /bin/bash openspiel`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:18.04 as base
 RUN apt update
 RUN dpkg --add-architecture i386 && apt update
 RUN apt-get -y install \
@@ -11,6 +11,8 @@ RUN apt-get -y install \
     python3-setuptools \
     python3-wheel \
     sudo
+RUN mkdir repo
+WORKDIR /repo
 
 RUN sudo pip3 install --upgrade pip
 RUN sudo pip3 install matplotlib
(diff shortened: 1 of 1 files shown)


lanctot:
Hi @yarncraft, could you also update the `docs/install.md` to reflect the new choices and commands?

lanctot:
> Hi @yarncraft, could you also update the `docs/install.md` to reflect the new choices and commands?

Nevermind, I'll add it on the import now.

lanctot:
Hold on... @yarncraft , there's something I don't understand. 

Why is the second one smaller? It seems to require everything the first one uses, doesn't it? Using `COPY --from=base`?

Either that.... or it doesn't build OpenSpiel at all, which will not work. The Python API depends on the base OpenSpiel being built.

So one of these needs to be fixed.

yarncraft:
> Hold on... @yarncraft , there's something I don't understand.
> 
> Why is the second one smaller? It seems to require everything the first one uses, doesn't it? Using `COPY --from=base`?
> 
> Either that.... or it doesn't build OpenSpiel at all, which will not work. The Python API depends on the base OpenSpiel being built.
> 
> So one of these needs to be fixed.

Ok so this is exactly the purpose of having a multi-stage build in Docker. We use the base image in Ubuntu in order to compile everything. Finally we move these files to a new runtime which runs in fact on a smaller OS (more info: https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3). As commented in the Dockerfile itself, the second stage should solely be used in order to run python examples and scripts (not their C++ counterparts).

yarncraft:
Now the fun thing with multi-stage builds is that you can select your target! So in a CICD setup you can for example build two distinct images (a full blown runtime and a reduced one) by just having to change the target stage! 

You can learn more about Docker in this free guide: https://docker-curriculum.com/.

lanctot:
I see, but are we guaranteed that the target `FROM python:3.6-slim-buster as python-slim` is still an Ubuntu 18.04 machine (but just with less stuff)?

yarncraft:
> I see, but are we guaranteed that the target `FROM python:3.6-slim-buster as python-slim` is still an Ubuntu 18.04 machine (but just with less stuff)?

Buster is the codename for Debian 10 if I'm not mistaken

lanctot:
I don't understand how that coukd work.. the binaries built would be dynamically linking to certain libs, and they may not have the same versions on Debian 10.

Did you test it? It runs the python examples?

yarncraft:
> I don't understand how that could work.. the binaries built would be dynamically linking to certain libs, and they may not have the same versions on Debian 10 and so not found?
> 
> I guess it might work because Debian and Ubuntu are so similar.
> 
> Did you test it? It runs the python examples?

Never thought about that, but since there are many blog posts writing about how to reduce image sizes (such as https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3) I think it does not matter since you'll be installing the same requirements.txt (with versioning constraints) on the new OS. I am rebuilding now to show you an example

lanctot:
Ok, great. I'm fine to merge this, but would like to know that you've tested and run it :)

yarncraft:
Schermafbeelding 2020-03-16 om 19 05 16

So as you can see, Docker just completed the 29 steps for the python-slim image and running the matrix example works as expected!

PR #174, opened 2020-03-22
Add complete no limit action space.

jhtschultz:
@lanctot here's the patch to allow full no limit betting. My guiding philosophy was to make as few changes as possible. It passes all the tests on linux, aside from one which I had to remove because it made assumptions incompatible with the new larger action space. Oddly some tests failed on my mac but the full no limit action space tests work.

Before diving in too much deeper, I think it's worth taking a step back and making some broader design decisions. In particular, do we want to support action abstractions within the universal poker game itself? My answer is no, it doesn't seem like the OpenSpiel way. Action abstractions should be handled on the bot/algo end of things.

The abstractions are sort of baked into the code right now, so I wanted to get the team's input on this before making any further changes. I think I've got a pretty good handle on the OpenSpiel game API, and I'm happy to put a more work into the ACPC interface and build a rigorous test suite.

Submitting this to get the ball rolling, hope it helps!

-John

### DIFF
--- a/open_spiel/games/universal_poker.cc
+++ b/open_spiel/games/universal_poker.cc
@@ -159,6 +159,20 @@ std::string UniversalPokerState::ToString() const {
   } else {
     buf << "Player node for player " << cur_player_ << std::endl;
   }
+
+  if (betting_abstraction_ == BettingAbstraction::kFC ||
+        betting_abstraction_ == BettingAbstraction::kFCPA) {
+    buf << "PossibleActions (" << GetPossibleActionCount() << "): [";
+    for (auto action : ALL_ACTIONS) {
+      if (action & possibleActions_) {
+        buf << ((action == ACTION_ALL_IN) ? " ACTION_ALL_IN " : "");
+        buf << ((action == ACTION_BET) ? " ACTION_BET " : "");
+        buf << ((action == ACTION_CHECK_CALL) ? " ACTION_CHECK_CALL " : "");
+        buf << ((action == ACTION_FOLD) ? " ACTION_FOLD " : "");
+        buf << ((action == ACTION_DEAL) ? " ACTION_DEAL " : "");
+      }
+    }
+  }
   buf << "]" << std::endl;
   buf << "Round: " << acpc_state_.GetRound() << std::endl;
   buf << "ACPC State: " << acpc_state_.ToString() << std::endl;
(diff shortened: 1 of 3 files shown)


lanctot:
Hey @jhtschultz , thanks!

Seems like there's an issue with the playthrough test, can you take a look (I guess you just have to regenerate the playthrough, see `scripts/regenerate_playthroughs.sh`

image

lanctot:
> Before diving in too much deeper, I think it's worth taking a step back and making some broader design decisions. In particular, do we want to support action abstractions within the universal poker game itself? My answer is no, it doesn't seem like the OpenSpiel way. Action abstractions should be handled on the bot/algo end of things.

I see the argument for this but it is easier said than done. Arguably it _is_ a different actual game, and so doing it on the game side makes sense. And then any number of agents can play the game with a flick of a switch.

Not sure what you are referring to when you say "it's not the OpenSpiel way". This seems simple enough? People from our team implemented the wrapper, and I would have done it this way too... maybe because I view the different abstractions as different games, and then agents don't have to make the domain assumptions. 

What is the main benefit to doing the abstractions on the agent side?

jhtschultz:
I'll be sure to get the playthroughs working.

You're right, I definitely see the value in having action abstractions supported on the game side. I think the issue is the way the abstractions are implemented. Right now it's built into the game mechanics. Suppose you wanted to add a new abstraction, you'd have to rewrite the DoApplyAction function...clearly not ideal. The action abstraction should be separated out, and then functions like LegalActions can interact with it as necessary.

And a good action abstraction mechanism should be flexible. I can imagine a couple different types of abstractions you might want to support. One could be where you shrink the action space by only allowing bets sizes that are some multiple of the big blind (e.g. if blinds are 50-100 and we use a 10x multiple, bets would be 1000, 2000, ...). Another idea is to offer bet sizes as fractions of the pot, the way many poker players think about bet sizing.

Appreciate the feedback. Let me know if you have any other thoughts.

jhtschultz:
Playthrough test is now working. When using the existing action abstractions, universal_poker should function exactly as it did prior to the patch. The full no limit action space is only lightly tested at the moment, more tests to come.

lanctot:
> Playthrough test is now working. When using the existing action abstractions, universal_poker should function exactly as it did prior to the patch. The full no limit action space is only lightly tested at the moment, more tests to come.

Sweet. Build is currently broken, so bear with us. You will likely need to pull and/or restart Travis-CI when it's fixed.

lanctot:
> > Playthrough test is now working. When using the existing action abstractions, universal_poker should function exactly as it did prior to the patch. The full no limit action space is only lightly tested at the moment, more tests to come.
> 
> Sweet. Build is currently broken, so bear with us. You will likely need to pull and/or restart Travis-CI when it's fixed. Should have it fixed momentarily.

Took a bit longer than I thought. I just restarted the Travis-CI build now. Will look at importing this PR some time this week so it go be merged in next Monday's update.

lanctot:
Hi @jhtschultz , I'm now looking at the code (with @finbarrtimbers).  I noticed you removed one of the tests (`HUNLRegressionTests`). Did you just forget to put it back in, or was there another reason? Seems to be passing so we want to leave this test in there.

lanctot:
Also, I noticed there are no tests for the full action space? Did you intend that? At the very least, we should at least do the simulation tests. And I guess there should also be a playthrough.

lanctot:
> Also, I noticed there are no tests for the full action space? Did you intend that? At the very least, we should at least do the simulation tests. And I guess there should also be a playthrough.

Actually, I will skip the playthrough because it's an optional game that's not thoroughly supported anyway. I added simulaltion tests, and a few comments that were missing. Looks like it's good to go!

jhtschultz:
Ok great! Yeah I was planning on adding the `HUNLRegressionTests` back in with the latest patch now that it's fully backward compatible. I'm still planning on writing more tests (for the full no limit betting and ACPC wrapper as a whole), but I think the simulation tests should suffice for now. Sounds like you have everything ready to go!

lanctot:
Great. Ok this one has been imported now, so will be merged on next Monday's update. If you plan to add more tests, please wait until it's merged (further commits on this PR won't be pulled in). Thanks again!

PR #178, opened 2020-03-30
Add full no limit betting tests.

jhtschultz:
Here's some tests for full no limit betting in universal poker. We test different combinations of blinds, stack sizes, number of players, and betting patterns. Everything worked as expected.

### DIFF
--- a/open_spiel/games/universal_poker_test.cc
+++ b/open_spiel/games/universal_poker_test.cc
@@ -234,9 +234,8 @@ void ChumpPolicyTests() {
   }
 }
 
-void FullNLBettingTests() {
-  // Test #1
-  // Checks min raising functionality.
+// Checks min raising functionality.
+void FullNLBettingTest1() {
   std::shared_ptr<const Game> game = LoadGame(
       "universal_poker(betting=nolimit,"
                       "numPlayers=2,"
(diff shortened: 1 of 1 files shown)


open_spiel/games/universal_poker_test.cc
```diff
@@ -234,6 +234,167 @@ void ChumpPolicyTests() {
...
+    state->ApplyAction(i);
+  state->ApplyAction(1);  // call last raise
+  state->ApplyAction(state->LegalActions()[0]);  // deal turn
+  state->ApplyAction(state->LegalActions()[0]);  // deal river
+  SPIEL_CHECK_EQ(state->Returns()[0], state->Returns()[1]);  // hand is a draw
+  SPIEL_CHECK_TRUE(absl::StrContains(state->ToString(),
+      "ACPC State: STATE:0:cc/r4r6r8r10r12r14r16r18r20c//"
+      ":2c2d|2h2s/3c3d3h/3s/4c"));
+
+  // Test #2
```
tewalds: Shouldn't these separate tests be separate test functions?

jhtschultz: Makes sense, didn't know which was preferred. It's updated.

lanctot:
Good stuff @jhtschultz , importing it now. Will get merged on next Monday's update.

PR #181, opened 2020-04-05
fixed SampleCards bug in universal_poker/card_set

Liuweiming:
According to universal_poker_card_set_test,

>  ./universal_poker_card_set_test
CardSet: AhKsQhJhTh
Card: "
Card: &
Card: *
Card: /
Card: 2
Rank: 9632
Count Cards: 5
CardSet: AsAhAdAcKsKhKdKcQsQhQdQcJsJhJdJcTsThTdTc9s9h9d9c8s8h8d8c7s7h7d7c6s6h6d6c5s5h5d5c4s4h4d4c3s3h3d3c2s2h2d2c
Rank: 12115
Count Cards: 52
CardSet: 5c3c2c
CardSet: 5c4c2c
...

the first CardSet `4c3c2c` is missed.  After fixing, it looks like this

> ./universal_poker_card_set_test
CardSet: AhKsQhJhTh
Card: "
Card: &
Card: *
Card: /
Card: 2
Rank: 9632
Count Cards: 5
CardSet: AsAhAdAcKsKhKdKcQsQhQdQcJsJhJdJcTsThTdTc9s9h9d9c8s8h8d8c7s7h7d7c6s6h6d6c5s5h5d5c4s4h4d4c3s3h3d3c2s2h2d2c
Rank: 12115
Count Cards: 52
CardSet: 4c3c2c
CardSet: 5c3c2c
...

### DIFF
--- a/open_spiel/games/universal_poker/logic/card_set.cc
+++ b/open_spiel/games/universal_poker/logic/card_set.cc
@@ -136,7 +136,7 @@ std::vector<CardSet> CardSet::SampleCards(int nbCards) {
 
   for (uint64_t n = bit_twiddle_permute(p); n > p;
        p = n, n = bit_twiddle_permute(p)) {
-    uint64_t combo = n & cs.cards;
+    uint64_t combo = p & cs.cards;
     if (__builtin_popcountl(combo) == nbCards) {
       CardSet c;
       c.cs.cards = combo;


lanctot:
Cool, thanks.. we will take a look!

In the mean time, can you sign the CLA?

Liuweiming:
@googlebot I signed it!

Liuweiming:
@googlebot I fixed it.

lanctot:
@dennisjay can you take a look?

jblespiau:
Thanks.

I remember, when I refactoring this, that I had found that "SampleCards" was not being used, except in the tests:

  // TODO(author2): Remove?
  std::vector<CardSet> SampleCards(int nbCards);

In practice, we never sample cards, this is done in the game-tree, and we only add/remove cards.

Personally, I do not understand what SampleCards does;  it takes as argument "nbCards", but I do not see how the number of cards varies and what is the source of randomness.

Do you have understood what it is doing?

And do you need this function? What do you think of removing it completly? And change the tests a little to actually tests the features (Add/Remove) that are being used?

lanctot:
Hi @Liuweiming,

We did an update this morning that caused a conflict with master. It should be a trivial merge; can you pull from master and commit again after merging?

lanctot:
> // TODO(author2): Remove?
> std::vector SampleCards(int nbCards);
> 
> In practice, we never sample cards, this is done in the game-tree, and we only add/remove cards.
> 
> Personally, I do not understand what SampleCards does; it takes as argument "nbCards", but I do not see how the number of cards varies and what is the source of randomness.
> 
> Do you have understood what it is doing and can explain it to me?

This function can certainly use better documentation, I agree. @Liuweiming: Could you also add more comments to the code explaining what this function does?

> And do you need this function? What do you think of removing it completely? And change the tests a little to actually tests the features (Add/Remove) that are being used?

I'm inclined toward keeping since generally there will be a use for it. @jhtschultz , any opinion on this?

Liuweiming:
1. If I am correct, the `SampleCards` does not return a sampled `CardSet`, but returns all possible combination of `nbCards` cards (in lexicographical order). It works by enumerating all the  `uint64_t` integers that with `nbCards` 1-bits. Below is the bitset-view of `p`.

> p: 0000000000000000000000000000000000000000000000000000000000000111
p: 0000000000000000000000000000000000000000000000000000000000001011
p: 0000000000000000000000000000000000000000000000000000000000001101
p: 0000000000000000000000000000000000000000000000000000000000001110
p: 0000000000000000000000000000000000000000000000000000000000010011
p: 0000000000000000000000000000000000000000000000000000000000010101
p: 0000000000000000000000000000000000000000000000000000000000010110
p: 0000000000000000000000000000000000000000000000000000000000011001
...

2. In my opinion, the function should be used in function `UniversalPokerState::LegalActions` and `UniversalPokerState::ChanceOutcome` to return possible combinations of cards.  So we can deal all the hole cards or all the board cards at a time, instead of dealing one card at a time.
3. I think yes we need to change the tests. Instead of printing all the combinations, checking the number of the unique combinations should be more reasonable. But I am a bit busy right now. If I could, I will open another pull request.

Liuweiming:
@lanctot , yes, I will.

Liuweiming:
@jhtschultz, agree. The current implementation is more "open_spiel".

lanctot:
Closing this in favor of #183. See https://github.com/deepmind/open_spiel/pull/183#issuecomment-610040801

PR #185, opened 2020-04-08
Added support for Reversi/Othello

jacobaustin123:
This is a native implementation of [Reversi/Othello](https://en.wikipedia.org/wiki/Reversi). This is modified from the TicTacToe source and is fairly well tested. Several repositories have used Othello as a simpler game for AlphaZero implementations, and I found speed to be a bottleneck in [my own implementation](https://github.com/jacobaustin123/pytorch-alphazero). All tests pass and I added code for regression tests.


jacobaustin123:
A quick question - one inconvenience for both the TicTacToe and this Othello implementation is the `SPIEL_CHECK_GE(player, 0);` check when loading the `ObservationTensor` or `ObservationString`. This means you cannot get the observation tensor for any terminal state (since player is set to -4 on game end). Is this intention behavior, or can this check be removed? Being able to load the observation tensor for terminal states makes a variety of RL algorithms easier.

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  CellState cell = PlayerToState(player);
+  for (int i = 0; i < kNumCells; i++) {
+    if (board_[i] == cell) count++;
+  }
+
+  return count;
+}
+
+bool OthelloState::NoValidActions() const {
+  return (LegalRegularActions(Player(0)).empty() & LegalRegularActions(Player(1)).empty());
```
tewalds, line 153: This would be better with shortcutting logic: && instead of &

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+}
+
+std::tuple<int, int> OthelloState::XYFromCode(int move) const {
+  if (move >= kNumCells || move < 0) {
```
tewalds, line 157: It'd be more conventional to flip the order to check the bottom range first.

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  return count;
+}
+
+bool OthelloState::NoValidActions() const {
+  return (LegalRegularActions(Player(0)).empty() & LegalRegularActions(Player(1)).empty());
+}
+
+std::tuple<int, int> OthelloState::XYFromCode(int move) const {
```
tewalds, line 156: Why a tuple instead of a pair?

You can also use the c++17 destructuring to get them back into x,y.

Alternatively, a specific class like https://github.com/deepmind/open_spiel/blob/master/open_spiel/games/havannah.h#L67 may be useful.

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  if (!ValidAction(current_player_, move)) {
+    SpielFatalError(absl::StrCat("Invalid move ", move));
+  }
+
+  std::tuple<int, int> rowcol = XYFromCode(move);
+  CellState cell = PlayerToState(current_player_);
+
+  int row = std::get<0>(rowcol);
+  int col = std::get<1>(rowcol);
+  board_[row * kNumCols + col] = cell;
```
tewalds, line 209: This x,y to move is done in a few places. Is it worth pulling into a function?

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  
+  for (int direction = Direction::kUp; direction < Direction::kLast; direction++) {
+    int steps = 0;
+    if ((steps = CountSteps(current_player_, row, col, static_cast<Direction>(direction))) > 0) {
```
tewalds, line 213: this is awkward. Why not move the CountSteps call one line up?

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+}
+
+std::string OthelloState::ActionToString(Player player,
+                                           Action action_id) const {
+
+  if (action_id == passMove) {
+    return absl::StrCat(StateToString(PlayerToState(player)), "(pass)");
+  } else {
+    return absl::StrCat(StateToString(PlayerToState(player)), "(",
+                      action_id / kNumCols, ",", action_id % kNumCols, ")");
```
tewalds, line 263: Would a more standard and compact representation make sense? 
ie follow: https://en.wikipedia.org/wiki/Reversi#Rules
eg: https://github.com/deepmind/open_spiel/blob/master/open_spiel/games/havannah.cc#L147

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  }
+}
+
+bool OthelloState::IsFull() const { 
+  for (int i = 0; i < kNumCells; i++) {
```
tewalds, line 268: Is this called often enough to be worth caching? You track num_moves_ but not number of pieces placed.

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+    if (board_[i] == CellState::kEmpty) return false;
+  }
+
+  return true;
+}
+
+OthelloState::OthelloState(std::shared_ptr<const Game> game) : State(game) {
+  std::fill(begin(board_), end(board_), CellState::kEmpty);
```
tewalds, line 276: absl::c_fill

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+  board_[27] = CellState::kWhite;
+  board_[28] = CellState::kBlack;
+  board_[35] = CellState::kBlack;
+  board_[36] = CellState::kWhite;
+}
+
+std::string OthelloState::ToString() const {
+  std::string str;
```
tewalds, line 284: It could be useful to put the coordinate system on here too, especially if you make the suggested change of action string.

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+
+void OthelloState::ObservationTensor(Player player,
+                                       std::vector<double>* values) const {
+  SPIEL_CHECK_GE(player, 0);
+  SPIEL_CHECK_LT(player, num_players_);
+
+  // Treat `values` as a 2-d tensor.
+  TensorView<2> view(values, {kCellStates, kNumCells}, true);
+  for (int cell = 0; cell < kNumCells; ++cell) {
+    view[{static_cast<int>(board_[cell]), cell}] = 1.0;
```
tewalds, line 328: Many learning algorithms work better if these observations are player relative instead of absolute, so the layers would be: mine, opponent, empty; instead of p1, p2, empty.

open_spiel/games/othello.h
```diff
@@ -0,0 +1,141 @@
...
+namespace open_spiel {
+namespace othello {
+
+// Constants.
+inline constexpr int kNumPlayers = 2;
+inline constexpr int kNumRows = 8;
+inline constexpr int kNumCols = 8;
+inline constexpr int kNumCells = kNumRows * kNumCols;
+inline constexpr int kCellStates = 1 + kNumPlayers;  // empty, 'x', and 'o'.
+inline constexpr int passMove = kNumCells;
```
tewalds, line 41: kPassMove

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+
+  if (NoValidActions()) {
+    int count_zero = DiskCount(Player(0));
+    int count_one = DiskCount(Player(1));
+
+    if (count_zero > count_one) {
+      outcome_ = Player(0);
+    } else if (count_zero < count_one) {
+      outcome_ = Player(1);
+    }
```
tewalds, line 226: and if they have the same?

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+}
+
+bool OthelloState::CanCapture(Player player, int move) const {
+  if (board_[move] != CellState::kEmpty) return false;
+
+  std::tuple<int, int> row_col = XYFromCode(move);
+  int row = std::get<0>(row_col);
+  int col = std::get<1>(row_col);
+
+  bool valid = false;
```
tewalds, line 113: unused variable

open_spiel/games/othello.cc
```diff
@@ -0,0 +1,344 @@
...
+
+    board_[row * kNumCols + col] = cell;
+    
+    rowcol = GetNext(row, col, dir);
+    row = std::get<0>(rowcol);
+    col = std::get<1>(rowcol);
+  }
+}
+
+int OthelloState::DiskCount(Player player) const {
```
tewalds, line 142: Is this number worth caching?

lanctot:
Thanks @jacobaustin123 , this is a great contribution.. another important game in the history of AI!

lanctot:
> Quick question: one inconvenience when using OpenSpiel is the `SPIEL_CHECK_GE(player, 0);` check in `spiel.h` when loading the `ObservationTensor` or `ObservationString`. This means you cannot get the observation tensor for any terminal state (since player is set to -4 on game end). Is this intentional behavior/is there some way to circumvent this? Being able to load the observation tensor for terminal states makes a variety of RL algorithms easier.

Ahh yeah I think that's out-of-date; that check was put there before we had a special player id for terminal states. I completely agree, the observations should be defined at terminal states for exactly the reason you describe; in fact, I'm a bit surprised this is the first time we're running into this. 

In fact, having a bit in every observation/information state indicating whether or it's a terminal would also be good. But that's a bigger job. 

We should certainly standardize this (ensuring that observations are defined at terminals); I don't see any downside and AFAIR we even have algorithms that depend on this being true. So we should also add a check to the simulation tests. 

For now, please yes feel free to remove it in TicTacToe (as long as the observation still gets composed properly at terminal nodes). Doesn't matter if it's a new commit in this PR or a separate PR. We can take care of the other games over time.

elkhrt:
Thanks for the contribution!
I notice that ObservationString and ObservationTensor don't allow the
player to work out which pieces are theirs, so an agent would have to know
which colour it is as an additional input (in tic-tac-toe this can be
figured out from the parity, but not here). Options are:
a. Always show the game from the player's point-of-view, so planes are
(like connect-4)
b. Add another feature plane indicating which colour the observing player
is (like go)

elkhrt:
I'd rather keep the check that player > 0.
If you always specify the player_id when calling ObservationTensor, there's
no issue with negative values (rl_environment already does this).
(And different players may have different observations, so we should be
specifying the player id).

lanctot:
> I'd rather keep the check that player > 0. If you always specify the player_id when calling ObservationTensor, there's no issue with negative values (rl_environment already does this). (And different players may have different observations, so we should be specifying the player id).

Ok that makes sense. Then we can add a check to the default `ObservationTensor` that just forwards the current player to ensure it's never called at terminals, forcing the caller to use the player-based one at terminals. Same for corresponding functions. WDYT?

elkhrt:
> Then we can add a check to the default `ObservationTensor` that just forwards the current player to ensure it's never called at terminals, forcing the caller to use the player-based one at terminals. Same for corresponding functions. WDYT?

SGTM

jacobaustin123:
@elkhrt I'll add the ability to view the board from either perspective. Almost ready to push the changes. 

The only problem I see with keeping the checks as is that sometimes you need to get the current player after the game has ended. Let's say I'm using state.observation_tensor() to get the board from the current player's perspective and I suddenly find myself in a terminal state, suddenly I don't know _who_ won the game or how to display the board in the terminal state. Same problem for state.rewards() since I need to know the current player to determine what the reward is.

I feel like it would be more convenient to use another flag to disable the board after a terminal state is reached, but keep the current_player_ as the player who made the last move.

elkhrt:
The idea is that the methods give you information for all players in the
game - rewards() returns a vector, observation_tensor(player) takes a
player parameter.

The algorithm should be keeping track of which agent is controlling which
player - see rl_environment.py for an example. That agent should be
supplied with observations from the perspective of the player it is
controlling and rewards for that player.

So if you reach a terminal state, each player is awarded an individual
reward, and each player can get the observation of the terminal state from
their own point of view.

jacobaustin123:
So basically algorithms shouldn't be relying on the current_player() function? They should be keeping track of this and specifying which perspective they want to see the game from/which rewards they care about. Since they lose access to the current_player() function once a terminal state is reached?

open_spiel/games/othello.cc
jacobaustin123: I ended up discovering std::tie which allows you to do automatic unpacking from tuples/pairs. Thanks!

open_spiel/games/othello.cc
jacobaustin123: Added the function RowColToMove(row, col) and RowColFromMove(move).

open_spiel/games/othello.cc
jacobaustin123: I'm now using the standard [letter col][row number] and this is displayed on the board as well.

  a b c d e f g h
1 - - - - - - - - 1
2 - - - - - - - - 2
3 - - - - - - - - 3
4 - - - o x - - - 4
5 - - - x o - - - 5
6 - - - - - - - - 6
7 - - - - - - - - 7
8 - - - - - - - - 8
  a b c d e f g h

open_spiel/games/othello.cc
jacobaustin123: I ended up removing IsFull because it doesn't matter for Othello.

open_spiel/games/othello.cc
jacobaustin123: Added.

open_spiel/games/othello.cc
jacobaustin123: Done. You can now pass the Player, and the board string and observation tensor will switch.

open_spiel/games/othello.cc
jacobaustin123: I now set this to Player(kTerminalPlayerId).

open_spiel/games/othello.cc
jacobaustin123: Good question. This can be done for performance reasons, and it is probably not too hard to do. You just need to update it when disks are captured. I can add this if we think it's worthwhile.

elkhrt:
Relying on current_player() is perfectly reasonable, but it should be the
current_player() at the decision point, e.g. something like this:

player = state.current_player()
observation = state.observation_tensor(player)
action = agent[player].action(observation)
state.apply_action(action)
rewards = state.rewards()
next_observation = state.observation_tensor(player)
agent[player].step(observation, action, rewards[player], next_observation)

After line 4, the state may be terminal (and current_player() will no
longer be a valid player), but this should not cause any problem.

It would be an error to use state.observation_tensor() in line 6 even in
non-terminal states, because current_player() will now be the next player
to act, not the one which has just acted.

jacobaustin123:
Ok great that makes sense. I just pushed a new commit that addresses the comments @tewalds made. state.observation_tensor(player) and state.observation_string(player) now work, the player checks remain in place, and the observation_tensor/string defaults to the current player when not specified.

open_spiel/games/othello.cc
tewalds: Ah, I meant https://en.cppreference.com/w/cpp/language/structured_binding so you can even do:
auto [row, col] = GetNext(...)

though I think it'd be even better to add a Move class that decomposes it into row and col member variables. Then you'd do:
Move m(move);
f(BoardAt(m))
f(BoardAt(m.Next(direction)));
and pretty much never need to deal in row/col, but when you do you can do:
g(m.row, row.col)

open_spiel/games/othello.cc
tewalds: I meant: move < 0 || move >= kNumCells

open_spiel/games/othello.cc
```diff
@@ -150,15 +142,24 @@ int OthelloState::DiskCount(Player player) const {
...
     SpielFatalError(absl::StrCat("Move too large: ", move));
   }
 
-  return std::make_tuple(move / kNumCols, move % kNumRows);
+  return std::make_pair(move / kNumCols, move % kNumRows);
+}
+
+
+int OthelloState::RowColToMove(int row, int col) const {
+  if (row < 0 || col < 0 || row * col >= kNumCells) {
```
tewalds, line 158: row * col!?

open_spiel/games/othello.cc
```diff
@@ -173,14 +174,16 @@ CellState PlayerToState(Player player) {
   }
 }
 
-std::string StateToString(CellState state) {
+std::string StateToString(Player player, CellState state) {
```
tewalds, line 177: This is kind of odd. Usually you keep the string as absolute but just the observation tensor as player relative.

open_spiel/games/othello.cc
jacobaustin123: Not sure what I was thinking there.

open_spiel/games/othello.cc
jacobaustin123: Ok. I reverted this part to keep it absolute.

jacobaustin123:
@tewalds I added a Move class with a pretty nice API which definitely simplifies the rest of the design. Seems good.

elkhrt:
Hey Jacob - we merged your PR before your most recent change, adding the Move class. If you'd like to keep that change, you can either resolve the conflict in this PR here, or close it and open another PR.
Sorry for the hassle, and thanks again for the great contribution.

jacobaustin123:
@elkhr Glad it was merged! I just fixed the merge conflicts for the Move class version. Two differences from master:

1. ObservationString actually uses the player argument, so you can see the board from either perspective, although the default is always player 0. Doesn't change anything unless you pass ObservationString(1).
2. The observation string doesn't contain the current player name. I think you added this to the master version? It's great except it means you can't bring the board in a terminal position, since CurrentPlayer() is invalid. We could add it back and have the current player be none, or ended, or we can just remove it, like I did here.

elkhrt:
Thanks for the update!

For the Observation change, I think its neater in games of perfect information to have ToString() = ObservationString(0) = ObservationString(1), to make it clear that each player has the same information. So I've reverted to that convention, with an explicit check for terminal states.

jacobaustin123:
@elkhrt Great. Looks good to me! Thanks a lot.

PR #198, opened 2020-04-30
Fix: Returns of non-terminal KuhnState was hardcoded for 2 players

VonTum:
KuhnState::Returns should return a std::vector with an entry for every player, but for any number of players this returns {0.0, 0.0} if the game isn't terminal. This was changed to return a vector of size num_players_


VonTum:
@googlebot I signed it!

lanctot:
Nice catch, thanks!

lanctot:
I imported it, will get merged in next Monday's update.

PR #199, opened 2020-05-01
Fix attempts to make illegal moves

adam4130:
The following code attempts to compute the Nash convergence of a naive random strategy on Leduc poker:
```
import pyspiel
import numpy as np

from open_spiel.python import policy
from open_spiel.python.algorithms import exploitability

game = pyspiel.load_game("leduc_poker")
num_actions = game.num_distinct_actions()

def action_probabilities(state):
    legal_actions = state.legal_actions()
    probs = np.zeros(num_actions)
    probs[legal_actions] = 1 / len(legal_actions)
    return {action: probs[action] for action in range(num_actions)}

conv = exploitability.nash_conv(
    game,
    policy.PolicyFromCallable(game, action_probabilities)
)
```
In this game only 2 raise actions are allowed per round. However, every action is attempted regardless of the round. This results in illegal moves being made and the following error:
```
RuntimeError: ~/open_spiel/open_spiel/games/leduc_poker.cc:156 num_raises_ < kMaxRaises
num_raises_ = 2, kMaxRaises = 2
```

### DIFF
--- a/open_spiel/python/algorithms/best_response.py
+++ b/open_spiel/python/algorithms/best_response.py
@@ -110,8 +110,9 @@ class BestResponsePolicy(openspiel_policy.Policy):
       if parent_state.current_player() == self._player_id:
         yield (parent_state, 1.0)
       for action, p_action in self.transitions(parent_state):
-        for state, p_state in self.decision_nodes(parent_state.child(action)):
-          yield (state, p_state * p_action)
+        if action in parent_state.legal_actions():
+          for state, p_state in self.decision_nodes(parent_state.child(action)):
+            yield (state, p_state * p_action)
 
   def transitions(self, state):
     """Returns a list of (action, cf_prob) pairs from the specifed state."""
@@ -138,7 +139,10 @@ class BestResponsePolicy(openspiel_policy.Policy):
 
   def q_value(self, state, action):
     """Returns the value of the (state, action) to the best-responder."""
-    return self.value(state.child(action))
+    if action in state.legal_actions():
+      return self.value(state.child(action))
+    else:
+      return state.player_return(self._player_id)
 
   @_memoize_method
   def best_response_action(self, infostate):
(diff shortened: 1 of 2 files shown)


elkhrt:
Hi - thanks for the contribution!
The fix you've suggested adds quite a bit of computation to some fairly hot code paths, so I'd rather explore other options if possible. My recommended way to do what you're trying to do would be:

```
import pyspiel

from open_spiel.python import policy
from open_spiel.python.algorithms import exploitability

game = pyspiel.load_game("leduc_poker")
my_policy = policy.TabularPolicy(game)     # Default initialization is uniform random
conv = exploitability.nash_conv(game, my_policy)
```

This creates an explicit tabular policy once, so is quite a bit quicker than the approach you tried.
You can then change the policy by directly updating `my_policy.action_probability_array`.

We should also help other people not to fall into this trap. Since `PolicyFromCallable` is a legacy code path, my inclination is to add a check there that illegal actions aren't being returned by the callable.
We should also remove all references to `PolicyFromCallable` in the example code.

How does all that sound to you?

adam4130:
Thanks for the reply. I will use the `action_probability_array` going forward. I experimented a bit with instead adding a check in `PolicyFromCallable`. I tried only allowing probabilities greater than 0 as does `TabularPolicy`. This ends up breaking `action_value_vs_best_response.Calculator` where a callable policy can potentially be used to compute a `pyspiel.TabularBestResponse` which needs all legal action probabilities. I'd be willing to look into removing the callable policy from the `eva` and `fictitious_play_example` examples.

elkhrt:
Yes, `PolicyFromCallable` would have to check for legal actions. I don't mind that so much, because its a legacy code path that I don't mind slowing down.

I would be very happy to see PRs adding checks to `PolicyFromCallable` and reducing its use. Please let me know what you want to do, and I'll pick up the rest.

PR #200, opened 2020-05-03
Use `open_spiel` namespace prefix for float assertions

dmorrill10:
This is to allow float assertion macros to be used without using the `open_spiel` namespace itself. Most of the other macros use the `open_spiel` prefixes in this way so I think these two were just missed.


lanctot:
Great, thanks! Importing now, should be merged in tomorrow's update.

PR #203, opened 2020-05-04
Only build and register `run_python_test` when `BUILD_WITH_PYTHON=ON`

dmorrill10:
While `run_python_test` doesn't depend on `pyspiel`, it still requires setting the python path to include the `open_spiel` directory, which is probably not necessary if one isn't using `pyspiel`. An alternative would be to change `run_python_test` to not run a local module, but from the comment in `RunPython` and the test, it looks like this is the main purpose of `RunPython`.


lanctot:
Great, thanks!

PR #205, opened 2020-05-05
Install CMake if it is not present (macOS install.sh)

mrdaliri:
The installation script does not check CMake presence on macOS (`darwin`). `brew install cmake` is added to install it in case it is not available.


lanctot:
Nice catch, thanks!

PR #208, opened 2020-05-06
Remove python version specifier from Nox tests definition

mrdaliri:
There is no need to specify the Python version for Nox tests. Without a specific version, Nox will use whatever Python interpreter its parent environment has.

So, if we only have Python 3.6 on Travis, and Python 3.7 or 3.8 on other deployments, tests will run using the same interpreter.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -27,7 +27,7 @@ def get_distutils_tempdir():
   )
 
 
-@nox.session
+@nox.session(python="3")
 def tests(session):
   session.install("-r", "requirements.txt")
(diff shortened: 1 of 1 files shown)


lanctot:
Hmm seems like the Travis tests are failing with nox. It is picking up python 2.7. Maybe add the constraint >= 3.6 ?

lanctot:
Latest commit worked, thank you for your contribution! Importing now, will be merged in Monday's update.

mrdaliri:
Cool! Thanks.

PR #209, opened 2020-05-07
Fix unsorted legal actions.

jhtschultz:
The last update broke compatibility with some of my EFG files that used to work. The EFG file specification does not guarantee the action names will be sorted. And even if the action names are always sorted within a given node, they may not have been first encountered in that order, so the `action_ids` they map to may be unsorted. Consider first encountering the infoset with actions {"B" "C"}, followed by {"A" "B" "C"}. The `action_ids` are then B->0, C->1, A->2. So when calling `LegalActions()` on the second infoset, we get [2, 0, 1]. Sorting the `action_ids` after parsing all the actions within each node fixed the issue.

### DIFF
--- a/open_spiel/games/efg_game.cc
+++ b/open_spiel/games/efg_game.cc
@@ -14,6 +14,7 @@
 
 #include "open_spiel/games/efg_game.h"
 
+#include <algorithm>
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


lanctot:
Whoops... good catch, thanks John!

tewalds:
Is there a general test that could catch this?

lanctot:
> Is there a general test that could catch this?

Yeah pretty sure we already have one-- Ed added it when we added the requirement for sorted actions. John is using .efg files that he has not submitted yet, so my guess is it was in one of those. (But maybe we indeed don't have the test.)

lanctot:
Confirmed: https://github.com/deepmind/open_spiel/blob/4c02fe8c74ced5d448e98c68520c7394e241051d/open_spiel/tests/basic_tests.cc#L110

lanctot:
Confirmed: https://github.com/deepmind/open_spiel/blob/4c02fe8c74ced5d448e98c68520c7394e241051d/open_spiel/tests/basic_tests.cc#L287

jhtschultz:
Yeah that's how I caught it. It just so happens that none of the current EFG files trigger the error.

PR #215, opened 2020-05-08
Add efg_writer.

jhtschultz:
Converts an OpenSpiel game to a Gambit .efg file. Game transforms seemed like a good place for it but I can move it elsewhere if that's preferable.

### DIFF
--- a/open_spiel/game_transforms/CMakeLists.txt
+++ b/open_spiel/game_transforms/CMakeLists.txt
@@ -1,6 +1,8 @@
 add_library (game_transforms OBJECT
   coop_to_1p.cc
   coop_to_1p.h
+  efg_writer.cc
+  efg_writer.h
   misere.cc
   misere.h
   turn_based_simultaneous_game.cc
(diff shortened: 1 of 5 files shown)


jhtschultz:
Just looking at pull requests now and saw that Michal submitted the same thing today. Sorry about that! I’ll leave this open in case there’s any value to having a C++ version, otherwise feel free to close it.

jhtschultz:
One thing though…it’s useful to support separate information set indices for each player. If we close this it would be helpful to add that to the other version.

michalsustr:
Funny coincidence :) 

> it’s useful to support separate information set indices for each player.
As in, allow to have the same index for two different players? Can you give an example of a use-case when that would be useful? I'm not sure I can think of one.

lanctot:
Hi guys,

I think it's ok to have it both in python and C++.

Michal, the .efg files I found seem to count the infosets separately per player (including the sample one on the page explaining the format), so AFAICT it would be more consistent with the "canonical" format, but also consistent internally too.

jhtschultz:
Yeah I was basing my comment on the fact that Gambit supports that format and most of the .efg files in their repo adhere to it.

Good thing this topic got brought up! Just found a bug in efg_game as a result...

https://github.com/deepmind/open_spiel/blob/master/open_spiel/games/efg_game.cc#L497-L499

If the .efg file is counting infosets separately per player, this will incorrectly label a perfect information game as imperfect information. Confirmed it with a test on tic tac toe.

lanctot:
Yikes, nice catch! Please open an issue as a reminder for us to fix it.

PR #216, opened 2020-05-08
move python path commands to before tests

samcorzine:
This dockerfile build has been failing for me on the run_python_tests test. This edit fixes it.


lanctot:
Nice catch, thanks!

Issue #224, opened 2020-05-20
laser tag players do not act as commanded.


qmaai:
Hi I have been reading through laser_tag.cc, and open_spiel/open_spiel/integration_tests/playthroughs/laser_tag(horizon=20).txt. 

Some of the agents' behaviors are strange. Here is one example where agent A fails to step right even when it is still within the grid even after doing so. There exist numerous similar problems in the afore-mentioned txt file. Would you kindly look into that?
[laser_tag(horizon=20).txt](https://github.com/deepmind/open_spiel/files/4658201/laser_tag.horizon.20.txt)

lanctot:
Hi Gary, took a look-- yeah, I agree... that appears to be a bug. 

I will take a look, hopefully have it fixed by Monday's update.. but unfortunately can't guarantee any timeline for now, a bit busy with some paper submissions at the moment.

alexminnaar:
Been a while since I've looked at this but it looks like there's a joint action of ["step right", "move forward"] for A and B respectively, then the chance node decides to apply B's action first (which it then does) but then A's action is never applied.  I could maybe look into this this weekend.

qmaai:
@lanctot @alexminnaar  thank you for looking into this.

alexminnaar:
Looks like the issue is on [this line](https://github.com/deepmind/open_spiel/blob/c2683ca03d85e39d867594ef7f3c7f9319f18673/open_spiel/games/laser_tag.cc#L237).  The left step is there twice and one of them should be the right step.  I created a PR https://github.com/deepmind/open_spiel/pull/227/  I'm not sure about the other problems you mentioned but this should at least cover the linked issue.

qmaai:
Hi Alex. Yup I think that is it. All of the other cases that I have spotted so far are step-right related.

lanctot:
Thanks for the quick fix @alexminnaar !

lanctot:
Fixed by https://github.com/deepmind/open_spiel/pull/227 which was just merged. Closing.

PR #227, opened 2020-05-23
Laser tag fix

alexminnaar:
`kStepLeft` is mistakenly checked twice within the if statment - one of them should be `kStepRight`.  Also included the updated playthrough.


alexminnaar:
Fixing the bug from https://github.com/deepmind/open_spiel/issues/224.  I got a bunch of merge conflicts from the new playthrough.  To generate the playthrough I ran `./open_spiel/scripts/generate_new_playthrough.sh 'laser_tag(horizon=20)'` and the resulting file looks a lot different than what was on master so let me know if there is a better way of generating this playthrough.

elkhrt:
Thanks! It's cleaner to run regenerate_playthroughs.sh

lanctot:
Thanks for the fix @alexminnaar !

> Thanks! It's cleaner to run regenerate_playthroughs.sh

Will do this in when I import it, so no need to do it on your side.

Issue #228, opened 2020-05-25
Breakthrough Swift implementation is slow


sbodenstein:
Playing random games with Swift environments is generally faster than the combination of Python/C++. But Breakthrough is significantly slower for some reason. Times below are in μs:
|  | **Swift** | **Python** |
| :---         |     :---:      |          ---: |
| TicTacToe  | 87     | 255    |
| Breakthrough     | 9340      | 2506      |
| Kuhn Poker     | 15       | 149      |

These numbers were generated on a 2019 15' Macbook Pro, and the Swift implementation was built with `-O` and `Whole Module` optimization flags:
```swift
import Dispatch
import Foundation
import OpenSpiel

func playRandomGame<T: GameProtocol>(_ game: T) {
  var state = game.initialState
  while !state.isTerminal {
    state.apply(state.legalActions.randomElement()!)
  }
}

let game = Breakthrough()
let n: Int = 1000
let start = DispatchTime.now()
for _ in 1...n {
    playRandomGame(game)
}
let timeDiff = Double(DispatchTime.now().uptimeNanoseconds - start.uptimeNanoseconds) / Double(n)
print("Mean Random Game Time (μs): \(timeDiff / 1e3)")
```
and 
```python
import numpy as np
import pyspiel
import time

def play_random_game(game):
    game_state = game.new_initial_state()
    while not game_state.is_terminal():
        random_action = np.random.choice(game_state.legal_actions())
        game_state.apply_action(random_action)

game = pyspiel.load_game("breakthrough")
n = 1000
start = time.time()
for _ in range(n):
    play_random_game(game)
time_diff = (time.time() - start) / n

print("Mean Random Game Time (μs): ", 1e6 * time_diff)
```

lanctot:
Hi @sbodenstein,

I'm not too surprised by this, but I don't know much about Swift. @elkhrt or @dan-zheng , any ideas?

The reason for my thinking: the python games are all wrappers around C++ implementations, not actual Python implementations. So, the constant overhead you pay for the round trip from python to C++ and back should eventually be less of a problem as the game gets larger because the time spent executing the logic of the rules dominates the running time. 

Now, I'm not sure that's what is happening here-- I actually have no idea what the relative latencies are like-- but it is one theory.

sbodenstein:
[...]

PR #230, opened 2020-05-27
Improve Swift Breakthrough Performance

saeta:
This PR defines some benchmarks for Swift OpenSpiel, and modestly improves Swift's `Breakthrough` performance.

On my machine, performance improves by ~1.65x based on the benchmarks:

Before:

```
name                      time        std        iterations
-----------------------------------------------------------
random game: Breakthrough 173380.0 ns ±  24.30 %       7276
```

After:

```
name                      time        std        iterations
-----------------------------------------------------------
random game: Breakthrough 104894.5 ns ±  24.74 %      13624
```

I suspect future optimizations are quite possible. In particular, I suspect flattening out `State.board` from being: `Array<Optional<BreakthroughPlayer>>` to `Array<FlattenedEnum>` might help, but have not tried that yet.

Issue: https://github.com/deepmind/open_spiel/issues/228

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ cmake-build-*/
 dist/
 pyspiel.egg-info/
 
+# Swift build directory
+.build
+
  # External git modules
 open_spiel/abseil-cpp/
 open_spiel/games/bridge/double_dummy_solver/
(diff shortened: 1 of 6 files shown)


elkhrt:
Thanks for the update, Brennan.
The Package.resolved file seems to be causing a problem for the git -> piper import. Could you remove it (maybe add it to .gitignore?)
Alternatively, if you can fix the import problem, that would be good too!

saeta:
Sorry for the delay; only getting to it now. I've sent a Piper CL that I think (🤞) should allow this PR to go through smoothly.

PR #237, opened 2020-06-07
goofspiel: fix current point card

ekarais:
When returning the current point card, the methods `ObservationString` and `ObservationTensor` look at the index of the current point card in a data structure that contains only the remaining cards, and increment it. The returned value does not always correspond to the actual value of the current point card however, especially after the first couple of moves.

One way to fix that is what I did: `ObservationString` and `ObservationTensor` now access the last element of the same vector that is also accessed by `InformationStateString` and `InformationStateTensor,` which contains the values of the point card sequence.

It is my first time contributing to a public repo. I'm sorry if I unknowingly breached some guidelines or rules.

### DIFF
--- a/open_spiel/games/goofspiel.cc
+++ b/open_spiel/games/goofspiel.cc
@@ -432,8 +432,15 @@ std::string GoofspielState::ObservationString(Player player) const {
   //   - everyone's current points
   //   - my current hand
   //   - current win sequence
+  int curr_point_card_;
+  if (!point_card_sequence_.empty()) {
+	curr_point_card_ = point_card_sequence_.back();
+  }
+  else {
+	curr_point_card_ = 0;
+  }
   std::string current_trick =
-      absl::StrCat("Current point card: ", point_card_sequence_.back());
+      absl::StrCat("Current point card: ", curr_point_card_);
   std::string points_line = "Points: ";
   std::string hands = "";
   std::string win_seq = "Win Sequence: ";
@@ -576,9 +583,18 @@ void GoofspielState::ObservationTensor(Player player,
   //   - current win sequence
 
   // Current point card.
-  for (int i = 0; i < num_cards_; ++i) {
-    values->push_back(i == point_card_sequence_.back() ? 1.0 : 0.0);
+  int curr_point_card_;
+  if (!point_card_sequence_.empty()) {
+	curr_point_card_ = point_card_sequence_.back();
   }
+  else {
+	curr_point_card_ = 0;
+  }
+	
+  for (int i = 0; i < num_cards_; ++i) {
+	values->push_back(i == (curr_point_card_ - 1) ? 1.0 : 0.0);
+  }
+
 
   // Point totals: one-hot vector encoding points, per player.
   Player p = player;
--- a/open_spiel/games/goofspiel.cc
+++ b/open_spiel/games/goofspiel.cc
@@ -433,7 +433,7 @@ std::string GoofspielState::ObservationString(Player player) const {
   //   - my current hand
   //   - current win sequence
   std::string current_trick =
-      absl::StrCat("Current point card: ", point_card_index_ + 1);
+      absl::StrCat("Current point card: ", point_card_sequence_.back());
   std::string points_line = "Points: ";
   std::string hands = "";
   std::string win_seq = "Win Sequence: ";
@@ -577,7 +577,7 @@ void GoofspielState::ObservationTensor(Player player,
 
   // Current point card.
   for (int i = 0; i < num_cards_; ++i) {
-    values->push_back(i == point_card_index_ ? 1.0 : 0.0);
+    values->push_back(i == point_card_sequence_.back() ? 1.0 : 0.0);
   }
 
   // Point totals: one-hot vector encoding points, per player.


elkhrt:
Thanks for the contribution!
I think the new code is equivalent to the old code (but with a memory bug). Can you give an example where this change makes a difference? Either as a play-through or as a unit test.

open_spiel/games/goofspiel.cc
```diff
@@ -577,7 +577,7 @@ void GoofspielState::ObservationTensor(Player player,
 
   // Current point card.
   for (int i = 0; i < num_cards_; ++i) {
-    values->push_back(i == point_card_index_ ? 1.0 : 0.0);
+    values->push_back(i == point_card_sequence_.back() ? 1.0 : 0.0);
```
elkhrt, line 580: This is not safe if point_card_sequence is empty, which it will be at the start of the game.

lanctot:
@elkhrt f this is not a bug then I think there is one in `InformationStateString` (and maybe tensor) which currently assumes point_card_sequence contains the current one AFAICT but only took a quick look.(If so, my bad)

lanctot:
Definitely think it's better/cleaner if we do not make that assumption. I can fix it this week if you don't beat me to it. :)

lanctot:
Also I still think this is a legit bug for the random ordering. It should index into the point card deck.

open_spiel/games/goofspiel.cc
Batom3: Isn't the numbers of the cards are from 1 to 13 in the point_card_sequence? This would be a misinformation regarding which card has been played on the current trick in terms of ObervationTensor?

ekarais: @elkhrt You are right about it being unsafe. I addressed this issue and the other issue raised by @Batom3 in the subsequent commit.

ekarais:
@elkhrt Regarding what difference the new code makes: Here is an isolated example that demonstrates the issue:

### Old behavior:

_Number of opened cards = 0_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: 
Points: 0 0 

_Output of state.observation_string(0):_ 
Current point card: 0
Points: 0 0 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 

_Number of opened cards = 1_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: **5** 
Points: 0 0 

_Output of state.observation_string(0):_ 
Current point card: **5**
Points: 0 0 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 

_Number of opened cards = 2_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 6 7 8 9 10 11 12 13 
P1 hand: 1 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: 5 **13** 
Points: 5 0 

_Output of state.observation_string(0):_ 
Current point card: **12(false!)**
Points: 5 0 
P0 hand: 1 2 3 4 6 7 8 9 10 11 12 13 
P1 hand: 1 3 4 5 6 7 8 9 10 11 12 13 

### New behavior (achieved with the 2nd commit):

_Number of opened cards = 0_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: 
Points: 0 0 

_Output of state.observation_string(0):_ 
Current point card: 0
Points: 0 0 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 

_Number of opened cards = 1_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: **5** 
Points: 0 0 

_Output of state.observation_string(0):_ 
Current point card: **5**
Points: 0 0 
P0 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 
P1 hand: 1 2 3 4 5 6 7 8 9 10 11 12 13 

_Number of opened cards = 2_

_Output of state.information_state_string(0):_ 
P0 hand: 1 2 3 4 6 7 8 9 10 11 12 13 
P1 hand: 1 3 4 5 6 7 8 9 10 11 12 13 
Point card sequence: 5 **13** 
Points: 5 0 

_Output of state.observation_string(0):_ 
Current point card: **13(correct)**
Points: 5 0 
P0 hand: 1 2 3 4 6 7 8 9 10 11 12 13 
P1 hand: 1 3 4 5 6 7 8 9 10 11 12 13

PR #240, opened 2020-06-09
Add python bindings for MCCFR solvers

inejc:
It is a bit cumbersome to expose the second constructor for both types since they accept a shared pointer to a custom default `Policy` instance and the default binding holder type for `Policy` is a unique pointer. Changing the holder type to a shared pointer works but breaks at least all CFR solvers `AveragePolicy()` bindings since unique pointers are returned. For this reason, I didn't add bindings for these two constructors and it is thus only possible to construct instances with the default `UniformPolicy` within python. Please let me know if there is a known, simpler solution for this.


lanctot:
Actually I can fix the comment upon the import so no need.

Issue #234, opened 2020-06-05
Example using a trained model/policy


aliyakamercan:
Hi,

I am trying to run Kuhn poker playthrough using different policies (NSFP and CFR from examples), but I could not find a way to save these. Pickling doesn't seem to work. Is there some other method to save/load these?

Thanks

lanctot:
Hi, can I double-check that you are doing this all in python? I think I know why the pickling does not work, and there is an easy fix (convert the policies to tabular policies and pickle those). We don't have an example but it would be easy to add. I will wait for your reply first.

aliyakamercan:
I am using python.

Here are two examples:

```python
# 1
random_policy = policy.UniformRandomPolicy(game)
print(pickle.dumps(random_policy))

# 2
cfr_solver = cfr.CFRSolver(game)
iterations = 1000

for i in range(iterations):
  cfr_value = cfr_solver.evaluate_and_update_policy()
  print("Game util at iteration {}: {}".format(i, cfr_value))

cfr_policy = cfr_solver.average_policy()
print(pickle.dumps(random_policy))
```

Both pickles are failing with
```
can't pickle pyspiel.GameType objects
```

I think, it is because there is no `py::pickle` for `GameState` in [pyspiel.cc](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/pybind11/pyspiel.cc)

lanctot:
That is odd because we definitely support pickling states, we even test it for all games in the simulation tests:

 https://github.com/deepmind/open_spiel/blob/0e9b1b75dea3e87f0326fe1012d1b0367196b7dc/open_spiel/python/tests/games_sim_test.py#L98

So something seems very wrong indeed. I will look into it. Did you perhaps add something to GameType that does not serialize?

lanctot:
And the `py::pickle` is here:

inejc:
I might be missing something here but I don't think there is support for serialization of any of the policy types currently, i.e. it is not possible to pickle `TabularPolicy` or `CFRAveragePolicy` which I believe the above python code tries to do.

lanctot:
Oh I think you meant `GameType` which is missing the `py::pickle`. 

Yes it is probably right that we can't pickle tabular policies right now but these are simple objects. We should support pickling them and I think the fix is easy (or should be).

inejc:
[...]

PR #241, opened 2020-06-09
serialize / deserialize GameType and GameParameters

aliyakamercan:
addresses #234


lanctot:
Very nice implementation, thanks!

I have now imported it.. should get merged in the next update.

I made a few cosmetic changes, only one real change is replacing regex with absl::StrReplaceAll, as regex is an unapproved C++ header (hence against our style guide).

Thanks again, this is a great addition, and kudos on the timely submission!

PR #247, opened 2020-06-17
Fix bug in get_weights when using a2c loss. 

buregab:
Calling get_weights() when using the a2c loss in policy_gradient currently throws: "AttributeError: 'PolicyGradient' object has no attribute '_q_values_layer'". This is because there is no _q_values_layer when using a2c loss, this is instead replaced with _baseline_layer. Updated the get_weights() function to check which loss is being used and append the appropriate weights. Also, added a2c loss into the enum in the kuhn_policy_gradient example.


lanctot:
Thanks! @fzvinicius , can you take a look and import when satisfied?

lanctot:
@fzvinicius this seems pretty straight-forward: I will import it and add you as a reviewer.

PR #254, opened 2020-06-24
Hearts updates

jhtschultz:
Some assorted updates including an important bug fix in InformationStateTensor().


lanctot:
Hey John can you update requirements.txt as in this commit: https://github.com/deepmind/open_spiel/commit/63ede7dee69c79510ff99dac99c68b8e368f6a9a  , it should fix the travis test, then I will import

PR #259, opened 2020-06-25
Added Lewis Signaling Game

vbhatt-cs:
Added a simple Lewis Signaling Game. The sender receives a random private state from the environment and sends a message to the receiver. The receiver takes an action based on the message. Both agents get a reward based on the state and the action. Supports arbitrary payoff matrices.


vbhatt-cs:
@googlebot I signed it!

Issue #264, opened 2020-06-26
DeepCFR with Reservoir sampling 


bakanaouji:
Dear authors.

I have been reading the code of DeepCFR and was wondering why a reservoir sampling is not chosen?
https://github.com/deepmind/open_spiel/blob/021f3585efed8bb9eb4173e36f65c65263996c69/open_spiel/python/algorithms/deep_cfr.py#L62-L76

According to the following paper, the performance with a reservoir sampling is better than one with a sliding window memory.
https://arxiv.org/pdf/1811.00164.pdf
From this point of view, I think it is better to choose a reservoir sampling memory.

lanctot:
Hi @bakanaouji yes I think that is a bug... or, rather, something the author did not realize about the algorithm on reading the paper.

Do you feel like contributing a fix? There is reservoir sampling code you can take directly from NFSP. Maybe we should make that data structure more accessible generally, a lot of algorithms in this class are starting to use them now. But I'd be happy with an implementation local to this file.

If you don't, I will fix this, but I've got several higher-priority items on the list, so it might be a few weeks.

bakanaouji:
Hi @lanctot , thanks for your polite response.

I would like to contribute to the fix this bug. Since I have already confirmed that the performance with the fixed code will improve in my local environment, I can send the pull request immediately.

lanctot:
Ok great. Can you make sure it's the most reason version using our simple_nets not sonnet?

Also can you try the example to see if you get close to the proper values in Kuhn? I had to introduce a flag to toggle reinitializing the networks and had to set it to false to get good values. Curious if that is still true with the reservoir buffers.

Details here: https://github.com/deepmind/open_spiel/pull/248

bakanaouji:
I have checked with both versions using simple_nets and sonnet.
The results of trying the example are shown in the following PR.
#268

lanctot:
Great thanks. @aliyakamercan tagging you since you were also looking into this code.

lanctot:
Fixed now by https://github.com/deepmind/open_spiel/commit/f67f3e65b1c66fe3a2291cf555957e9378a29456

PR #268, opened 2020-06-28
Change data structure of Deep CFR to reservoir buffer

bakanaouji:
This PR addresses #264.

I changed the data structure of Deep CFR from a ring buffer to a reservoir buffer.

In order to check whether the performance with a reservoir buffer can be improved, I ran the example code with the following settings.
* policy_network_epochs: 400
* advantage_network_epochs: 20
* memory_capacity: 1e4

The results of the three trials with a reservoir buffer are:
```
NashConv: 0.15078695548342402 / 0.2201359770839587 / 0.18288091484563324
Computed player 0 value: -0.05145201795586418 / -0.05499330024598259 / -0.052542900906346945
Expected player 0 value: -0.05555555555555555
Computed player 1 value: 0.05145201795586418 / 0.05499330024598259 / 0.052542900906346945
Expected player 1 value: 0.05555555555555555
```

The results of the three trials with a ring buffer (original) are:
```
NashConv: 0.3581176702143096 / 0.30669551399644407 / 0.48862825304358304
Computed player 0 value: -0.05807950847105664 / 0.004885766864908492 / -0.09493437445965347
Expected player 0 value: -0.05555555555555555
Computed player 1 value: 0.05807950847105664 / -0.004885766864908492 / 0.09493437445965347
Expected player 1 value: 0.05555555555555555
```
I ran the example code in the version of b65db20bffe0cc45c4bbb768947626693e058de1.

By the way, we have to set `reinitialize_advantage_networks` to false to get good values...
Maybe, setting `advantage_network_train_steps` to is not enough for learning from scratch.

### DIFF
--- a/open_spiel/python/algorithms/deep_cfr.py
+++ b/open_spiel/python/algorithms/deep_cfr.py
@@ -18,7 +18,7 @@ See https://arxiv.org/abs/1811.00164.
 
 The algorithm defines an `advantage` and `strategy` networks that compute
 advantages used to do regret matching across information sets and to approximate
-the strategy profiles of the game. To train these networks a fixed ring buffer
+the strategy profiles of the game. To train these networks a reservoir buffer
 (other data structures may be used) memory is used to accumulate samples to
 train the networks.
 """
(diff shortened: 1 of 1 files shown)


bakanaouji:
@googlebot I signed it!

lanctot:
Excellent, thanks! And thank you for checking out the numbers. 

Yes, I agree, probably with such small values in the example you just can't afford to reset the advantage networks, but on larger games and higher values it might be significantly better. So at least we have the option now.

PR #271, opened 2020-06-29
NFSP updates

jhtschultz:
Adds support for more than 2 players in the leduc_nfsp.py example and checkpointing for NFSP policies. 

NFSP converges in 3p Kuhn poker. 

Checkpoints load successfully. After restoring checkpoints and evaluating the agents' joint policy, the NashConv was the same as the most recent saved policy.


lanctot:
Hey @jhtschultz this is odd, it's saying I can't import it because it's missing the CLA. But then I don't see it mentioned anywhere here.

PR #273, opened 2020-06-30
Implement Public State API for Kuhn Poker 

michalsustr:


open_spiel/spiel.cc
```diff
@@ -536,6 +536,10 @@ std::ostream& operator<<(std::ostream& stream, GameType::ChanceMode value) {
   }
 }
 
+std::ostream& operator<<(std::ostream& stream, const State& state) {
+  return stream << state.History();
```
lanctot, line 540: Why not state.ToString() ?

open_spiel/spiel.h
```diff
@@ -51,11 +51,29 @@ enum PlayerId {
 // Constant representing an invalid action.
 inline constexpr Action kInvalidAction = -1;
 
-// Constant that signifies that player did not receive any private observation.
```
lanctot, line 54: This is taking up a fair bit of space in spiel.h, can we move to a new file, like `fog_constants.h` which we can include here? 

I'd like to keep all the FOG-related stuff contained since it's only relevant if you're using FOGs stuff, whereas this is mostly a spot to put more widely used stuff.

Issue #276, opened 2020-07-01
Better de/serialization for stateful game objects


inejc:
Strictly speaking, I believe stateful game objects are currently not serialized correctly. For example, `NegotiationGame` which holds an internal RNG loses the state of the RNG after serialization and deserialization (using `ToString` and `LoadGame`). I guess this issue is mostly relevant for sampled stochastic implementations, e.g. I have the same problem in https://github.com/deepmind/open_spiel/pull/274.

While there is the ability to override `Serialize` and `Deserialize` for state objects, the same should probably be true for game objects. The default implementation would assume stateless game objects and would simply be identical to the current implementation.

The issue occurs in cases where the `rng_seed` parameter is set by a user and then the game and state objects are de/serialized. While the state object is deserialized correctly, a `NewInitialState` call on the deserialized game object, would essentially result in the same sequence of RNG samples and thus the same sequence of new initial states (compared to prior to serialization).

lanctot:
Hi @inejc , ah yes I understand the issue. 

But, I think we should probably not encourage stateful objects generally, could be error-prone. We can allow it for this specific exceptional case of `kSampledStochastic` chance mode and the RNG state. 

I think there is a simple fix which is to include the current seed (or whatever is needed to reconstruct the RNG state) in the game's parameters and invlude it in game's ToString so it gets deserialized properly.

We should also add a test for all the games using that chance mode to ensure that the produce the same sequence of random numbers after serialization.

What do you think?

inejc:
Agree with not encouraging stateful game objects in general. Yeah, I think that fix would work, I will give it a try later in the day.

lanctot:
Closing this since it's actively being working on in https://github.com/deepmind/open_spiel/pull/279

PR #279, opened 2020-07-02
Fix sampled stochastic games serialization

inejc:
This PR addresses https://github.com/deepmind/open_spiel/issues/276.

### DIFF
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -446,5 +446,14 @@ std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
       num_redeals_);
 }
 
+std::string UncontestedBiddingGame::GetRNGState() const {
+  return std::to_string(rng_seed_);
+}
+
+void UncontestedBiddingGame::SetRNGState(const std::string& rng_state) const {
+  if (rng_state == "") return;
+  rng_seed_ = std::stoi(rng_state);
+}
+
 }  // namespace bridge_uncontested_bidding
 }  // namespace open_spiel
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -391,22 +391,14 @@ UncontestedBiddingGame::UncontestedBiddingGame(const GameParameters& params)
   }
 }
 
-// Override ToString() so that internal RNG state is also peristed.
-std::string UncontestedBiddingGame::ToString() const {
-  GameParameters params = game_parameters_;
-  params["name"] = GameParameter(game_type_.short_name);
-  params["rng_seed"] = GameParameter(rng_seed_);
-  return GameParametersToString(params);
-}
-
 // Deserialize the deal and auction
 // e.g. "AKQJ.543.QJ8.T92 97532.A2.9.QJ853 2N-3C"
 std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
     const std::string& str) const {
   if (str.empty()) {
-    return std::make_unique<UncontestedBiddingState>(
+    return absl::make_unique<UncontestedBiddingState>(
         shared_from_this(), reference_contracts_, deal_filter_, forced_actions_,
-        rng_seed_);
+        rng_seed_, num_redeals_);
   }
   SPIEL_CHECK_GE(str.length(),
                  kNumPlayers * (kNumCardsPerHand + kNumSuits) - 1);
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -451,8 +451,8 @@ std::string UncontestedBiddingGame::GetRNGState() const {
 }
 
 void UncontestedBiddingGame::SetRNGState(const std::string& rng_state) const {
-  if (rng_state == "") return;
-  rng_seed_ = std::stoi(rng_state);
+  if (rng_state.empty()) return;
+  absl::SimpleAtoi(rng_state, &rng_seed_);
 }
 
 }  // namespace bridge_uncontested_bidding
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -446,8 +446,8 @@ std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
   }
 
   return absl::make_unique<UncontestedBiddingState>(
-      shared_from_this(), reference_contracts_, Deal(cards), actions,
-      ++rng_seed_, num_redeals_);
+      shared_from_this(), reference_contracts_, Deal(cards), actions, rng_seed_,
+      num_redeals_);
 }
 
 }  // namespace bridge_uncontested_bidding
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -404,9 +404,9 @@ std::string UncontestedBiddingGame::ToString() const {
 std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
     const std::string& str) const {
   if (str.empty()) {
-    return std::unique_ptr<State>(
-        new UncontestedBiddingState(shared_from_this(), reference_contracts_,
-                                    deal_filter_, forced_actions_, rng_seed_));
+    return std::make_unique<UncontestedBiddingState>(
+        shared_from_this(), reference_contracts_, deal_filter_, forced_actions_,
+        rng_seed_);
   }
   SPIEL_CHECK_GE(str.length(),
                  kNumPlayers * (kNumCardsPerHand + kNumSuits) - 1);
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -17,9 +17,9 @@
 #include <cstring>
 #include <memory>
 
-#include "open_spiel/games/bridge/double_dummy_solver/include/dll.h"
 #include "open_spiel/game_parameters.h"
 #include "open_spiel/games/bridge/bridge_scoring.h"
+#include "open_spiel/games/bridge/double_dummy_solver/include/dll.h"
 #include "open_spiel/spiel.h"
 #include "open_spiel/spiel_utils.h"
 
@@ -389,6 +389,14 @@ UncontestedBiddingGame::UncontestedBiddingGame(const GameParameters& params)
           {7, kNoTrump, kUndoubled, 1}};
     }
   }
+}
+
+// Override ToString() so that internal RNG state is also peristed.
+std::string UncontestedBiddingGame::ToString() const {
+  GameParameters params = game_parameters_;
+  params["name"] = GameParameter(game_type_.short_name);
+  params["rng_seed"] = GameParameter(rng_seed_);
+  return GameParametersToString(params);
 }
 
 // Deserialize the deal and auction
--- a/open_spiel/games/bridge_uncontested_bidding.cc
+++ b/open_spiel/games/bridge_uncontested_bidding.cc
@@ -403,7 +403,11 @@ std::string UncontestedBiddingGame::ToString() const {
 // e.g. "AKQJ.543.QJ8.T92 97532.A2.9.QJ853 2N-3C"
 std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
     const std::string& str) const {
-  if (str.empty()) return NewInitialState();
+  if (str.empty()) {
+    return std::unique_ptr<State>(
+        new UncontestedBiddingState(shared_from_this(), reference_contracts_,
+                                    deal_filter_, forced_actions_, rng_seed_));
+  }
   SPIEL_CHECK_GE(str.length(),
                  kNumPlayers * (kNumCardsPerHand + kNumSuits) - 1);
   std::array<int, kNumCards> cards{};
--- a/open_spiel/games/negotiation.cc
+++ b/open_spiel/games/negotiation.cc
@@ -587,9 +587,9 @@ std::unique_ptr<State> NegotiationGame::DeserializeState(
     // Take the chance action, but then reset the quantities. Make sure game's
     // RNG state is not advanced during deserialization so copy it beforehand
     // in order to be able to restore after the chance action.
-    std::mt19937* rng = new std::mt19937(*rng_);
+    std::unique_ptr<std::mt19937> rng = std::make_unique<std::mt19937>(*rng_);
     nstate.ApplyAction(0);
-    rng_.reset(rng);
+    rng_ = std::move(rng);
     nstate.ItemPool().clear();
     nstate.AgentUtils().clear();
     // Max steps
@@ -628,7 +628,7 @@ std::string NegotiationGame::GetRNGState() const {
 }
 
 void NegotiationGame::SetRNGState(const std::string& rng_state) const {
-  if (rng_state == "") return;
+  if (rng_state.empty()) return;
   std::istringstream rng_stream(rng_state);
   rng_stream >> *rng_;
 }
--- a/open_spiel/games/negotiation.cc
+++ b/open_spiel/games/negotiation.cc
@@ -602,8 +602,12 @@ std::unique_ptr<State> NegotiationGame::DeserializeState(
     std::unique_ptr<State> state = NewInitialState();
     SPIEL_CHECK_EQ(lines.size(), 5);
     NegotiationState& nstate = static_cast<NegotiationState&>(*state);
-    // Take the chance action, but then reset the quantities.
+    // Take the chance action, but then reset the quantities. Make sure game's
+    // RNG state is not advanced during deserialization so copy it beforehand
+    // in order to be able to restore after the chance action.
+    std::mt19937* rng = new std::mt19937(*rng_);
     nstate.ApplyAction(0);
+    rng_.reset(rng);
     nstate.ItemPool().clear();
     nstate.AgentUtils().clear();
     // Max steps
--- a/open_spiel/games/negotiation.cc
+++ b/open_spiel/games/negotiation.cc
@@ -55,6 +55,7 @@ const GameType kGameType{
      {"num_items", GameParameter(kDefaultNumItems)},
      {"num_symbols", GameParameter(kDefaultNumSymbols)},
      {"rng_seed", GameParameter(kDefaultSeed)},
+     {"rng_state", GameParameter(std::string(kDefaultRNGState))},
      {"utterance_dim", GameParameter(kDefaultUtteranceDim)}}};
 
 static std::shared_ptr<const Game> Factory(const GameParameters& params) {
@@ -508,6 +509,13 @@ NegotiationGame::NegotiationGame(const GameParameters& params)
       seed_(ParameterValue<int>("rng_seed", kDefaultSeed)),
       legal_utterances_({}),
       rng_(new std::mt19937(seed_ >= 0 ? seed_ : std::mt19937::default_seed)) {
+  // Internal RNG state is set when serializing the game instance with
+  // NegotiationGame::ToString() so we try to deserialize it here.
+  std::string rng_state = ParameterValue<std::string>("rng_state");
+  if (rng_state != kDefaultRNGState) {
+    std::istringstream rng_stream(rng_state);
+    rng_stream >> *rng_;
+  }
   ConstructLegalUtterances();
 }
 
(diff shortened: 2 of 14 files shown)


inejc:
@lanctot please let me know if this looks like what you had in mind? There are now failing (existing) tests that I need to fix, I suspect there two distinct problems:
- existing game serialization tests should probably skip / treat sampled stochastic games differently
- I suspect overridden `Game::ToString` is not called in some of the tests due to usage of references rather than pointers to game objects (i.e. the method is not called polymorphically)

Anyway, please let me know what you think.

lanctot:
>     * existing game serialization tests should probably skip / treat sampled stochastic games differently

Hmm, why? I'm open to that, as long as they get tested as thoroughly. 
 
>     * I suspect overridden `Game::ToString` is not called in some of the tests due to usage of references rather than pointers to game objects (i.e. the method is not called polymorphically)

Wait... really? The very foundation of my C++ knowledge shaken (this has been known to happen on occasion).......

Ok, experiment conducted... sanity restored! This code prints out `Hi-B` so I am pretty sure that function _should_ be called polymorphically even through references:
```
#include <iostream>

class A {
 public:
  virtual void func() const { std::cout << "Hi-A" << std::endl; }
};

class B : public A {
 public:
  void func() const override { std::cout << "Hi-B" << std::endl; }
};

void doit(const A& a) {
  a.func();
}

int main() {
  B b;
  doit(b);
}
```

Maybe `Game::ToString` is (acidentally) not virtual? But no, then none of the game's ToStrings would work.

inejc:
I should've closed the computer much earlier yesterday :). Anyway, the problem comes down to `Game::DeserializeState` mutating the internal RNG state. The fix was quite easy as sampled stochastic games already have to override the method.

All tests are passing now, except that playthroughs for negotiation and bridge uncontested bidding changed (and not just parameter specification). Could this be related to the fact that the two games are sampled stochastic (or anything related) as I can't seem to find the proposed changes that would affect the actual playthroughs? And the game and state objects are not de/serialized during playthrough generation as far as I can see.

> Hmm, why? I'm open to that, as long as they get tested as thoroughly.

This is not needed after all (but I think stochastic games are tested at least with game_sim etc.).

open_spiel/games/bridge_uncontested_bidding.cc
```diff
@@ -387,11 +387,23 @@ UncontestedBiddingGame::UncontestedBiddingGame(const GameParameters& params)
...
+  return GameParametersToString(params);
+}
+
 // Deserialize the deal and auction
 // e.g. "AKQJ.543.QJ8.T92 97532.A2.9.QJ853 2N-3C"
 std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
     const std::string& str) const {
-  if (str.empty()) return NewInitialState();
+  if (str.empty()) {
+    return std::unique_ptr<State>(
```
finbarrtimbers, line 403: Can you use `absl::make_unique` (or `std::make_unique`) instead?

open_spiel/python/tests/sampled_stochastic_games_test.py
```diff
@@ -0,0 +1,62 @@
...
+]
+assert len(SPIEL_SAMPLED_STOCHASTIC_GAMES_LIST) >= 2
+
+
+class SampledStochasticGamesTest(parameterized.TestCase):
+    def random_playout(self, state):
+        np.random.seed(0)
+        while not state.is_terminal():
+            state.apply_action(np.random.choice(state.legal_actions()))
+        return state
```
finbarrtimbers, line 39: Can you make this a free function? And add it to python/tests/test_utils.py? This is something that should be useful to a bunch of other tests.

Also, can you make seed an argument with a default value of 0?

inejc:
All comments have been addressed. Any ideas why playthrough tests are failing?

lanctot:
Not sure. We will take a look after we import it. Thanks!

finbarrtimbers:
Lgtm- I'll be merging this and you'll get a notification once it's in (probably next Monday).

finbarrtimbers:
Playthrough fails because of adding the rng_seed (which is fine):

```
  File "/usr/local/google/home/finbarrtimbers/tmp/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 73, in test_rerun_playthroughs  
    self.assertEqual(expected_line, actual_line,                                                                                                               
AssertionError: Wrong line 30 in bridge_uncontested_bidding.txt:                                                                                               
- ToString() = "bridge_uncontested_bidding(relative_scoring=True)"                                                                                             
+ ToString() = "bridge_uncontested_bidding(relative_scoring=True,rng_seed=0)"                                                                                  
?                                                               +++++++++++                                                                                    
```

The solution is to regenerate the playthroughs by running [`regenerate_playthroughs.sh`](https://github.com/deepmind/open_spiel/blob/master/open_spiel/scripts/regenerate_playthroughs.sh). I did it this time, but if you could do this in the future, that would be appreciated!

finbarrtimbers:
The `rng_state` in Negotiation is huge, can you come up with a more concise representation?

```
negotiation(num_symbols=3,rng_seed=100,rng_state=100 854698869 2051883563 537498773 2958973901 3942973360 3887377573 3677811589 363514342 21427
48103 1886836008 387189304 3152543780 1403413771 134100736 3015357199 869636401 2660739430 2369280134 4197814311 2185306696 1729241415 3470759604 2564777290 38
698368 1486947225 1536869138 1341727130 797128003 2937708172 3945843748 4036127362 2588544517 3605678052 2520382789 777841702 2182316322 3383497157 2865756484 
157142725 2836201697 1909445816 4242861607 4240390239 1730050168 1422739690 1936754661 2874843427 1860385589 3904974773 1287272704 3909158552 204312155 2007293
212 3056255399 2818749264 977641874 1139663059 3622311188 2843386190 4140257080 346412420 1800200530 2804895998 3471167404 513316684 3997130046 814771796 38977
19656 1350825084 3444660887 1956507819 3569413466 949044582 4248696712 2075725090 4160678427 3380687813 2534957932 3386316981 856696350 2090276903 3029523792 2
074096557 2072997680 946804650 354701416 3644023135 2850268580 1778397399 2316290760 2940673037 1299913031 2868997883 1354959259 2018260001 718647552 419943777
 3538884775 151026967 4022577527 3758944041 2470303224 2159949321 3024369087 3554040058 3288161191 2591812639 4111117533 3318415875 3140078190 1304733963 32937
76738 562128566 251587648 3435230643 3357449956 594497688 3778290798 3154883704 3683221146 875418582 793516520 4164641027 1565934972 2695122894 2384177146 3620
85719 3770450643 3759205521 2235922460 2994893401 3240546667 2486781581 1665424881 2446229559 3867669105 1837083011 4288836308 2725076574 19494616 98401221 144
2923079 4074806317 3010739382 2588166037 274796069 1709138476 1743180117 2908916409 3543912285 2888048301 3916743331 1771596217 3183291954 2628446603 196552004
1 1721167589 4154475666 2195957460 1341348110 2299800204 4053584040 3672878874 3579803265 1520653295 3722214284 990201618 2778442946 1595795561 3271842226 1687
035008 3973662865 1301136455 1558308428 1366954512 592132709 1672550282 1665445257 162414427 824423067 3286723292 580354481 4060406924 2921837603 3455498942 59
4661963 1233066834 3356111739 1775863573 1302855586 1284183822 2543853739 3024508014 2715884382 3841460239 1518180736 2302319018 2482079758 1751980419 54104142
6 3571316963 333648170 2364227677 1423719751 2358813547 3951733307 1783963879 1887113614 198997308 760962430 2156286345 1970180011 623290599 1794458361 3898746
031 3305931956 1173834764 2191455995 1739698712 1813620153 2278939253 915005649 2674691156 457663694 2713650215 3264704635 2218235963 536121441 1533534250 3961
484509 3624266365 4131713694 3981753818 3103197319 2225637988 1499017514 2007853540 2254001479 1086558760 2172623901 3133026348 773038360 916219755 922358059 5
37250284 3503027986 1965527980 1818220857 2643962897 275910521 4288076984 2143431107 1279347847 2808145884 3831712149 2292666158 764725597 1493252019 164673849
3 2925191344 2834819391 1917561623 2349024949 555996731 6691152 1023659418 2998765773 78037431 2474007872 3932757528 4082259126 3150137721 2092683672 7574383 3
817378526 461715269 3305487438 1864597623 603641253 3788412977 3880963795 4228442410 3765038920 1776782515 2743321943 1223317671 2725236893 3038471899 20431968
62 28044413 3721278324 2015755799 4136699347 3422788662 225967888 1095185016 1214819302 3164303949 229204310 4198788634 3935949322 1784000443 351199121 1624886
885 2585653925 3946650645 2411584481 399064259 20862244 1562631274 453524078 310766750 449374607 3442126245 2421053881 889944835 3010536556 3206823588 23717188
45 1384548763 1240315139 733248780 1349829375 3076316538 3094387101 1308341249 1671562567 2963835878 3902289469 3704001984 3167973946 1729366372 1102168870 360
0998065 105717641 1795998557 941183133 510383171 3338115778 2831702137 499236060 1703231522 3789163046 2910244593 826585656 3003271538 457435787 672984371 1321
132412 509992111 1578285418 1485466519 994687375 1061993933 3077871684 2020051714 4216550548 3044200441 2608226926 4048855556 3813322028 862813173 2704999956 2
770677274 1878919397 2419558242 3506096719 1095269740 4138979954 4255613191 3996978183 2912936712 2111499879 1572775348 3872931552 1227705351 3099080151 339419
0211 3633969915 1490468180 22950150 3515846364 2100409978 3768701703 2934007317 489723797 2254277708 3508542026 867856722 3765541088 1238879766 4128219035 3988
410497 2334480084 2299382265 3714385043 1211883101 1847719898 3423075830 331501625 103229710 1775088152 1025877616 318544836 2673957353 2482257485 484302274 12
6046498 3441964547 4022554522 3604367864 1509917859 943196039 1389574369 3303363071 4114429452 1367170188 1087896131 3656454061 3544272714 1127307890 390943437
3 3898205701 2579167238 765067069 503601595 2212068210 132949212 379242617 1202248299 1399252865 1308415792 2722177286 1970875462 1534163126 1314885863 2438764
67 2245772309 2084941258 3054253519 1345777818 78438881 1983744128 3151685025 1303762444 1319463391 435983445 1448805449 1160460841 3812973450 353923024 891973
332 651767657 636348211 503022566 286990726 3075404199 3371803875 3856264747 2995150740 1027931643 3993012437 894075965 1087754977 1180046385 2652088514 215274
8435 544431369 3784083298 3498791963 1134626639 4136804766 1738656202 2690597361 2441417146 3154343796 344344939 473054229 2190209832 3296619378 3380010614 231
7705995 67036528 2187949844 1844495251 698633344 1916973159 2675978278 139609117 1739810395 2953022829 169883063 3883213344 3010475197 3703048266 2533971901 24
30457676 2866852280 4258835029 167590882 1898091551 3636101068 581121442 3515210978 1126928318 1381363541 1950835231 1786627538 3208771900 3155920244 160234279
7 3810316348 1267261916 1995137843 2974934717 2902803039 539096502 1507743956 3155561232 3860660770 1249673742 3877787893 3370303769 3666025294 2591095918 3701
302442 1368795580 1824497313 1239234353 1226799106 1018754626 2224061470 3559912737 1829300864 2676658172 816561998 1078278623 2757338288 1767849557 2939850048
 746589991 4105362305 446115433 4133126797 460685863 3138665093 805197926 2138142818 266763316 4090681002 833496020 900503756 3362879141 332179368 3406761843 3
628337500 292850600 4082107254 1151204440 282014285 1606750098 2937705009 705837906 2452507534 2475015281 1891755669 4277218971 222215216 835055913 1967528039 
2055223417 504791444 3846526369 1050443048 3201288967 616810553 3438510526 1957385939 2966267165 1717091967 1013356539 3349957453 4066932237 2898662350 3310484
933 848435048 2785142355 4214765889 2627791703 885020631 3653613858 3069949269 2394627236 1118905808 1373095624 4278321313 177140799 3973387825 743474705 17780
57229 862024469 1838917283 1896051269 1046379312 3380825165 3267681060 2734905282 2156355872 676497099 1715393401 1581033659 3109035462 2095424697 1288206078 1
740532738 235092631 2411851516 828062368 1706935179 1112080862 1508316776 201919195 1180532182,utterance_dim=2)
```

lanctot:
Yes this is very large. What is rng_state? Is it the sequence of random numbers produced from the start? 

If so, there must be a better way.. people must serialize RNGs, without requiring memory linear in the sequence. Can be look at the copy constructor of `std::mt19937`?

michalsustr:
Maybe an off-topic, but at some point I added https://github.com/deepmind/open_spiel/blob/master/open_spiel/utils/random.h wrapper. This could support a general serialization function for MT. But maybe you can't use the `Random` wrapper in this case, I didn't look.

lanctot:
Oh now I see you are getting it from `operator>>`. Ok this feels a bit heavy to be in the game parms. I had no idea the `rng_state` would turn out to be this long. Hmm.. I will try to think of something.

lanctot:
Ok maybe we need to add and RNGState section to the serialized string for these types of games rather than do this through game parameters. Sorry @inejc :(

@sustr I would prefer to avoid custom RNGs when the ones in the standard library suffice.

inejc:
A more concise alternative for storing the RNG state into game params might be storing the seed (like in Bridge) and the number of sampled numbers, this should then suffice to recreate the engine by calling http://www.cplusplus.com/reference/random/mersenne_twister_engine/discard.

lanctot:
I was thinking of something like that too but are we 100% certain that every different type of call to the RNG (e.g suppose we want a random double, next we want a random int between certain ranges, etc.) changes the RNG state in the same way? Seems a bit safer and cleaner to just put the full information in the serialized string, doesn't it? @elkhrt any opinion on this?

inejc:
Since the negotiation game already uses the RNG in different ways it does make sense to serialize it properly I believe. Also mixing properties for serialization into other logic seems messy.

finbarrtimbers:
Hmmm. Should we just move the rng to the state? (I can do that.) I'm not sure there is any alternative to serializing the state otherwise as we have no guarantees about how many concurrent states are calling the game.

There are also thread-safety concerns as we don't lock the rng, and it's not thread-safe.

lanctot:
I'm not too worried about threads: each thread should only hold one game, and that game creates and manages all the states in that thread. That's really simple.

On the surface I am not against the RNG in states, the problem is the game would then initialize each RNG with the same seed and you'd get the same sequence of random numbers. The only way I can think to break that is to hold an RNG in the game again and give each state RNG a randomly generated seed. But then that means we still need an RNG in the game.

Right now, I believe the game section of our serialized string only contains the output of Game::ToString() and we pass it directly to LoadGame(). @inejc What's wrong with adding a separate line in that section with the RNG state string for `kSampledStochastic` games, and adding a `Game::SetRNGState(const std::string& rng_state)`?

inejc:
> What's wrong with adding a separate line in that section with the RNG state string for kSampledStochastic games, and adding a Game::SetRNGState(const std::string& rng_state)?

@lanctot that sounds good to me.

lanctot:
Wdyt @finbarrtimbers?

finbarrtimbers:
I believe this is still being worked on- please let me know when you're ready for me to review it.

inejc:
@finbarrtimbers I was a bit busy during the last week but should hopefully get back to this next week and will let you know.

inejc:
@finbarrtimbers the PR is now ready for review. I had to make `Game::SetRNGState` const if it is to be used on return values of `LoadGame`, I wrote comments about this [here](https://github.com/deepmind/open_spiel/blob/5a877c71a2fa636ce5f60bd540bf5f50c6a9bd69/open_spiel/spiel.h#L851) cc @lanctot.

I also had to disable two assertions [here](https://github.com/deepmind/open_spiel/blob/5a877c71a2fa636ce5f60bd540bf5f50c6a9bd69/open_spiel/python/tests/sampled_stochastic_games_test.py#L46) as I believe the Stones and Gems game does not use the RNG with the default parameters.

open_spiel/games/bridge_uncontested_bidding.cc
```diff
@@ -438,8 +442,17 @@ std::unique_ptr<State> UncontestedBiddingGame::DeserializeState(
...
+      shared_from_this(), reference_contracts_, Deal(cards), actions, rng_seed_,
+      num_redeals_);
+}
+
+std::string UncontestedBiddingGame::GetRNGState() const {
+  return std::to_string(rng_seed_);
+}
+
+void UncontestedBiddingGame::SetRNGState(const std::string& rng_state) const {
+  if (rng_state == "") return;
```
finbarrtimbers, line 454: Can you use rng_state.empty()?

open_spiel/games/negotiation.cc
```diff
@@ -584,8 +584,12 @@ std::unique_ptr<State> NegotiationGame::DeserializeState(
     std::unique_ptr<State> state = NewInitialState();
     SPIEL_CHECK_EQ(lines.size(), 5);
     NegotiationState& nstate = static_cast<NegotiationState&>(*state);
-    // Take the chance action, but then reset the quantities.
+    // Take the chance action, but then reset the quantities. Make sure game's
+    // RNG state is not advanced during deserialization so copy it beforehand
+    // in order to be able to restore after the chance action.
+    std::mt19937* rng = new std::mt19937(*rng_);
```
finbarrtimbers, line 590: Can you use unique_ptr here? i.e. rng = absl::make_unique<std::mt19937>(*rng_);
rng_ = std::move(rng);

open_spiel/games/negotiation.cc
```diff
@@ -617,5 +621,17 @@ std::unique_ptr<State> NegotiationGame::DeserializeState(
   }
 }
 
+std::string NegotiationGame::GetRNGState() const {
```
finbarrtimbers, line 624: Can you use StrCat here instead, i.e.:

return absl::StrCat(*rng_);

Using std::ostringstream has caused a lot of performance issues in the past. If this isn't possible, then this is fine.

open_spiel/games/negotiation.cc
```diff
@@ -617,5 +621,17 @@ std::unique_ptr<State> NegotiationGame::DeserializeState(
...
+  std::ostringstream rng_stream;
+  rng_stream << *rng_;
+  return rng_stream.str();
+}
+
+void NegotiationGame::SetRNGState(const std::string& rng_state) const {
+  if (rng_state == "") return;
```
finbarrtimbers, line 631: empty()

open_spiel/games/stones_and_gems.cc
```diff
@@ -1241,11 +1244,20 @@ std::unique_ptr<State> StonesNGemsGame::DeserializeState(
...
+      rng_seed_, obs_show_ids, id_counter, cur_player);
+}
+
+std::string StonesNGemsGame::GetRNGState() const {
+  return std::to_string(rng_seed_);
+}
+
+void StonesNGemsGame::SetRNGState(const std::string &rng_state) const {
+  if (rng_state == "") return;
+  rng_seed_ = std::stoi(rng_state);
```
finbarrtimbers, line 1260: Can you use absl::SimpleAtoi instead?

open_spiel/spiel.cc
```diff
@@ -675,6 +694,15 @@ std::istream& operator>>(std::istream& stream, GameType::RewardModel& var) {
   return stream;
 }
 
+std::string Game::Serialize() const {
+  std::string str = ToString();
+  if (GetType().chance_mode == GameType::ChanceMode::kSampledStochastic) {
+    absl::StrAppend(&str, "\n", kSerializeGameRNGStateSectionHeader, "\n");
```
finbarrtimbers, line 700: Can you combine these?

open_spiel/python/tests/sampled_stochastic_games_test.py
```diff
@@ -0,0 +1,60 @@
...
+class SampledStochasticGamesTest(parameterized.TestCase):
+    @parameterized.parameters(*SPIEL_SAMPLED_STOCHASTIC_GAMES_LIST)
+    def test_stateful_game_serialization(self, game_info):
+        game = pyspiel.load_game(game_info.short_name,
+                                 {"rng_seed": pyspiel.GameParameter(0)})
+        # mutate game's internal RNG state
+        state = random_playout(game.new_initial_state())
+        deserialized_game = pickle.loads(pickle.dumps(game))
+
+        # make sure initial states are the same after game deserialization
```
finbarrtimbers, line 43: Nit: Can you make these proper sentences? (i.e. start with a capital letter and end with a period)

open_spiel/games/stones_and_gems.cc
```diff
@@ -1241,11 +1244,20 @@ std::unique_ptr<State> StonesNGemsGame::DeserializeState(
...
-      ++rng_seed_, obs_show_ids, id_counter, cur_player));
+      rng_seed_, obs_show_ids, id_counter, cur_player);
+}
+
+std::string StonesNGemsGame::GetRNGState() const {
+  return std::to_string(rng_seed_);
+}
+
+void StonesNGemsGame::SetRNGState(const std::string &rng_state) const {
+  if (rng_state == "") return;
```
finbarrtimbers, line 1259: empty?

open_spiel/games/negotiation.cc
inejc: I can't seem to find a solution without the stream. Any ideas?

finbarrtimbers: Don't worry about it, this is fine. I'll look into it if this turns out to be a bottleneck (which it almost certainly won't be).

finbarrtimbers:
Should be good- I just got distracted. I hope to merge this internally today, and it should appear in the tool the next time Marc does an export.

inejc:
Ok, thanks for the update.

PR #280, opened 2020-07-02
Adding alpha_beta search option to AlphaZero

BedirT:
Includes a simple alpha-beta search option in AlphaZero, instead of MCTS.


open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -46,7 +46,7 @@
 
 import numpy as np
 
-from open_spiel.python.algorithms import mcts
+from open_spiel.python.algorithms import mcts, minimax
```
tewalds, line 49: each import should be on its own line

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -59,8 +59,7 @@
 class TrajectoryState(object):
   """A particular point along a trajectory."""
 
-  def __init__(self, observation, current_player, legals_mask, action, policy,
-               value):
+  def __init__(self, observation, current_player, legals_mask, action, policy, value):
```
tewalds, line 62: We follow the google style guide: https://google.github.io/styleguide/pyguide.html , which has an 80 char line limit.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -194,6 +201,8 @@ def _init_bot(config, game, evaluator_, evaluation):
       dirichlet_noise=noise,
       child_selection_fn=mcts.SearchNode.puct_value,
       verbose=False)
+
```
tewalds, line 204: 2 lines between module level functions.

tewalds, line 204: probably worth adding an else: raise

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -184,8 +184,15 @@ def _watcher(*, config, num=None, **kwargs):
 
 def _init_bot(config, game, evaluator_, evaluation):
   """Initializes a bot."""
-  noise = None if evaluation else (config.policy_epsilon, config.policy_alpha)
-  return mcts.MCTSBot(
+  if config.search_type.lower() == "minimax":
```
tewalds, line 187: probably good to make this into an enum so the list of search types is known.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -204,20 +213,26 @@ def _play_game(logger, game_num, game, bots, temperature, temperature_drop):
...
     policy = np.zeros(game.num_distinct_actions())
-    for c in root.children:
-      policy[c.action] = c.explore_count
-    policy = policy ** (1 / temperature)
-    policy /= policy.sum()
-    if len(actions) >= temperature_drop:
-      action = root.best_child().action
-    else:
-      action = np.random.choice(len(policy), p=policy)
+    if type(cur_bot) == minimax.minimaxBot:
```
tewalds, line 218: prefer isinstance over type, though by the enum would be even better so it's all consistent.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -204,20 +213,26 @@ def _play_game(logger, game_num, game, bots, temperature, temperature_drop):
...
+      value, action = cur_bot.alpha_beta_search(state=state, maximum_depth=30)
```
tewalds, line 219: maximum_depth should definitely be configurable. Depth of 30 will never finish in any non-trivial game.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
```diff
@@ -284,16 +299,24 @@ def evaluator(*, game, config, logger, queue):
     az_player = game_num % 2
     difficulty = (game_num // 2) % config.eval_levels
     max_simulations = int(config.max_simulations * (10 ** (difficulty / 2)))
-    bots = [
-        _init_bot(config, game, az_evaluator, True),
-        mcts.MCTSBot(
+    bots = [_init_bot(config, game, az_evaluator, True)]
+    if config.search_type.lower() == "minimax":
```
tewalds, line 303: Do you want to compare against the minimax bot? or keep mcts as the evaluation opponent. I've considered adding others, eg external gtp bots. Keeping mcts as the evaluation opponent would make the results more comparable to usual AZ.

open_spiel/python/algorithms/minimax.py
```diff
@@ -208,3 +208,22 @@ def expectiminimax(state, depth, value_function, maximizing_player_id):
         value = child_value
         best_action = action
     return value, best_action
+
+
+class minimaxBot(pyspiel.Bot):
```
tewalds, line 213: Classes should be in CapWords format.

open_spiel/python/algorithms/minimax.py
```diff
@@ -208,3 +208,22 @@ def expectiminimax(state, depth, value_function, maximizing_player_id):
...
+
+  def __init__(self,
+               game,
+               max_simulations,
```
tewalds, line 217: what would this be used for?

open_spiel/python/algorithms/minimax.py
```diff
@@ -208,3 +208,22 @@ def expectiminimax(state, depth, value_function, maximizing_player_id):
...
+               evaluator):
+   
+    pyspiel.Bot.__init__(self)
```
tewalds, line 220: prefer super()

open_spiel/python/algorithms/minimax.py
```diff
@@ -208,3 +208,22 @@ def expectiminimax(state, depth, value_function, maximizing_player_id):
...
+               max_simulations,
+               evaluator):
+   
+    pyspiel.Bot.__init__(self) 
+    self._game = game
+    # self.max_simulations = max_simulations
+    # self.evaluator = evaluator
+
+  def alpha_beta_search(self, state=None,
+                      value_function=None,
```
tewalds, line 226: Hmm, can it be said to be AZ-like if there is no value function? Does it only work by solving the entire game on every search?

open_spiel/python/examples/alpha_zero.py
```diff
@@ -53,6 +53,7 @@
     ("Play evaluation games vs MCTS+Solver, with max_simulations*10^(n/2)"
      " simulations for n in range(eval_levels). Default of 7 means "
      "running mcts with up to 1000 times more simulations."))
+flags.DEFINE_string("search_type", 'mcts', 'Search type that will be used, either mcts or minimax')
```
tewalds, line 56: use an enum.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
BedirT: Yeah, that was kind of the aim, but yeah I guess evaluating vs MCTS is better. I will fix that.

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
BedirT: Do I need a raise, since I am checking the flag on the example file?

open_spiel/python/algorithms/alpha_zero/alpha_zero.py
BedirT: Can you clarify what you mean by having them as enum ?

open_spiel/python/algorithms/minimax.py
BedirT: Yes, that's the idea. This is the initial version, so I didn't add a base value function, maybe we could add softmax?

BedirT:
@tewalds I am having an issue adding the second parameter to config. Max_depth is giving me TypeError... I am certain that everything is on place, I don't know what am I doing wrong here is the error:

```
<class 'open_spiel.python.algorithms.alpha_zero.alpha_zero.Config'>
  File "alpha_zero.py", line 104, in <module>
    app.run(main)
  File "/Users/mustafatapkan/anaconda3/lib/python3.7/site-packages/absl/app.py", line 299, in run
    _run_main(main, args)
  File "/Users/mustafatapkan/anaconda3/lib/python3.7/site-packages/absl/app.py", line 250, in _run_main
    sys.exit(main(argv))
  File "alpha_zero.py", line 97, in main
    quiet=FLAGS.quiet,
TypeError: __new__() got an unexpected keyword argument 'max_depth'
```

For some reason I guess it doesn't see **"max_depth"** in config tuple. Any suggestions what might cause this ?

open_spiel/python/algorithms/minimax.py
lanctot: AZ does learn a value function. It is what separates it from vanilla MCTS. Minimax could use thst learned value function in the same way as it uses the heuristic evaluation function when supplied with one.

lanctot:
@BedirT @tewalds can you give give an update on where we are with this PR?

PR #288, opened 2020-07-04
Support chance nodes in LegalActionsMask

### LINKED ISSUE #283: Inconsistent length of `LegalActions` and `LegalActionsMask` for chance nodes
For example, in the game `kuhn_poker`, the `LegalActions` of the first chance node is `[0,1,2]`, while the `LegalActionsMask` fallbacks to the default implementation and returns `[1,1]`. Is this expected? (If so, does it mean that I should always rely on `LegalActions` instead of `LegalActiosnMask` for chance nodes?)

findmyway:
- Add an extra field of `max_chance_outcomes_` in `State`
- Keep the length of `LegalActionsMask` and `LegalActions` equal for chance nodes. (fix #283 )

In the meanwhile, I find that the return of `LegalActions` of the game `pig` with chance node is between `[1, MaxChanceOutcomes]` instead of `[0, MaxChanceOutcomes-1]`, which breaks the test.

https://github.com/deepmind/open_spiel/blob/8c8e0863380acaf93fe0b00780196d9117eaac66/open_spiel/games/pig.cc#L240

I'm not sure it is intentional or not. So not fixed yet. @lanctot  Any idea?

### DIFF
--- a/open_spiel/games/first_sealed_auction.h
+++ b/open_spiel/games/first_sealed_auction.h
@@ -83,7 +83,7 @@ class FPSBAGame : public Game {
     return std::unique_ptr<State>(new FPSBAState(shared_from_this()));
   }
   int MaxChanceOutcomes() const override {
-    return std::max(max_value_, num_players_);
+    return std::max(max_value_, num_players_) + 1;
   }
   int NumPlayers() const override { return num_players_; }
   double MinUtility() const override { return 0; }
--- a/open_spiel/games/first_sealed_auction.h
+++ b/open_spiel/games/first_sealed_auction.h
@@ -83,7 +83,7 @@ class FPSBAGame : public Game {
     return std::unique_ptr<State>(new FPSBAState(shared_from_this()));
   }
   int MaxChanceOutcomes() const override {
-    return std::max(max_value_, num_players_) + 1;
+    return std::max(max_value_+1, num_players_);
   }
   int NumPlayers() const override { return num_players_; }
   double MinUtility() const override { return 0; }
--- a/open_spiel/games/pig.cc
+++ b/open_spiel/games/pig.cc
@@ -201,7 +201,7 @@ void PigState::DoApplyAction(Action move) {
   } else if (IsChanceNode()) {
     // Resolve chance node outcome. If 1, reset turn total and change players;
     // else, add to total and keep going.
-    if (move == 1) {
+    if (move == 0) {
       // Reset turn total and loses turn!
       turn_total_ = 0;
       turn_player_ = NextPlayerRoundRobin(turn_player_, num_players_);
@@ -237,7 +237,7 @@ std::vector<std::pair<Action, double>> PigState::ChanceOutcomes() const {
   // All the chance outcomes come after roll and stop.
   outcomes.reserve(dice_outcomes_);
   for (int i = 0; i < dice_outcomes_; i++) {
-    outcomes.push_back(std::make_pair(i + 1, 1.0 / dice_outcomes_));
+    outcomes.push_back(std::make_pair(i, 1.0 / dice_outcomes_));
   }
 
   return outcomes;
--- a/open_spiel/games/trade_comm.h
+++ b/open_spiel/games/trade_comm.h
@@ -96,7 +96,7 @@ class TradeCommGame : public Game {
     return std::unique_ptr<State>(
         new TradeCommState(shared_from_this(), num_items_));
   }
-  int MaxChanceOutcomes() const override { return num_items_; }
+  int MaxChanceOutcomes() const override { return num_items_ * num_items_; }
 
   int MaxGameLength() const override { return 4; }
 
--- a/open_spiel/games/liars_dice.cc
+++ b/open_spiel/games/liars_dice.cc
@@ -410,7 +410,7 @@ std::unique_ptr<State> LiarsDiceGame::NewInitialState() const {
   return state;
 }
 
-int LiarsDiceGame::MaxChanceOutcomes() const { return kDiceSides + 1; }
+int LiarsDiceGame::MaxChanceOutcomes() const { return kDiceSides; }
 
 int LiarsDiceGame::MaxGameLength() const {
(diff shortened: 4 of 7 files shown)


lanctot:
Yeah not intentional, another bug. Leftover from a refactor from the old API. You can fix it in this PR too thanks. But you need to make sure the logic of the ApplyAction makes sense for the new range too.

findmyway:
- [ ] first_sealed_auction
- [ ] liars_dice
- [ ] oshi_zumo
- [ ] trade_comm

open_spiel/games/first_sealed_auction.h
```diff
@@ -83,7 +83,7 @@ class FPSBAGame : public Game {
     return std::unique_ptr<State>(new FPSBAState(shared_from_this()));
   }
   int MaxChanceOutcomes() const override {
-    return std::max(max_value_, num_players_);
+    return std::max(max_value_+1, num_players_);
```
findmyway, line 86: Based on the implementation in `ChanceOutcomes`:

https://github.com/deepmind/open_spiel/blob/8c8e0863380acaf93fe0b00780196d9117eaac66/open_spiel/games/first_sealed_auction.cc#L195

and 

https://github.com/deepmind/open_spiel/blob/8c8e0863380acaf93fe0b00780196d9117eaac66/open_spiel/games/first_sealed_auction.cc#L189

the range should be `[0, max(max_value_+1, num_players)]`.

lanctot:
Oops I mistakenly resolved the conversation now cannot get it back. Can you add your comment in first_sealed_auction again? I will point the author to this thread.

Out of curiousity, why are you using LegalActionsMask at chance nodes?

open_spiel/games/pig.cc
```diff
@@ -201,7 +201,7 @@ void PigState::DoApplyAction(Action move) {
   } else if (IsChanceNode()) {
     // Resolve chance node outcome. If 1, reset turn total and change players;
     // else, add to total and keep going.
-    if (move == 1) {
+    if (move == 0) {
```
findmyway, line 204: It seems that this is the only needed change. The following line already increments the `move`.

https://github.com/deepmind/open_spiel/blob/8c8e0863380acaf93fe0b00780196d9117eaac66/open_spiel/games/pig.cc#L211

open_spiel/games/trade_comm.h
```diff
@@ -96,7 +96,7 @@ class TradeCommGame : public Game {
     return std::unique_ptr<State>(
         new TradeCommState(shared_from_this(), num_items_));
   }
-  int MaxChanceOutcomes() const override { return num_items_; }
+  int MaxChanceOutcomes() const override { return num_items_ * num_items_; }
```
findmyway, line 99: This change is based on:

https://github.com/deepmind/open_spiel/blob/master/open_spiel/games/trade_comm.cc#L234

findmyway:
> Oops I mistakenly resolved the conversation now cannot get it back. Can you add your comment in first_sealed_auction again? I will point the author to this thread.

Done

> Out of curiosity, why are you using LegalActionsMask at chance nodes?

I'd like to treat chance player just like other general players. And initial the policy with:

```julia
probs[legal_actions_mask] .= [v for (k,v) in chance_outcomes(state)]
```

In OpenSpiel, the policy is always `(action, prob)` pairs. While in my implementation, for policies working with discrete actions, they only know the index.

open_spiel/games/trade_comm.h
lanctot: Yes this is right, nice catch!

open_spiel/games/pig.cc
lanctot: Yep, this looks good.

open_spiel/games/first_sealed_auction.h
lanctot: Will ask the author to take a look, thanks.

lanctot: @jkramar wdyt?

jkramar: As defined on line 189, the valuations are in the range `[1, max_value_]`, so I think `MaxChanceOutcomes()` as written is correct. How do you see `max_value_ + 1` arising?

findmyway: `max_value_ + 1` will not arise. But since `max_value_ ∈ [1, max_value_]` and `player ∈ [0, num_players_-1]`, so the possible chance outcome might be in range `[0, max(max_value_, num_players-1]`. Then I think `MaxChanceOutcomes()` is `max(max_value_+1, num_players_)`

jkramar: Icwym. But I think that's answering a different question: `MaxChanceOutcomes()`'s doc comment is that it's `// Maximum number of chance outcomes for each chance node`, whereas it seems you're using it analogously to `NumDistinctActions()`, which includes actions combined from different nodes/states (as you're doing). If this is how `MaxChanceOutcomes()` is used then I think the documentation should be updated to clarify that. Alternately you might use a separate function like `NumDistinctChanceOutcomes()`. @lanctot, wdyt ?

lanctot: We should be treating `MaxChanceOutcomes` as the chance node equivalent of `NumDistinctActions`, that was always my intention... apologies @jkramar if the documentation implies otherwise. @findmyway could you also fix the doc of the function to clarify this as well?

findmyway: Updated in [the latest commit](https://github.com/deepmind/open_spiel/pull/288/files/94d62eac2c8ba9223bb02060c7ee516e3e025f04..6794f2be80a5969643c4d1abcad6810ff170a4d7#diff-bdb8eda5e3f025cd1d06a389c4c99ee5L769).

lanctot:
Hi @findmyway , just checking; is this one good to start review now?

findmyway:
Yes 😃

elkhrt:
Thanks for this contribution - it's smoked out a couple of bugs, which I'm fixing as I integrate the PR.

PR #304, opened 2020-07-15
Repeated games and roshambo bots

jhtschultz:
The International Roshambo Programming Competition was held in 1999 and 2000 by Darse Billings at the University of Alberta (https://webdocs.cs.ualberta.ca/~darse/rsbpc.html). The player pool was seeded with dummy bots, requiring competitive entrants to exploit sub-optimal play while avoiding getting exploited by other sophisticated bots. All of the code used in the first competition, including the entire field of bots, is open source (https://webdocs.cs.ualberta.ca/~darse/rsb-ts1.c). This submission provides an OpenSpiel interface to these bots.

No additional code from the second competition was officially released, but Andrzej Nagorko, author of the winning entrant Greenberg, independently released the bot’s source code (http://www.mathpuzzle.com/older.htm, http://www.mathpuzzle.com/greenberg.c). Greenberg is also included as an OpenSpiel bot.

Each tournament match consisted of 1000 rounds of rock-paper-scissors. To play a match, this pull request includes a game transform that creates a repeated game from a normal-form stage game (https://en.wikipedia.org/wiki/Repeated_game).

In providing a link to a historically interesting experiment in games research, this addition to OpenSpiel has the potential to serve as a useful benchmark for opponent modeling, safe exploitation, and population based RL.


jhtschultz:
Implementation Notes:

There was no obvious place to house the roshambo bots, so I created a bot directory. This seemed sensible since there is an analogous bot folder within the Python directory. 

My guiding philosophy in designing the bot interface was to modify the original C code as little as possible. Some slight modifications (like commenting out `main()`) were unavoidable. That prevents downloading the code directly from the source, so I’m storing a modified version in my GitHub, which is downloaded if the optional dependency is enabled.

Feedback is welcome and I’m happy to implement suggested improvements. Thanks!

elkhrt:
Very cool contribution - thank you!
Looks like the actual bot code is missing though?

jhtschultz:
Thanks a lot for reviewing the submission Ed! 

My understanding from discussing this with @lanctot is that the actual bot code should be hosted externally. Running install.sh with the global variable BUILD_WITH_ROSHAMBO=ON will download that code. 

So I didn't think I should include the external bot code in the pull request, but if including it makes reviewing everything easier on your end then I can update the PR.

elkhrt:
Have you committed the latest version of the Greenberg code?
I get an error on this line: short int my_history_hash[4][trials];
 (trial not defined)

jhtschultz:
Yes I can confirm the latest version of Greenberg is present in the `rsb-ts1-modified.c` file that gets downloaded by `install.sh`. I'm guessing that error came from `roshambo_bot_test`? Just now I ran the test on a fresh OpenSpiel install on a linux machine and the test ran fine. Not sure why it's throwing an error. 

Do you still get an error if you switch Greenberg for another bot?

elkhrt:
Oh, I don't see rsb-ts1-modified.c in
https://github.com/jhtschultz/roshambo/tree/master/IntlRoshamboProgrammingCompetition
Should I be looking somewhere else?

jhtschultz:
Ah that explains it! `rsb-ts1-modified.c` is located on a separate branch, `open_spiel`. The `install.sh` script clones from the correct branch: https://github.com/jhtschultz/open_spiel/blob/roshambo/open_spiel/scripts/install.sh#L130

There are other roshambo bots I've been looking at adding to my repo, so the idea is to have a separate, stable branch dedicated to hosting roshambo files imported by OpenSpiel. I noticed that a few of the imports in `install.sh` specified a particular branch, so it seemed like a reasonable thing to do. Sorry for any ambiguity and hope that solves the error.

elkhrt:
Thanks - I've now found the code.
Could you make the external repository a bit more C++ friendly? In
particular, the number of global symbols is rather scary. Prefixing them
all with ROSHAMBO_BOT_ would be one option.
It would also be preferable to have a function in the external repo that
returned a list of bots so that we don't have to make changes in two places
when adding a new one.

Thanks,

Ed

jhtschultz:
Sure thing. I'm out most of the day today but I'll start on that tonight and let you know when it's been updated.

jhtschultz:
Hi Ed,

Taking a closer look at this now and perhaps you could shed some light on how exactly the global symbols in the C code are dangerous to the C++ code? I ran some tests and the global variables seem only to be accessible in C++ if they are explicitly declared with extern. The #define macros don’t appear to have any effect either when I tried reusing the identifier in C++. 

Happy to throw ROSHAMBO_BOT_ in front of anything that could be an issue, just trying to better understand how C and C++ interact in this context. Thanks :)

elkhrt:
Yes, the issue with the symbols is that they are global, so would cause a
conflict if the name was reused elsewhere. That's not the case now for any
uses of OoenSpiel that I know of, but it could be in future. Having a
#define prefix on them (a) makes conflicts less likely and (b) makes it
easy to change if a conflict does arise. Have a look at what I did to
integrate the double dummy code for an example.

jhtschultz:
Ah that's a useful reference, thanks. I will let you know when that and the bot list function are completed.

jhtschultz:
Hi Ed,

Completed the changes. Moved the `bot_map` from `roshambo_bot.cc` into the external repo (which has also been updated). Adding a new bot from the competition can now be handled entirely externally. As for the global symbols, I followed your suggestion and prefixed everything in sight with ROSHAMBO_BOT_. If there’s a more elegant way of doing it I’d love to know!

`rsb-ts1-modified.c` compiles and runs fine if `main()` is uncommented. And the outcome of the tournament remains the same, so I’m fairly confident the C code is working correctly after the modifications.

Thanks again and if there’s anything else I can do let me know. Cheers

PR #306, opened 2020-07-20
InformationStateTensor() fixes in catch and deep_sea

jhtschultz:
`InformationStateTensor()` in `catch.cc` contained a bug. The offset was off by one, so in place of the player's first action, it recorded the chance action that determined the column of the ball. As per @lanctot's suggestion, this submission removes the function entirely.

The `GameType` in `deep_sea.cc` read `/*provides_information_state_as_normalized_vector=*/false,` when in fact it did provide the function `InformationStateTensor()`. This submission fixes the outdated `as_normalized_vector` description, and removes information states from deep_sea.

In addition, the observation tensor in catch.py has been slightly modified to agree with the observation tensor in catch.cc.


finbarrtimbers:
Lgtm- merging! Thanks for the contribution!

PR #314, opened 2020-07-29
Add trick-taking card game Oh Hell

solinas:
This change includes an implementation of the game [Oh Hell](https://en.wikipedia.org/wiki/Oh_Hell) with InformationStateTensor/String and ResampleFromInfoState implemented. 

Oh Hell is good for research in trick-taking card games because changing the number of cards in the deck is a straightforward way to scale the size of the game up or down without seriously changing how it works. The default variant uses all 52 cards in a standard deck, but reducing the number of cards to around 10 or so makes the game small enough to apply tabular algorithms.
The implementation accepts GameParameters to control the number of cards for this reason.

As commented above the function, the implementation I provided for ResampleFromInfoState can produce inconsistent samples because it ignores void suits. I plan to fix this in a later PR.


lanctot:
Thanks @solinas, had not heard of the game, looks cool!

lanctot:
Hi @solinas, just a heads-up that we still plan to import this.. just got busy with vacation and an influx of small PRs. Planning to get to this in the coming weeks.

open_spiel/games/oh_hell.cc
```diff
@@ -0,0 +1,589 @@
...
+  bids_.resize(num_players_);
+  // need to differentiate between no bid and a bid of 0
+  std::fill(bids_.begin(), bids_.end(), kInvalidBid);
+  num_tricks_won_.resize(num_players_);
+  returns_.resize(num_players_);
+  holder_.resize(deck_props_.NumCards());
+  initial_deal_.resize(deck_props_.NumCards());
+}
+
+std::string OhHellState::ToString() const {
```
bartdevylder, line 112: I think it would be helpful to add (a string representation) of the the current game phase ("ChooseNumTricks", "Dealer", ...) at the beginning of the state string

open_spiel/games/oh_hell_test.cc
```diff
@@ -0,0 +1,241 @@
...
+
+#include "open_spiel/games/oh_hell.h"
+#include "open_spiel/spiel.h"
+#include "open_spiel/tests/basic_tests.h"
+
+namespace open_spiel {
+namespace oh_hell {
+namespace {
+
+void GameConfigSimTest() {
```
bartdevylder, line 25: Very nice tests! 
It could be I missed it but in case there is no test yet for the case num_tricks_fixed != kRandomNumTricks, could be nice to also test that case.

open_spiel/games/oh_hell.h
```diff
@@ -0,0 +1,248 @@
...
+  double MaxUtility() const override {
+    if (num_tricks_fixed_ > 0) return num_tricks_fixed_ + kMadeBidBonus;
+    return MaxNumTricks() + kMadeBidBonus;
+  }
+  std::shared_ptr<const Game> Clone() const override {
+    return std::shared_ptr<const Game>(new OhHellGame(*this));
+  }
+  int MaxGameLength() const override {
+    return 2 * MaxNumTricks() * num_players_ + num_players_
+             + kNumPreDealChanceActions + kNumTrumpDeal;
```
bartdevylder, line 229: Maybe add comment to explain the terms that make up the MaxGameLength

bartdevylder (commented):
Hi @solinas, thanks for adding this game to openspiel (I like the game a lot ;)! It's a very nice implementation, only few small remarks

solinas:
Thanks for the review @bartdevylder. Your comments should hopefully be addressed now.

I had to make a few other changes to be consistent with the changes upstream:

- Removing OhHellGame::Clone() (ea4da902f21443996e1f7254813198c499eb1dc2)
- Adding an implementation for OhHellGame::MaxChanceNodesInHistory() (b222c4b296332c04959bbafd6a204266eb667a6e)

My last commit (4ed7a8f) makes this implementation of Oh Hell pass the `observation_consistency` test in the `python_api_test`. If I understand correctly, the consistency check involves ensuring that action histories with different lengths are not permitted to have the same InformationSetString. This seems strict for games with chance nodes because it means that a player's information set string must change after an unobserved chance action. I suspect this is why many card games are currently failing (and are excluded from) this test. satisfy this constraint.

PR #324, opened 2020-08-31
Better game serialization during CFR serialization

inejc:
Use game serialization that also covers sampled stochastic game implementations when de/serializing CFR solvers.


lanctot:
Hi @inejc thanks for the PRs. I am off until the 10th so will take a look when i am back.

PR #328, opened 2020-09-04
Add registration of SpielException so it can be caught in Python with…

michalsustr:
… more specificity than just some RuntimeError.

### DIFF
--- a/open_spiel/integration_tests/api_test.py
+++ b/open_spiel/integration_tests/api_test.py
@@ -173,7 +173,7 @@ class EnforceAPIOnFullTreeBase(parameterized.TestCase):
   def test_information_state_no_argument_raises_on_terminal_nodes(self):
     for state in self.all_states:
       if state.is_terminal():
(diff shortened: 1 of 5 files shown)


elkhrt:
Thanks. I'm merging this and fixing the various tests you missed in the process (so there's no need for you to fix this PR).

michalsustr:
Oh, I didn't notice the comment! Ha, well, I just pushed the changes as well. I tried to make SpielError to inherit from RuntimeError, but I didn't find an easy way to do so with pybind11 -- maybe the first thing I couldn't find in otherwise this awesome library :)

michalsustr:
Ok, turns out the feature was not documented, but it is easy to do so!

lanctot:
I think this was merged but did not get auto-closed because there were extra commits. @michalsustr can you check to see if everything you want is now in master?

michalsustr:
No commits were merged. Can you import it again with the new changes? Thank you!

PR #331, opened 2020-09-07
Fix the sampled stochastic games test

inejc:
I believe the current test was inadequate as the RNG state of sampled stochastic games was not actually advanced before the de/ser cycle (line 47). In order to advance the state, a full playout of a "single game" should be performed. The RNG state needs to be advanced to avoid serializing the initial RNG state value which would be "deserialized" correctly even in cases where deserialization is simply creating an instance with default initial values. cc @finbarrtimbers


finbarrtimbers:
Nice. I didn't actually understand how the state was being advanced, so this makes sense to me.

finbarrtimbers:
This will be automatically merged next week!

inejc:
Great, thanks for letting me know!

PR #343, opened 2020-09-08
Add shorthand methods to ask for private info type.

michalsustr:


elkhrt:
I don't think this is an improvement. It makes the code less transparent at the call site.

michalsustr:
Yeah, I think you're right. We don't need to do things this way.

PR #344, opened 2020-09-08
Add GoofspielObserver.

michalsustr:


elkhrt:
This appears not to be backwards-compatible - the observation tensors have changed.
Also, please remove the changes to Kuhn poker and to common code.

michalsustr:
Ok. I created a separate PR with requested changes #364  as it was easier to do.

PR #346, opened 2020-09-08
Infer tensor shapes automatically from game's Observer.

michalsustr:


elkhrt:
I don't think this is an improvement - it makes KuhnPoker less comprehensible; this is an important entry point for new users and it should be as easy as possible to understand.

Another (fixable) issue is that it moves a class into the external API without adequate documentation.

michalsustr:
Will it be OK if I revert Kuhn changes and add documentation?

elkhrt:
Actually, I don't see why TrackingVectorAllocator has to move into the
public API at all.

michalsustr:
Not in here, true.

michalsustr:
I will do it in a different PR: #358

PR #349, opened 2020-09-08
Add a specific bounds check.

michalsustr:
I spent bunch of time on this one -- subspan throws an exception which is translated into IndexError in Python without a stacktrace, making it difficult to locate the source of problem. By duplicating this boundary check it is easier to track down future bugs.


michalsustr:
One travis test did not pass through because of network timeout on the server, not sure if I can restart the check (without a new commit).

PR #354, opened 2020-09-15
Fix developer guide docs

inejc:
I had to run the `generate_new_playthrough.sh` script from the root dir or the command failed with `python: can't open file 'open_spiel/python/examples/playthrough.py': [Errno 2] No such file or directory`.


lanctot:
Right, nice catch. Thanks!

PR #355, opened 2020-09-16
Make sure names of tensors are unique.

michalsustr:


michalsustr:
Ok, this one is already in, I have too many branches on my local copy of open spiel :) What I am confused about though is that github claims there is a diff between master and the PR, while this is already in master! Weird.

PR #357, opened 2020-09-16
Add distinction between (im)perfect recall in perfect information obs…

michalsustr:
…ervers.


open_spiel/observer.cc
```diff
@@ -118,15 +118,26 @@ std::shared_ptr<Observer> Game::MakeObserver(
...
+  if (GetType().information == GameType::Information::kPerfectInformation) {
+    // Handle the dummy case, where we do not use any public information.
+    // The game will just have empty private observations.
+    if (!iig_obs_type->public_info) {
       return absl::make_unique<NoPrivateObserver>(*this);
     }
+
+    // Distinguish two variants:
+    // Provide observer with or without perfect recall.
+    SPIEL_CHECK_TRUE(iig_obs_type->public_info);
```
elkhrt, line 130: Remove this. Perhaps use an else clause.

open_spiel/observer.cc
```diff
@@ -118,15 +118,26 @@ std::shared_ptr<Observer> Game::MakeObserver(
 
   // Perfect information games provide public information regardless
```
elkhrt, line 119: This comment is no longer accurate.

open_spiel/observer.cc
elkhrt: Actually maybe this logic can be further simplified.

const bool perfect_info_game = GetType().information == GameType::Information::kPerfectInformation;
  if (perfect_info_game && !iig_obs_type->public_info) {
      return absl::make_unique<NoPrivateObserver>(*this);
 }

if (iig_obs_type->perfect_recall && (game_type_.provides_information_state_tensor || game_type_.provides_information_state_string)) {
  SPIEL_CHECK(perfect_info_game || iig_obs_type->private_info == PrivateInfoType::kSinglePlayer);
  return absl::make_unique<InformationStateObserver>(*this);
}

/ same for default observer

michalsustr: I simplified the logic.

open_spiel/observer.cc
michalsustr: Why?

elkhrt:
Looks like this affects some playthroughs. Is that deliberate?

michalsustr:
I added the forgotten playthroughs. Now the perfect information games also log the public observations in playthroughs. It is redundant information, but I think it's ok for the tests.

open_spiel/observer.cc
elkhrt: Because NoPrivateObserver does not provide public information.

PR #358, opened 2020-09-16
Collect tensor information for an observer.

michalsustr:


elkhrt:
I don't see the advantage over just getting an Observation for the initial state.

michalsustr:
Yes, it is very similar. It does not hold the observation data and it only needs a const & to an observer, so you do not need to pass an owning shared_ptr that is actually used just once.

elkhrt:
But you will always have a shared_ptr in practice. I don't think is worth the additional complexity.

michalsustr:
> But you will always have a shared_ptr in practice. I don't think is worth the additional complexity.

Sure. I just have some other code (not published yet) that needs this information, and then shared_ptr is passed around without it needing to be so with this change. But I guess it's not too important, so I will close the PR without merging.

PR #360, opened 2020-09-17
Fix incorrect Kuhn Poker max game length:

michalsustr:
The original computation probably did not include the chance nodes.

### DIFF
--- a/open_spiel/games/kuhn_poker.h
+++ b/open_spiel/games/kuhn_poker.h
@@ -108,7 +108,7 @@ class KuhnGame : public Game {
   double UtilitySum() const override { return 0; }
   std::vector<int> InformationStateTensorShape() const override;
   std::vector<int> ObservationTensorShape() const override;
-  int MaxGameLength() const override { return num_players_ * 2 - 1; }
+  int MaxGameLength() const override { return num_players_ * 3 - 1; }
   std::shared_ptr<Observer> MakeObserver(
       absl::optional<IIGObservationType> iig_obs_type,
       const GameParameters& params) const override;


lanctot:
Max game length is not supposed to include chance nodes: https://github.com/deepmind/open_spiel/blob/5354afc54d8cb1f96ebc320af16243ac6bdcb0cb/open_spiel/spiel.h#L756

michalsustr:
Oh sorry, I missed that. However some code needs to get a number from the game, which upper bounds (and ideally equals) the maximum MoveNumber(). Should I add a method Game::MaxGameMoveNumber() which will include chance as well?

lanctot:
Sorry, can you give more context? What do you need it for? Are these move numbers only used in the games under public states API or in fogs? If so, can we make them part of that (sub-)API only?

michalsustr:
Sure. I am generating (action) histories based on sequence of public observations in the game. I need to pre-allocate a buffer to store the observations. Since chance nodes contribute to public observations as well, I can't currently compute the size of the buffer just based on MaxGameLength.

lanctot:
Ok I guess we can add a virtual MaxLengthWithChance or, maybe an even better name: MaxHistoryLength, with defaults to MaxGameLength() but has to be overridden in games with chance. But if we do that, it would be good to implement it for all our games that have chance.

CAn you think of way we could put this function in the public states API or fog-specific code not in the core API?

michalsustr:
Hi, I can't think of a clean way to do so without putting it in Game. I like `MaxHistoryLength()`. We can say it has to return upper bound on the value, and authors should try to make the bound as tight as possible. I will make it tight for games where I know what the exact value is. I can also add it to remaining games, and return `MaxGameLength() * 2` -- that should cover all the remaining ones, as AFAIK in OpenSpiel there aren't games which have more chance nodes than player nodes. I will also implement a test for the sparse sampling to check the samples aren't longer than the bound. What do you think @lanctot ?

lanctot:
Sounds perfect to me. @elkhrt you ok with this?

I can also tighten the ones I know more about upon import.

I think it's ok to assume should be fine. So we will close this PR and you will open a new one?

michalsustr:
Yes, sounds good.

elkhrt:
Sounds good to me.

michalsustr:
Actually, there is a small issue with sim move games, because the actions are flattened for history. `MaxHistoryLength()` would be a number unnecessarily larger than `MoveNumber()`, up to multiplicative constant of `NumPlayers()`. So I thought to do it differently, and factorize the computation. As someone might also want to recover the maximum number of chance events (like number of times we roll dice), I thought I can add a 

1. `MaxChanceNodesOnTrajectory()` 
2. `MaxMoveNumber()` which will return the sum of `MaxGameLength() + MaxChanceNodesOnTrajectory()`
3. `MaxHistoryLength()` which will return `MaxGameLength() * NumPlayers() + MaxChanceNodesOnTrajectory()` for sim games, and without player multiplication for sequential games.

If you think the change is too much, I'm happy with just adding the first method, and compute the remainder in algorithm code.

lanctot:
Fine by me!

elkhrt:
Sure, SGTM.

PR #361, opened 2020-09-17
Fix includes in integration test example.

michalsustr:


lanctot:
This one requires a rewrite rule on our side, and the copybara import sometimes interferes with that, so I'll just do it from our side. Nice catch (also noticed it it in the AlphaZero C++ LibTorch PR.

lanctot:
Nevermind, it doesn't actually. Importing now.

PR #362, opened 2020-09-17
Add link to libtorch c++ examples.

michalsustr:


lanctot (changes requested):
Can you put this comment in our C++ example itself, and then instead point the user to our libtorch example?

michalsustr:
Sure!

lanctot (approved):

PR #364, opened 2020-09-18
Add GoofspielObserver.

michalsustr:


elkhrt:
Is there a good reason for the change to ObservationString? Can you make it backwards-compatible?

michalsustr:
Now it's almost completely compatible, except for `Win sequence` / `Win Sequence`

PR #365, opened 2020-09-18
Observing player is not a public information.

michalsustr:


elkhrt (commented):
iiuc this reverses the order of private_card and player. Is that deliberate?
Also, why is no test affected by this change?

michalsustr:
> iiuc this reverses the order of private_card and player. Is that deliberate?
No it doesn't, player is still before private_card.

> Also, why is no test affected by this change?
I think at the time of making the PR the public observations were not in the playthroughs yet.

PR #368, opened 2020-09-21
Add MaxChanceNodesInHistory() to Game API.

michalsustr:
This PR properly addresses issue raised in #360 and adds following methods to Game:
`MaxChanceNodesInHistory`, `MaxMoveNumber` and `MaxHistoryLength` both in C++ and Python. For some games I didn't know how to make the bound tight fast, so I left a TODO and just used the `MaxGameLength()` call-through value. Some games had a wrong declaration of chance mode in their game type. The `first_sealed_auction` was kind of weird, I left a comment.

### DIFF
--- a/open_spiel/game_transforms/coop_to_1p.h
+++ b/open_spiel/game_transforms/coop_to_1p.h
@@ -163,7 +163,7 @@ class CoopTo1pGame : public Game {
   int NumPlayers() const override { return 1; }
   std::vector<int> ObservationTensorShape() const override;
   int MaxGameLength() const override;
-  // TODO() make a tighter bound.
+  // TODO: verify whether this bound is tight and/or tighten it.
   int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
 
   int NumDistinctActions() const override {
--- a/open_spiel/game_transforms/coop_to_1p.h
+++ b/open_spiel/game_transforms/coop_to_1p.h
@@ -163,6 +163,8 @@ class CoopTo1pGame : public Game {
   int NumPlayers() const override { return 1; }
   std::vector<int> ObservationTensorShape() const override;
   int MaxGameLength() const override;
+  // TODO() make a tighter bound.
+  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
 
   int NumDistinctActions() const override {
     return game_->NumDistinctActions();
--- a/open_spiel/game_transforms/coop_to_1p.h
+++ b/open_spiel/game_transforms/coop_to_1p.h
@@ -163,8 +163,9 @@ class CoopTo1pGame : public Game {
   int NumPlayers() const override { return 1; }
   std::vector<int> ObservationTensorShape() const override;
   int MaxGameLength() const override;
-  // TODO: verify whether this bound is tight and/or tighten it.
-  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
+  int MaxChanceNodesInHistory() const override {
+    return game_->MaxGameLength();
+  }
 
   int NumDistinctActions() const override {
     return game_->NumDistinctActions();
--- a/open_spiel/game_transforms/repeated_game.h
+++ b/open_spiel/game_transforms/repeated_game.h
@@ -65,6 +65,8 @@ class RepeatedGame : public SimMoveGame {
                const GameParameters& params);
   std::unique_ptr<State> NewInitialState() const override;
   int MaxGameLength() const override { return num_repetitions_; }
+  // There aren't chance nodes in this game.
+  int MaxChanceNodesInHistory() const override { return 0; }
   int NumPlayers() const override { return stage_game_->NumPlayers(); }
   int NumDistinctActions() const override {
     return stage_game_->NumDistinctActions();
--- a/open_spiel/game_transforms/repeated_game.h
+++ b/open_spiel/game_transforms/repeated_game.h
@@ -65,8 +65,6 @@ class RepeatedGame : public SimMoveGame {
                const GameParameters& params);
   std::unique_ptr<State> NewInitialState() const override;
   int MaxGameLength() const override { return num_repetitions_; }
-  // There aren't chance nodes in this game.
-  int MaxChanceNodesInHistory() const override { return 0; }
   int NumPlayers() const override { return stage_game_->NumPlayers(); }
   int NumDistinctActions() const override {
     return stage_game_->NumDistinctActions();
--- a/open_spiel/games/backgammon.h
+++ b/open_spiel/games/backgammon.h
@@ -264,6 +264,9 @@ class BackgammonGame : public Game {
 
   // There is arbitrarily chosen number to ensure the game is finite.
   int MaxGameLength() const override { return 1000; }
+  // It is possible to have only chance nodes in the game. As the game
+  // is limited, apply the same limit.
+  int MaxChanceNodesInHistory() const override { return 1000; }
 
   int NumPlayers() const override { return 2; }
   double MinUtility() const override { return -MaxUtility(); }
--- a/open_spiel/games/backgammon.h
+++ b/open_spiel/games/backgammon.h
@@ -266,7 +266,7 @@ class BackgammonGame : public Game {
   int MaxGameLength() const override { return 1000; }
   // It is possible to have only chance nodes in the game. As the game
   // is limited, apply the same limit.
-  int MaxChanceNodesInHistory() const override { return 1000; }
+  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
 
   int NumPlayers() const override { return 2; }
   double MinUtility() const override { return -MaxUtility(); }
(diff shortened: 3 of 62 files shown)


lanctot:
Changing `MaxChanceNodesInHistory` to have a default implementation that errors "unimplemented". We cannot make the API not backward-compatible at this point.

michalsustr:
SGTM

lanctot:
Actually we have a few changes to request, which are easier to do here.

open_spiel/spiel.h
```diff
@@ -757,6 +757,30 @@ class Game : public std::enable_shared_from_this<Game> {
   // of chance nodes are not included in this length.
   virtual int MaxGameLength() const = 0;
 
+  // The maximum number of chance nodes occurring in any history of the game.
+  // This is typically something like the number of times dice are rolled.
+  virtual int MaxChanceNodesInHistory() const = 0;
```
lanctot, line 762: Can you add a default implementation here rather than making pure virtual. It can simply call SpielFatalError. I think Ed will have another suggestion for this call point as well.

open_spiel/game_transforms/coop_to_1p.h
```diff
@@ -163,6 +163,8 @@ class CoopTo1pGame : public Game {
   int NumPlayers() const override { return 1; }
   std::vector<int> ObservationTensorShape() const override;
   int MaxGameLength() const override;
+  // TODO() make a tighter bound.
```
lanctot, line 166: Can you re-format all the TODOs as the following (notice no parentheses): 

TODO: verify whether this bound is tight and/or tighten it.

open_spiel/integration_tests/api_test.py
```diff
@@ -284,6 +284,29 @@ def setUpClass(cls):
             depth_limit=7).values())
     cls.game_type = cls.game.get_type()
 
+  def test_sequence_lengths(self):
+    max_history_len = self.game.max_history_length()
+    max_move_number = self.game.max_move_number()
+    max_chance_nodes_in_history = self.game.max_chance_nodes_in_history()
```
lanctot, line 290: Can you put this into a try / except block, and just return if max_chance_nodes_in_history fails?

open_spiel/game_transforms/coop_to_1p.h
lanctot, line 166: Can you reformat all the TODOs as the following (notice no parentheses after the TODO):

TODO: verify whether bound is tight and/or tighten it.

lanctot (changes requested):

lanctot:
@elkhrt can you add your comments too?

open_spiel/games/backgammon.h
```diff
@@ -264,6 +264,9 @@ class BackgammonGame : public Game {
 
   // There is arbitrarily chosen number to ensure the game is finite.
   int MaxGameLength() const override { return 1000; }
+  // It is possible to have only chance nodes in the game. As the game
+  // is limited, apply the same limit.
+  int MaxChanceNodesInHistory() const override { return 1000; }
```
lanctot, line 270: Should return MaxGameLength() here

lanctot (changes requested):

open_spiel/game_transforms/coop_to_1p.h
```diff
@@ -163,6 +163,8 @@ class CoopTo1pGame : public Game {
   int NumPlayers() const override { return 1; }
   std::vector<int> ObservationTensorShape() const override;
   int MaxGameLength() const override;
+  // TODO: verify whether this bound is tight and/or tighten it.
+  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
```
elkhrt, line 167: get this from game_

open_spiel/game_transforms/repeated_game.h
```diff
@@ -65,6 +65,8 @@ class RepeatedGame : public SimMoveGame {
                const GameParameters& params);
   std::unique_ptr<State> NewInitialState() const override;
   int MaxGameLength() const override { return num_repetitions_; }
+  // There aren't chance nodes in this game.
+  int MaxChanceNodesInHistory() const override { return 0; }
```
elkhrt, line 69: I think this is wrong

open_spiel/games/backgammon.h
```diff
@@ -264,6 +264,9 @@ class BackgammonGame : public Game {
 
   // There is arbitrarily chosen number to ensure the game is finite.
   int MaxGameLength() const override { return 1000; }
+  // It is possible to have only chance nodes in the game. As the game
```
elkhrt, line 267: This is not possible in backgammon, iiuc

open_spiel/games/bridge.h
```diff
@@ -233,6 +233,9 @@ class BridgeGame : public Game {
     return UseDoubleDummyResult() ? kMaxAuctionLength
                                   : kMaxAuctionLength + kNumCards;
   }
+  // TODO: make a tighter bound.
+  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
```
elkhrt, line 237: kNumCards

open_spiel/games/bridge_uncontested_bidding.h
```diff
@@ -225,6 +225,8 @@ class UncontestedBiddingGame : public Game {
     return {kStateSize};
   }
   int MaxGameLength() const override { return kNumActions; }
+  // TODO: verify whether this bound is tight and/or tighten it.
+  int MaxChanceNodesInHistory() const override { return MaxGameLength(); }
```
elkhrt, line 229: 1

open_spiel/games/chess.h
```diff
@@ -242,6 +242,8 @@ class ChessGame : public Game {
     return chess::ObservationTensorShape();
   }
   int MaxGameLength() const override { return chess::MaxGameLength(); }
+  // There aren't chance nodes in this game.
```
elkhrt, line 245: Let's put this in in the base Game - if type is deterministic, return 0 for MaxChanceNodes

open_spiel/spiel.h
```diff
@@ -757,6 +757,34 @@ class Game : public std::enable_shared_from_this<Game> {
   // of chance nodes are not included in this length.
   virtual int MaxGameLength() const = 0;
 
+  // The maximum number of chance nodes occurring in any history of the game.
+  // This is typically something like the number of times dice are rolled.
+  virtual int MaxChanceNodesInHistory() const {
```
elkhrt, line 762: This can return 0 if the game is deterministic

open_spiel/spiel.h
```diff
@@ -757,6 +757,34 @@ class Game : public std::enable_shared_from_this<Game> {
...
+  // The maximum number of moves in the game. The value State::MoveNumber()
+  // must never be higher than this value.
+  virtual int MaxMoveNumber() const {
+    return MaxGameLength() + MaxChanceNodesInHistory();
+  }
+
+  // The maximum length of any history in the game.
+  // The value State::History().size() must never be higher than this value.
+  virtual int MaxHistoryLength() const {
+    if (GetType().dynamics == GameType::Dynamics::kSimultaneous) {
```
elkhrt, line 777: Doesn't this belong in SimMoveGame? It's a particular implementation choice.

open_spiel/game_transforms/repeated_game.h
michalsustr: Based on https://github.com/deepmind/open_spiel/blob/master/open_spiel/game_transforms/repeated_game.cc#L188

michalsustr: I will drop it, as now we change it to return 0 if game is deterministic by default.

open_spiel/games/backgammon.h
michalsustr: Then maybe something is wrong with the implementation. I was able to expand the tree to history [0, 0, 0, ... ] = [0] * 1000

open_spiel/spiel.h
michalsustr: I think currently sim move games are not required to derive from SimMoveGame (see goofspiel). This is more generic implementation. I'd leave it here in one place rather than in many places over the codebase.

michalsustr:
There are 2 comments unresolved, but otherwise I implemented all the changes.

open_spiel/games/backgammon.h
lanctot: This is strictly true because of how the tie cases are currently handled for determining first player. We should just change this to not roll doubles for determining who starts. I will fix it this week.

But that was not the reason for choosing MaxGameLength here, at least not initially. That was because in Backgammon there is once chance move per decision move. 

So let's keep this as is for now snd when I fix the impl, I will remove/change this comment

open_spiel/spiel.h
lanctot: I'm ok with this. @elkhrt , does it work for you?

lanctot: Would like to include this in tomorrow's update, if possible.

PR #374, opened 2020-09-24
Updates games list with more details

andyljones:
I was having trouble finding games appropriate for the project I had in mind, and in the process of figuring out what I was after I put together a table of properties for the games in OpenSpiel. 

The table was a bit too dense to be readable, so here I've expanded it out into a list. 

[Preview](https://github.com/andyljones/open_spiel/blob/master/docs/games.md).


lanctot:
Thanks! Will briefly discuss with the team and get back to you.

lanctot:
I quite like having the extra information, this is really cool. But I also like the kind of quick bird's eye view of all the games in one place. 

How do you feel about a table with just names and statuses at the top followed by your extended list with all the extra info below it? You can even get fancy and add a http link in the table so we can click to get to the details below.

What do you think?

andyljones:
Good idea! I've knocked something together - check the [preview again](https://github.com/andyljones/open_spiel/blob/master/docs/games.md).

lanctot:
Beautiful! I love it, and looks even nicer having the game names as headings. I'll import it today, thanks for the contribution!

yukarinoki:
This merge (commit 37aed97) makes Go a non-deterministic game, but is this correct? 
My common sense tells me that it is a deterministic game, just like chess.

(I'm not familiar with the circumstances specific to this library.)

andyljones:
Nope, that looks like a mistake on my behalf. Sorry, and thanks for catching it!

I can put in a PR, but for removing one word it feels like overkill. Could the next repo owner passing through do me a favour and make the edit for me?

lanctot:
Nice spot!

Very easy for us to fix on our end, but I have a small preference for a PR (despite agreeing that it's overkill) so that @yukarinoki gets the public recognition for pointing this out.

@yukarinoki, would you like to submit a PR? (since it's so small in this case you can do it entirely from the github web interface)? If not, we'll happily do it on our side.

yukarinoki:
@lanctot @andyljones  
Thanks for the reply.

I have just sent a pull request #466.

Issue #373, opened 2020-09-23
Feature: Bot registration API


michalsustr:
It would be great if we could have a bot API similar to the one for games, such as 
```
std::unique_ptr<Bot> bot = LoadBot("xyz");
std::vector<std::string> RegisteredBots();
```

We could also add the following method to the `Bot` interface:
```
bool CanPlayGame(const Game&);
```

The bots right now are disorganized -- Roshambo is in `bots/`, but xinxin is illogically in `games/hearts/`. I can put it together more nicely. This will also provide a structure for adding other external bots in the future, like Stockfish. What do you think?

lanctot:
I think this is a neat idea, would make running tournaments (with customized bot e.g. built outside the lib, i.e. for course projects etc.) easier too.

@elkhrt @jblespiau wdyt?

lanctot:
Will be merged by https://github.com/deepmind/open_spiel/pull/384, so closing for now.

Issue #335, opened 2020-09-07
Question: use case for Observer params?


michalsustr:
Hi @elkhrt , what is the use case for `GameParameters params` in `Game::MakeObserver`? https://github.com/deepmind/open_spiel/blob/master/open_spiel/spiel.h#L792

AFAIK it is currently not used, and these parameters should be accessible via Game::GetParameters(). I think this implies the observer would fill tensors for a different game than is actually played.

elkhrt:
These are params for the observer, not for the game. The type name is
unfortunate.
For example, in bridge one might like to have the option to include
double-dummy results in the observation (or not), which could be controlled
by a flag in here.

michalsustr:
Makes sense. Maybe it should be renamed then to `Parameter`, and we can keep GameParameter as an alias for backwards compatibility? Also I added a better comment.

PR #384, opened 2020-09-26
Add Bot API. Register basic bots.

michalsustr:
Adds #373

There is a number of other bots in the library (for example MCTS), but they are more complicated to construct from default params, so I left out their registration at the moment.

I reused the `GameParameter` from games for bots. As pointed out in #335 , it's not a great name, and I think we should rename it to something like `Parameter` or `TypedParameter`, and alias it back for backwards compatibility.

### DIFF
--- a/open_spiel/spiel_bots.cc
+++ b/open_spiel/spiel_bots.cc
@@ -16,6 +16,7 @@
 
 #include <memory>
 #include <random>
+#include <string>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -32,13 +33,8 @@ class UniformRandomBot : public Bot {
  public:
   UniformRandomBot(Player player_id, int seed)
       : player_id_(player_id), rng_(seed) {}
-  explicit UniformRandomBot(GameParameters params)
-      : Bot(std::move(params)),
-        player_id_(ParameterValue<int>("player", 0)),
-        rng_(ParameterValue<int>("seed", 0)) {}
   ~UniformRandomBot() = default;
 
-  bool CanPlayGame(const Game& game) override { return true; }
   void RestartAt(const State&) override {}
   Action Step(const State& state) override {
     return StepWithPolicy(state).second;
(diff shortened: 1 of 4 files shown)


bartdevylder:
Hi @michalsustr 

very interesting idea of adding a central bot registry! Considering the bots I've worked with I'd like to propose a slight change in structure and check what you think:

Most of the bots (external to openspiel)  I work with actually take at least a Game (and player_id) in their constructor (and in openspiel at least StatefulRandomBot and MCTSBot also do this) which would be hard to fit into the current code.

So I would propose to augment the 'load_bot' arguments at with a Game& (and maybe player_id) argument:
std::unique_ptr<Bot> LoadBot(const std::string& bot_name,
        		      const Game& game,
			      const player_id,
                              const GameParameters& params)
                              
and also to move the CanPlayGame() method from the Bot to it's factory. So instead of a factory just being a function
returning a bot, I'd make it a class e.g. like:
class BotFactory {
 public:
  virtual bool CanPlayGame(const Game& game);  
  virtual std::unique_ptr<Bot> Create(const Game& game, int player_id, const GameParameters& bot_params);
}
which would also allow to implement BotsThatCanPlayGame without having to work with a 'throwaway_bot'.
Could this work for the usecases you had in mind?

michalsustr:
Yes definitely! I will update the PR.

open_spiel/spiel_bots.cc
```diff
@@ -179,10 +200,112 @@ std::unique_ptr<Bot> MakeFixedActionPreferenceBot(
...
+    SpielFatalError(absl::StrCat("Unable to create bot: ", bot_name));
+  }
+  return result;
+}
+std::vector<std::string> BotsThatCanPlayGame(const Game& game,
+                                             Player player_id) {
+  return BotRegisterer::CanPlayGame(game, player_id, {});
+}
+std::vector<std::string> BotsThatCanPlayGame(const Game& game, Player player_id,
+                                             const GameParameters& params) {
```
bartdevylder, line 307: As noted in another comment I think it's preferable that a bot's ability to play a game is independent of the specific bot parameters. This would also solve the ambiguity of the bot parameters in this method, because I think the specific  parameters supported depend on the type of bot, so it's not clear how this can work in general across bots.

bartdevylder (changes requested):
Thanks for the changes!

open_spiel/spiel_bots.h
```diff
@@ -160,6 +173,62 @@ std::unique_ptr<Bot> MakePolicyBot(const Game& game, Player player_id, int seed,
...
+class BotRegisterer {
+ public:
+  BotRegisterer(const std::string& bot_name,
+                std::unique_ptr<BotFactory> factory);
+
+  static std::unique_ptr<Bot> CreateByName(
+      const std::string& bot_name, std::shared_ptr<const Game> game,
+      Player player_id, const GameParameters& params);
+  static std::vector<std::string> CanPlayGame(
+      const Game& game, Player player_id, const GameParameters& params);
```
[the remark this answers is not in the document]
michalsustr: I had one use case in mind: policy bot from an offline strategy. The bot would read a few lines from the supplied file (in the params) and check if it corresponds to the same game. I can't think of another use case though. Maybe such a policy bot should not be exposed via the API and you can just call it directly like `MakePolicyBot(game, policy_filename)`

open_spiel/spiel_bots.cc
michalsustr: Yes, I agree. I mentioned the original motivation in the other comment.

open_spiel/spiel_bots.h
```diff
@@ -160,6 +172,66 @@ std::unique_ptr<Bot> MakePolicyBot(const Game& game, Player player_id, int seed,
...
+ public:
+  BotRegisterer(const std::string& bot_name,
+                std::unique_ptr<BotFactory> factory);
+
+  static std::unique_ptr<Bot> CreateByName(
+      const std::string& bot_name, std::shared_ptr<const Game> game,
+      Player player_id, const GameParameters& params);
+  static std::vector<std::string> CanPlayGame(
+      const Game& game, Player player_id);
+  static std::vector<std::string> CanPlayGame(const Game& game);
```
bartdevylder, line 190: Method name 'CanPlayGame' is a bit confusing here as it does not return a boolean. Maybe rename to BotsThatCanPlayGame to make it consistent?

bartdevylder (changes requested):
Thanks for the changes. Looks really good. Looks ready to me with the method rename (see comment).

michalsustr:
Sorry, I noticed the requested change only now.

Issue #382, opened 2020-09-26
`DeserializeGame` fails on double `GameParameters` that represent integer values


gabrfarina:
Hit as part of implementing the Battleship game (#376).

`DeserializeGame` fails to deserialize games that have (at least) one double game parameter whose value corresponds to an integer. While it's easy to trigger this failure by hand, it is important to note that it will also be triggered automatically by `TestSerializeDeserialize`, which automatically tries to serialize and then deserialize a game --- in fact, that's how I hit the issue.

The issue is due to the fact that the parser for game parameters (snippet below) is *too eager* to match integers: any sequence composed uniquely of characters "+-0123456789" is immediately parsed as an integer.
https://github.com/deepmind/open_spiel/blob/91da643188fcacd7a57bf5a64ee392d6c919dcfc/open_spiel/game_parameters.cc#L168-L181

Unfortunately, it is not clear to me how to fix this issue:
* Asking that all double values always include either 'e' or a period '.' is very non-user-friendly, and would still require changes in the serialization code, which currently serializes a double like "2.0" to just "2". I think that imposing a restriction on how doubles are serialized is going to impose unnecessary burden to users.
* On the other hand, it seems like the parsing of game parameters is not done taking into account the desired resulting types. The type of the game parameters is determined lexically at game deserialization time, and the deserialized parameters are them matched against the expected schema by the `ValidateParams` method.

I can see two ways to fix the issue:
1. Either have `DeserializeGame` and `GameParameterFromString` have access to information about the schema of game parameters they should parse. This seems the cleanest solution, but would require some heavy reworking of some internal pipelines from what I can see.
2. Have `ValidateParams` include some failure recovery strategy that will promote integers to doubles instead of failing. This has the advantage that the fix would be contained to a single method. But it seems very hacky: for example, if a string parameter has the misfortune of being the string `"true"`, the parameter would currently be parsed as `bool`, even if the schema says that it should be string.

I think as a first step towards solving this, the maintainers should agree on what solution strategy to adopt. I am happy to help with whatever you might determine is best.

This issue is currently a blocker for the implementation of the Battleship game, which includes a double game parameter whose value will often be integer --- as is 2.0 by default.

gabrfarina:
A couple more thoughts about why option 2 above is very hacky:
* The current parser fails to parse values represented in exponential notation. For example, "1e2" would be parsed as a string instead of the double value "100.0". So, if we were to go with option 2, we would need to consider a promotion path from string to double in some cases.
[...]

PR #386, opened 2020-09-26
Force a point when serializing a double game param

gabrfarina:
This is just a bandaid solution to address some quirks in the way
game parameters are serialized and parsed.

See: #382
Unblocks: #376 

I have added some tests to make sure the parser's behavior does not
change accidentally.

Should I rebase onto #381 once it lands?

### DIFF
--- a/open_spiel/game_parameters.cc
+++ b/open_spiel/game_parameters.cc
@@ -58,18 +58,18 @@ std::string GameParameter::ToString() const {
     case Type::kInt:
       return absl::StrCat(int_value());
     case Type::kDouble: {
-      // XXX(maintainers): We make sure to include a '.0' for doubles that hold
-      //      integer values to avoid deserializing them as integers (see
-      //      implementation in `GameParameterFromString`).
+      // XXX: We make sure to include a '.0' for doubles that hold
+      //     integer values to avoid deserializing them as integers (see
+      //     implementation in `GameParameterFromString`).
       //
-      //      See also: #382.
+      //     See also: #382.
       //
       //
       // We cannot use StrCat as that would default to exponential notation
       // sometimes. For example, the default format of 10^-9 is the string
       // "1e-9". For that reason, we use std::stringstream here.
       //
-      // FIXME(maintainers): Avoid hardcoding the number of digits after the
+      // TODO: Avoid hardcoding the number of digits after the
       //     point if possible?
       std::stringstream ss;
       ss << std::fixed << std::setprecision(10) << double_value();
--- a/open_spiel/game_parameters.cc
+++ b/open_spiel/game_parameters.cc
@@ -14,8 +14,10 @@
 
 #include "open_spiel/game_parameters.h"
 
+#include <iomanip>
 #include <list>
 #include <map>
+#include <sstream>
 #include <string>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


gabrfarina:
Some playthrough tests fail as a result of the new serialization code. I will look into it.

@lanctot What is the backwards compatibility story of OpenSpiel? Currently this patch would guarantee that anything that was successfully parising (deserializing) before would continue doing so, and result in the same game. Is that good enough?

lanctot:
You almost certainly need to regenerate the playthroughs, I would guess. BTW I think we might have used semicolon as a special token in serializing.. if you used that for your delimiter in your lists, could cause a problem. I have used hyphens to do this in some games (negotiation, I think). 

Yes, that is good enough. Master branch is expected to move, and we are doing releases for people who want greater stability (or they should fork). Of course this is all within reason, we don't arbitrarily break backward compatibility in the core API, even on master, without some amount of warning.. because that is unnecessarily disruptive, and in any case tend to do that only if it's absolutely necessary.

gabrfarina:
Great, thanks for the clarifications. I have fixed the playthrough tests by simply amending the `GameParameters` line in the txt files to always include 10 digits after the point. I think that's better than regenerating the playthrough, because this we we get to see if this patch affects the playthrough files in other surprising ways that we were not accounting for.

About semicolons: that's unfortunate! Using dash seems like a dangerous idea for lists of doubles (like the ship values), where the dash could represent a minus in front of the double. At this point perhaps should I use spaces to separate the items in the lists?

lanctot:
If the tests passed then I am either misremembering, or the serialization code manually escapes the existing semicolons, so this is fine!

gabrfarina:
Travis is green :rocket: 

What's your rebase policy? Should we just wait for #381 to land to `master` and then rebase? Or are you planning to roll them up together?

lanctot:
Great, importing now. I can pul them both in, but I will need someone else to review it. I assume this one will be fairly quick, Battleship might be longer. Hoping to get them both into Monday's update but the internal reviews might only come on Monday too.

open_spiel/game_parameters.cc
```diff
@@ -55,8 +57,24 @@ std::string GameParameter::ToString() const {
...
+      //      implementation in `GameParameterFromString`).
+      //
+      //      See also: #382.
+      //
+      //
+      // We cannot use StrCat as that would default to exponential notation
+      // sometimes. For example, the default format of 10^-9 is the string
+      // "1e-9". For that reason, we use std::stringstream here.
+      //
+      // FIXME(maintainers): Avoid hardcoding the number of digits after the
```
lanctot, line 72: Can you change this to TODO: Avoid hardcoding...  

It causes a problem and hence blocks the import. If you have it anywhere else, can you do the same? Thanks.

lanctot (changes requested):

open_spiel/game_parameters.cc
lanctot: To be clear, the problem is the FIXME. It needs to be TODO. No need to have anyone in parentheses.

gabrfarina: Ah that's cool that you have some `FIXME` hooks :-) I will move that `FIXME` to a `TODO` and remove the (maintainers) in a minute. I will keep the `XXX`s in place though without changing them to `TODO`s, since they are more like important considerations than stuff that needs to be done.

gabrfarina:
Done. I will not cherrypick this last commit onto the battleship branch for now, as I guess I'll just end up rebasing `battleship` onto `master` when this PR lands (on Monday hopefully if I understood correctly?)

lanctot:
Had to make a few edits because try / catch does not work as expected internally. But looking good. This means the master branch will have some local modifications from your Battleship PR, which you'll need to merge or rebase before we go ahead with Battleship.

lanctot (approved):

gabrfarina:
Perfect, thanks for the speedy review process, it's a pleasure to contribute to code bases this well managed :-)

The battleship branch is currently based on master. These fixes to the game parameters code were just cherry picked. I didn't want to rebase battleship onto this pr's branch while it still had outstanding code comments from you ---- I don't know what github would do if I were to push force after the rebase, and I didn't want to risk losing your comments /review.

When this lands I will simply rebase the battleship branch onto the head of master and drop the cherry picked commits.

PR #388, opened 2020-09-26
Expose game-specific APIs -- add a basic skeleton.

michalsustr:
This PR adds exporting game specific details so that we can query them in Python. This is useful for example for creating game-specific bots or UIs. This is a work-in-progress, I just wanted to know what you think about doing this in general. 

The code in this PR already allows for game specific rendering of the current state for Chess:

image

### DIFF
--- a/open_spiel/python/CMakeLists.txt
+++ b/open_spiel/python/CMakeLists.txt
@@ -13,6 +13,8 @@ set(PYTHON_BINDINGS ${PYTHON_BINDINGS}
   pybind11/algorithms_trajectories.h
   pybind11/bots.cc
   pybind11/bots.h
+  pybind11/games.h
+  pybind11/games.cc
   pybind11/games_bridge.cc
   pybind11/games_bridge.h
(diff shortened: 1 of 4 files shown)


elkhrt (commented):
I don't think putting every game in the same file is the right approach.
We already have:
  init_pyspiel_games_negotiation(m);        // Negotiation game.
  init_pyspiel_games_bridge(m);  // Game-specific functions for bridge.

I suggest copying that approach for now.
If we wanted to do something more general, adding an optional pybind initialization method to the game registration mechanism seems like the best route.

michalsustr:
Ah, I should've looked at those bindings! :)

> If we wanted to do something more general, adding an optional pybind initialization method to the game registration mechanism seems like the best route.

Then you'd have to #include pybind conditionally on BUILD_WITH_PYTHON in the game_xxx.h / game_xxx.cc files which seems like too much hastle and it would clutter the files. I agree, it's better to stick to adding it one by one in pyspiel.cc for now.

PR #393, opened 2020-10-01
Implement an efficient CFR solver.

michalsustr:
This PR is a work-in-progress for implementing efficient CFR solvers for infostate trees. 
@elkhrt @gabrfarina I would appreciate your feedback on the work.

- [x] Implement infostate tree representation.
- [x] Make sure the construction works also for a sequence of states where the same player chooses actions repeatedly. (correct alternation of decision/observation nodes, so that the number of decision nodes' children corresponds to the number of actions in those nodes). Also, the child index in the vector should correspond to the legal action index. 
- [x] Add more infostate tree tests. I am not sure how to make them so they are not brittle. Testing tensor observations is not a good idea at the moment, as they change quite a bit right now.
- [x] Implement CFR that will build infostate trees for both players, and then run iterations only on these lightweight trees.
- [x] Run this CFR on the same suite of tests as the current implementation.

I expect that this CFR implementation will be much faster, as making rollouts in the game is slow and locating CFRInfoStateValues in a map indexed by strings is nice for understanding the algorithm, but not for efficiency.

Some things I decided not to do (in this PR):
- Allow to prune the observation nodes. Because we cannot be sure about the construction of the infostate tree until it is finished, some observation nodes can be redundant. We can optionally prune them to get a smaller tree.
- Identify infoste nodes by a sequence (sequence-form). Not sure whether just having a vector if ints is good enough, but if more info is needed you can always use these to traverse the infostate trees.
- Compute sequence tuples for any State. I don't need that for my purposes, but I can add it if @gabrfarina would like to have it.

I outlined a number of welcome contributions in the code.

### DIFF
--- a/open_spiel/algorithms/infostate_cfr.h
+++ b/open_spiel/algorithms/infostate_cfr.h
@@ -112,8 +112,8 @@ class InfostateCFR {
   void RunSimultaneousIterations(int iterations);
   void RunAlternatingIterations(int iterations);
 
-  void PrepareReachProbs();
-  void PrepareReachProbs(Player pl);
+  void PrepareRootReachProbs();
+  void PrepareRootReachProbs(Player pl);
   void EvaluateLeaves();
   void EvaluateLeaves(Player pl);
 
--- a/open_spiel/algorithms/infostate_cfr.h
+++ b/open_spiel/algorithms/infostate_cfr.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef OPEN_SPIEL_ALGORITHMS_INFOSTATE_TREE_CFR_H_
-#define OPEN_SPIEL_ALGORITHMS_INFOSTATE_TREE_CFR_H_
+#ifndef OPEN_SPIEL_ALGORITHMS_INFOSTATE_CFR_H_
+#define OPEN_SPIEL_ALGORITHMS_INFOSTATE_CFR_H_
 
 #include <map>
 #include <memory>
@@ -145,4 +145,4 @@ class InfostateCFR {
 }  // namespace algorithms
 }  // namespace open_spiel
 
-#endif  // OPEN_SPIEL_ALGORITHMS_INFOSTATE_TREE_CFR_H_
+#endif  // OPEN_SPIEL_ALGORITHMS_INFOSTATE_CFR_H_
--- a/open_spiel/algorithms/infostate_cfr.h
+++ b/open_spiel/algorithms/infostate_cfr.h
@@ -67,6 +67,7 @@ struct InfostateTreeValuePropagator {
   // Mutable values to keep track of.
   std::vector<float> reach_probs;
   std::vector<float> cf_values;
+  float root_cf_value;
 
   static void CollectTreeStructure(
       CFRNode* node, int depth,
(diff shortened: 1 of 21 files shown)


michalsustr:
Uff, this turned out to be bigger than I thought. I can split it into several pieces for review, if you'd like.

- a. Infostate trees
- b. Infostate CFR
- c. Action view (for simultaneous move nodes)

Deps: 
```
b
| \
a  c
```

elkhrt (commented):
Could you revert the changes to the other algorithms and just add the new algorithm in this PR?

open_spiel/algorithms/infostate_cfr.h
```diff
@@ -0,0 +1,156 @@
...
+// If you decide to make contributions to this code, please open up an issue
+// on github first. Thank you!
+
+namespace open_spiel {
+namespace algorithms {
+
+// A helper struct that allows to propagate reach probs / cf values
+// up and down the tree.
+struct InfostateTreeValuePropagator {
+  // Tree and the tree structure information. These must not change!
```
elkhrt, line 61: This seems ugly. Shouldn't these values be parameters to the propagation function rather than members here?

open_spiel/utils/action_view.h
```diff
@@ -0,0 +1,70 @@
...
+
+// Provides a number of iterators that are useful for dealing 
+// with simultaneous move nodes.
+struct ActionView {
+  const Player current_player;
+  const std::vector<std::vector<Action>> legal_actions;
+  // Collects legal actions at the specified state.
+  ActionView(const State& state);
+
+  // Provides an iterator over flattened actions. This is equivalent to calling
```
elkhrt, line 34: This looks like unnecessary duplication.

open_spiel/integration_tests/api_test.py
```diff
@@ -751,3 +751,9 @@ def load_tests(loader, tests, pattern):  # pylint: disable=invalid-name,g-doc-ar
 
 if __name__ == "__main__":
   absltest.main()
+
+# TODO checks
```
elkhrt, line 755: Please make these TODOs clearer. You can link to github issues if you want, or just write a longer description of each in-line here.

open_spiel/algorithms/infostate_cfr.h
```diff
@@ -0,0 +1,156 @@
...
+        : cfr_(cfr), infostate_table_(cfr_.InfoStateValuesPtrTable()) {}
+    ActionsAndProbs GetStatePolicy(
+        const std::string& info_state) const override;
+  };
+  std::shared_ptr<Policy> AveragePolicy() const {
+    return std::make_shared<InfostateCFRAveragePolicy>(*this);
+  }
+
+ private:
+  std::array<InfostateTreeValuePropagator, 2> propagators_;
```
elkhrt, line 141: What's the array for?

open_spiel/algorithms/infostate_cfr.h
```diff
@@ -0,0 +1,156 @@
...
+  // Map from player 1 index (key) to player 0 (value).
```
elkhrt, line 142: Unclear.

michalsustr:
Thank you for the comments! I will go over them. 

Since I have a lot of other code piled up locally, I was thinking about how to incorporate it efficiently. I came up with the idea that I split them up into smaller PRs that are better structured. The code has mostly a linear chain of dependence, and I will post new PRs as we move forward the chain. This will allow us to incorporate the review feedback and have a cleaner git history.  I appreciate the quality code review, therefore please do not feel under a rush for reviewing the PRs. This change will only mean that the contributions will come slower from our side.

I will close the PR for now, but I will address the comments in the new smaller PRs. Thanks again!

Issue #401, opened 2020-10-09
Implement `GetAllHistories`


finbarrtimbers:
See discussion in https://github.com/deepmind/open_spiel/issues/399. 

Basically, we would like it if someone copied the implementation of [`GetAllStates`](https://github.com/deepmind/open_spiel/blob/de2c9ef57038829c2709becb0104e3b2d76725cc/open_spiel/algorithms/get_all_states.cc#L47-L53) but made two changes: 

1) Use `State::FullHistory` to check for equality, rather than `State::ToString`
2) Return a `std::vector<std::unique_ptr<State>>` rather than a `std::map<std::string, std::unique_ptr<State>>`.

lanctot:
Just a quick follow-up, this came up because currently the example in `count_all_states.cc` will not return the right set of information states for perfect information games.

Specifically, here: https://github.com/deepmind/open_spiel/blob/de2c9ef57038829c2709becb0104e3b2d76725cc/open_spiel/examples/count_all_states.cc#L42

This iterates over the states returned by `GetAllStates` which currently uses the ToString keys as uniqueness for states, whereas the information states are generally a function of the states' histories.

jhtschultz:
Took a quick crack at this and got it working. I have to be out for the next few hours but I can have a PR ready later tonight.

lanctot:
> Took a quick crack at this and got it working. I have to be out for the next few hours but I can have a PR ready later tonight.

Awesome! Thanks, @jhtschultz !

gabrfarina:
Should we consider moving the implementation of the tests for #376 to use `GetAllHistories` instead?

lanctot:
> Should we consider moving the implementation of the tests for #376 to use `GetAllHistories` instead?

Depends on what you did in those tests, I didn't have a close look. @finbarrtimbers ? 

In any case, if so please wait until next week because I've imported it already and it's going in Monday morning.

gabrfarina:
Ah, I was just counting infosets and sequences for each players. But it seems that "semantically" in order to do that you want to iterate over all histories (i.e., tree nodes), instead of all the states.

I imagine that the code that counts infosets under open_spiel/examples/ will move to GetAllHistories as well, so changing the tests for Battleship would align the implementations.

lanctot:
Oh yes that makes perfect sense! If John submits the PR before Monday I will import it and include in the update too.

finbarrtimbers:
[...]

Issue #399, opened 2020-10-07
`GetAllStates` relies on the the unreliable state string instead of the history


gabrfarina:
Hit as part of #376 .

+@lanctot +@michalsustr +@finbarrtimbers

The implementation of `GetAllStates` expects that `state->ToString()` be unique for states, which is not true. For example, in the Battleship game, `ToString()` returns an ascii art representation of the player boards. In Go, my understanding is that `ToString()` renders the Go board (**edit** see also [this](https://github.com/deepmind/open_spiel/pull/376#discussion_r496431791) discussion). Neither of these mappings are injective: different orders of moves can result in the same boards, and therefore the same `ToString()`.

https://github.com/deepmind/open_spiel/blob/de2c9ef57038829c2709becb0104e3b2d76725cc/open_spiel/algorithms/get_all_states.cc#L47-L53

This results in the method failing to explore the full game tree. Since the method is performing a DFS on a tree, it is impossible to hit the same state twice, so I am not sure what the comment `add only if not already present` was supposed to catch. I suggest the following:
- If we want to keep the map, we should have histories as key.
- There is no need to keep an expensive map: a vector of (unique) pointers to states would be more efficient and simplify the code even further. I'm not sure how breaking a change that would be though?

lanctot:
That depends how you define state. :) 

The `ToString` representation is actually unique across states. If the order of moves is important for determining the state, it should also be in the `ToString`. 

I don't see how the quoted code is problematic? The traversal of the game tree still happens further below it. 

If you want a similar method bases histories, we can have another one called `GetAllHistories` but I think in your case of battleship we should just add all the shot info to the ToString (see e.g. GoofspielState::ToString). 

The ToString needs to be the complete description of the state. Having a display of the board is good: please keep it! Just add the other information necessary to describe the full state.

finbarrtimbers:
I do think that adding a `GetAllHistories` that uses the optimization you describe would be great, and changing `count_all_states` to use it would be a good change.

finbarrtimbers:
However, I think Marc's correct.

gabrfarina:
> The `ToString` representation is actually unique across states. If the order of moves is important for determining the state, it should also be in the `ToString`.

OK that was also my worry when writing the generator. However, because of [this](https://github.com/deepmind/open_spiel/pull/376#discussion_r496431791) discussion I ended up throwing away `ToPrettyString` and replacing `ToString` with that implementation.

[...]

PR #402, opened 2020-10-10
Add get_all_histories

jhtschultz:
Addresses Issue #401. As requested, `GetAllHistories()` returns a `std::vector<std::unique_ptr<State>>`.

> Use State::FullHistory to check for equality, rather than State::ToString

Is there any reason we need to actually perform this check for equality or will a DFS suffice? I couldn’t come up with a scenario in which there’s a difference so I implemented it as a straightforward DFS. Wanted to check though because there’s been some discussion around this issue and I could well have missed something.

I gave `GetAllHistories()` its own file. Seemed like the right way to go but if there’s a preference for including it in the `get_all_states` file as a separate function I’m happy to do that.

And if there’s anything else I can do for this PR just let me know :)

### DIFF
--- a/open_spiel/algorithms/CMakeLists.txt
+++ b/open_spiel/algorithms/CMakeLists.txt
@@ -25,6 +25,8 @@ add_library (algorithms OBJECT
   expected_returns.h
   external_sampling_mccfr.cc
   external_sampling_mccfr.h
+  get_all_histories.cc
+  get_all_histories.h
   get_all_states.cc
   get_all_states.h
   get_legal_actions_map.cc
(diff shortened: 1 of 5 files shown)


gabrfarina:
I'm not a maintainer, but I think there was [consensus](https://github.com/deepmind/open_spiel/issues/401#issuecomment-706400869) that the code that counts infosets in `open_spiel/examples/count_all_states.cc` would need to move to GetAllHistories instead to avoid undercounting in some classes of games (#399).

Thanks for this, once it lands I will move the tests in #376 to use it :-)

jhtschultz:
That ought to do it. Thanks @gabrfarina!

lanctot:
Oh hey John, seems like there is a conflict with master, can you double check that your last commit was included in the update? (I may have done the update before your last commit.)

lanctot:
Just checked, it was not included. Sorry, completely missed that (I modified the same file internally.)

But also: I don't think we want to replace states entirely with histories in that example because it will double count states in the perfect information, right?

I thought the plan would be to use histories to count info states? Or we can just use this example for imperfect info games (in which case let's add a check).

PR #403, opened 2020-10-10
Adds a suggested-alpha function

andyljones:
This adds a function to the alpha-rank module that'll suggest an alpha large enough for the ranking to have 'settled out'. 

The logic behind this is that the transition probabilities can be expanded as a power series, as per Eqn 8 [in the paper](https://arxiv.org/pdf/1903.01373.pdf):

When alpha is sufficiently large, then the first or the last term will dominate. That happens roughly when

This is when the first term is e^2 (approx 10x) larger than the second term. 

Empirically, it seems like this dominating-term effect is what gets the ranking to settle down, and I've started to use it as a 'default alpha'. As an exemplar, here's the sweep from Kuhn poker

```python
payoff_tables = alpharank_example.get_kuhn_poker_data(num_players=2)
alpharank.sweep_pi_vs_alpha(payoff_tables, visualize=True)
```
image

for which `alpharank.suggest_alpha(payoff_tables)` gives 10^2.

### DIFF
--- a/open_spiel/python/egt/alpharank.py
+++ b/open_spiel/python/egt/alpharank.py
@@ -792,11 +792,11 @@ def compute(payoff_tables,
 
   return rhos, rho_m, pi, num_profiles, num_strats_per_population
 
-def suggest_alpha(payoff_tables):
+def suggest_alpha(payoff_tables, tol=.1):
   """Suggests an alpha for use in alpha-rank. 
 
   The suggested alpha is approximately the smallest possible alpha such that 
-  the ranking has 'settled out'. It is calculated as 2/min_gap_between_payoffs. 
+  the ranking has 'settled out'. It is calculated as -ln(tol)/min_gap_between_payoffs. 
 
   Args:
     payoff_tables: List of game payoff tables, one for each agent identity. Each
@@ -806,8 +806,10 @@ def suggest_alpha(payoff_tables):
   Returns:
     A suggested alpha.
   """ 
+  payoffs_are_hpt_format = utils.check_payoffs_are_hpt(payoff_tables)
+
   num_strats_per_population =\
-    utils.get_num_strats_per_population(payoff_tables, False)
+    utils.get_num_strats_per_population(payoff_tables, payoffs_are_hpt_format)
   num_profiles = utils.get_num_profiles(num_strats_per_population)
 
   gap = np.inf
@@ -820,12 +822,12 @@ def suggest_alpha(payoff_tables):
 
     for index_population_that_changed, col_profile in next_profile_gen:
       payoff_table_k = payoff_tables[index_population_that_changed]
-      f_r = _get_payoff(payoff_table_k, False, col_profile, 
+      f_r = _get_payoff(payoff_table_k, payoffs_are_hpt_format, col_profile, 
                           index_population_that_changed)
-      f_s = _get_payoff(payoff_table_k, False, row_profile, 
+      f_s = _get_payoff(payoff_table_k, payoffs_are_hpt_format, row_profile, 
                           index_population_that_changed)
       if f_r > f_s:
         gap = min(gap, f_r - f_s)
 
-  return 2/gap
+  return -np.log(tol)/gap
 
--- a/open_spiel/python/egt/alpharank.py
+++ b/open_spiel/python/egt/alpharank.py
@@ -791,3 +791,41 @@ def compute(payoff_tables,
     print_results(payoff_tables, payoffs_are_hpt_format, rhos, rho_m, c, pi)
 
   return rhos, rho_m, pi, num_profiles, num_strats_per_population
+
+def suggest_alpha(payoff_tables):
+  """Suggests an alpha for use in alpha-rank. 
+
+  The suggested alpha is approximately the smallest possible alpha such that 
+  the ranking has 'settled out'. It is calculated as 2/min_gap_between_payoffs. 
+
+  Args:
+    payoff_tables: List of game payoff tables, one for each agent identity. Each
+      payoff_table may be either a numpy array, or a _PayoffTableInterface
+      object.
+  
+  Returns:
+    A suggested alpha.
+  """ 
+  num_strats_per_population =\
+    utils.get_num_strats_per_population(payoff_tables, False)
+  num_profiles = utils.get_num_profiles(num_strats_per_population)
+
+  gap = np.inf
+  for id_row_profile in range(num_profiles):
+    row_profile = utils.get_strat_profile_from_id(num_strats_per_population,
+                                                id_row_profile)
+
+    next_profile_gen = utils.get_valid_next_profiles(num_strats_per_population,
+                                                    row_profile)
+
+    for index_population_that_changed, col_profile in next_profile_gen:
+      payoff_table_k = payoff_tables[index_population_that_changed]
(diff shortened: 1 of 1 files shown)


andyljones:
On a related note, you can use this same idea to come up with a simple approximation to the alpha-rank transitions:



when $f_tau >= f_sigma$, and zero otherwise. This eliminates $m$ as a parameter, and then you've got yourself a parameter-free evaluation method that comes out really close to the full method. Again, [on Kuhn poker](https://gist.github.com/andyljones/e4d655a2ee433010d9eecc35d328026b) it comes out within .006% error of the full method.

I haven't put this simplification in the PR though because I thought tearing the guts out of alpha-rank and changing the API'd be a bit much.

lanctot:
Cool! @shayegano @dhennes can you take a look?

lanctot:
Btw @andyljones how do you get latex to render in markdown, that looks really nice! Or are they images from e.g. latex2png?

andyljones:
Images I'm afraid to say! I spent an hour or so looking for a good way to get latex into Github comments, and I eventually gave up. Best I could find was [this Chrome extension](https://chrome.google.com/webstore/detail/mathjax-plugin-for-github/ioemnmodlmafdkllaclgeombjnmnbima), but the people reading the comment need to have the same thing installed.

lanctot:
> Images I'm afraid to say! I spent an hour or so looking for a good way to get latex into Github comments, and I eventually gave up. Best I could find was [this Chrome extension](https://chrome.google.com/webstore/detail/mathjax-plugin-for-github/ioemnmodlmafdkllaclgeombjnmnbima), but the people reading the comment need to have the same thing installed.

Ah, well thanks for the effort.. it is super nice to see latex here on github! And makes explaining your points rather smooth. (Btw I use latex2png extensively when preparing presentations with Google slides! :))

lanctot:
Quick follow-up, I spoke to Daniel and he says it looks good. Shayegan is the best person to review this, and he's on vacation until the 26th, so he'll take a look when he's back.

andyljones:
Thanks for updating me!

As an aside @lanctot , this is the second entirely-minor PR I've made and the second where you've been notably enthusiastic and encouraging. It's a lovely attitude in a maintainer, and it makes me want to contribute more in future. Cheers!

open_spiel/python/egt/alpharank.py
```diff
@@ -791,3 +791,41 @@ def compute(payoff_tables,
...
+  Args:
+    payoff_tables: List of game payoff tables, one for each agent identity. Each
+      payoff_table may be either a numpy array, or a _PayoffTableInterface
+      object.
+  
+  Returns:
+    A suggested alpha.
+  """ 
+  num_strats_per_population =\
+    utils.get_num_strats_per_population(payoff_tables, False)
```
shayegano, line 810: Would you mind adding `payoffs_are_hpt_format = utils.check_payoffs_are_hpt(payoff_tables)` at the top of the function, and using this in lieu of the hard-coded False here? Thanks!

open_spiel/python/egt/alpharank.py
```diff
@@ -791,3 +791,41 @@ def compute(payoff_tables,
...
+  for id_row_profile in range(num_profiles):
+    row_profile = utils.get_strat_profile_from_id(num_strats_per_population,
+                                                id_row_profile)
+
+    next_profile_gen = utils.get_valid_next_profiles(num_strats_per_population,
+                                                    row_profile)
+
+    for index_population_that_changed, col_profile in next_profile_gen:
+      payoff_table_k = payoff_tables[index_population_that_changed]
+      f_r = _get_payoff(payoff_table_k, False, col_profile,
```
shayegano, line 823: Same as above (regarding `payoffs_are_hpt_format`)

open_spiel/python/egt/alpharank.py
```diff
@@ -791,3 +791,41 @@ def compute(payoff_tables,
...
+                                                id_row_profile)
+
+    next_profile_gen = utils.get_valid_next_profiles(num_strats_per_population,
+                                                    row_profile)
+
+    for index_population_that_changed, col_profile in next_profile_gen:
+      payoff_table_k = payoff_tables[index_population_that_changed]
+      f_r = _get_payoff(payoff_table_k, False, col_profile, 
+                          index_population_that_changed)
+      f_s = _get_payoff(payoff_table_k, False, row_profile,
```
shayegano, line 825: Same as above (regarding `payoffs_are_hpt_format`)

open_spiel/python/egt/alpharank.py
```diff
@@ -791,3 +791,41 @@ def compute(payoff_tables,
...
+    for index_population_that_changed, col_profile in next_profile_gen:
+      payoff_table_k = payoff_tables[index_population_that_changed]
+      f_r = _get_payoff(payoff_table_k, False, col_profile, 
+                          index_population_that_changed)
+      f_s = _get_payoff(payoff_table_k, False, row_profile, 
+                          index_population_that_changed)
+      if f_r > f_s:
+        gap = min(gap, f_r - f_s)
+
+  return 2/gap
```
shayegano, line 830: Since the '2' here controls the series decay rate tolerance, perhaps make it an arg (with a default value of 2, which seems sensible!)

open_spiel/python/egt/alpharank.py
```diff
@@ -791,3 +791,41 @@ def compute(payoff_tables,
     print_results(payoff_tables, payoffs_are_hpt_format, rhos, rho_m, c, pi)
 
   return rhos, rho_m, pi, num_profiles, num_strats_per_population
+
+def suggest_alpha(payoff_tables):
+  """Suggests an alpha for use in alpha-rank. 
+
+  The suggested alpha is approximately the smallest possible alpha such that 
+  the ranking has 'settled out'. It is calculated as 2/min_gap_between_payoffs.
```
shayegano, line 799: Would you mind adding a line or two of added context here (similar to the transition probability series explanation you provided in the main thread), in case users are interested in where this derivation comes from? Thanks ahead!

shayegano (changes requested):
Hi @andyljones, thanks for the contribution! The math seems to check out (nice idea to use this approximation!). I have some very minor requests below. 

Out of curiosity, have you tried this with some other games? 0.006% error is quite nice for your 2-player Kuhn example, interested to see how it holds for games with more strategies and players!

Thanks ahead (and apologies for the delayed review, still catching up on various queues after being out-of-office for a bit)

andyljones:
@shayegano I've fixed all but the readme ask 🙂

As to the readme ask and the .006% error, I got to wondering if I could do better than 'empirically it works!'. 
I've [sketched a proof that you can get away with clamping all the fixation probs to 0 or 1](https://github.com/deepmind/open_spiel/files/5441845/suggest-alpha.pdf), though there are still two TODOs to resolve. <sup>which of course mean it's *not* actually a proof yet</sup>. 

Would appreciate your thoughts on it, though there's no urgency - I'll chew those TODOs over and see if I can fix them.

shayegano (approved):

shayegano:
@andyljones Thanks for the changes (approved). Thanks also for the attached doc looking into the proof! I still have to read through it in detail, but the remaining TODO is reminiscent to one we ran across when proving some perturbation bounds in a follow-up paper. I'll send you an email now with more details!

PR #226, opened 2020-05-22
[PyTorch] DeepCFR Implementation

MJ10:
The algorithms in the library are primarily written in Tensorflow. However, given the generic nature of the rest of the library, algorithms can be written with any learning library. Adding an example demonstrating how to use PyTorch would certainly be helpful for the community, as also mentioned in the [documentation](https://openspiel.readthedocs.io/en/latest/contributing.html#roadmap-and-call-for-contributions).

In this PR I have added a minimal notebook that implements DeepCFR in PyTorch which can serve as a reference.

Feedback welcome!

PR #406, opened 2020-10-13
Force matched_regrets in DeepCFR sum to 1

findmyway:
I think this also applies to https://github.com/deepmind/open_spiel/pull/226/files

### DIFF
--- a/open_spiel/python/algorithms/deep_cfr.py
+++ b/open_spiel/python/algorithms/deep_cfr.py
@@ -352,7 +352,7 @@ class DeepCFRSolver(policy.Policy):
       for action in legal_actions:
         matched_regrets[action] = advantages[action] / cumulative_regret
     else:
-      matched_regrets[max(legal_actions), key=lambda a: _advantages[a]] = 1
+      matched_regrets[max(legal_actions, key=lambda a: _advantages[a])] = 1
     
     return advantages, matched_regrets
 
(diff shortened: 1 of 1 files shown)


lanctot:
Thanks @findmyway, nice catch!

However, I think there is a better fix. I recently discovered (that DeepCFR uses a different version of regret matching that uses probability 1 on the highest-valued regret rather than the common uniform reset case (see Section 2, the text following Eq (4) on top of page 3 in https://arxiv.org/pdf/1811.00164.pdf).

This should be an easy change: instead of using uniform here, just set them all to 0 except the one with the highest advantage. Do you mind fixing this since you're already here?

findmyway:
I've made the necessary changes.

By the way, I think the `legal_actions_mask` is not well considered during the advantage network training in DeepCFR. The current implementation tries to minimize the MSE between $r\tilde$ and $V(I)$ among all the actions. By doing so we are forcing the network to approximate the value of non-legal actions to zero. This is only a minor issue though. In PokerRL, the output of the advantage network is masked with `legal_actions_mask` first.

https://github.com/EricSteinberger/PokerRL/blob/e02ea667061b96912e424231da071b6f20a262f7/PokerRL/rl/neural/DuelingQNet.py#L55

lanctot:
Thanks. Yes, I agree, we should do it as it's done in PokerRL. Can you open an issue so we can track it (or submit a PR :))? 

I will import this one now, thanks!

findmyway:
I'll submit a PR later.

findmyway:
There may be some other improvements to the current implementation of DeepCFR. But I'm kind of busy this week. I'll try to create a PR later this weekend.

lanctot:
> There may be some other improvements to the current implementation of DeepCFR. But I'm kind of busy this week. I'll try to create a PR later this weekend.

No rush. I'll post an issue in the mean time to keep as a reminder (maybe someone else can do it too) :)

PR #410, opened 2020-10-16
External sampling mccfr in python

colibri17:
I implemented the external version of MCCFR in Python (`external_sampling_mccfr.py`). I tried to mirror the C++ version (`external_sampling_mccfr.cc`) as much as possible, taking the scheleton from the outcome version in Python (`outcome_sampling_mccfr.py`). Some code is duplicated with respect to it and maybe a super-class should be created.

I also created a module for tests (`external_sampling_mccfr_test.py`) where I taken some of the threshold values from the C++ version. I added the remaining values without any particular logic, but just using small thresholds similar to the known cases.

Finally, I changed the `mccfr_example.py` by allowing to use the new external sampling.

### DIFF
--- a/open_spiel/python/algorithms/external_sampling_mccfr.py
+++ b/open_spiel/python/algorithms/external_sampling_mccfr.py
@@ -65,10 +65,10 @@ class ExternalSamplingSolver(object):
         An iteration consists of one episode for each player as the update player.
         """
         for player in range(self._num_players):
-            state = self._game.new_initial_state()
-            self._update_regrets(state, player)
+            self._update_regrets(self._game.new_initial_state(), player)
         if self._average_type == 'k_full':
-            raise NotImplementedError
+            reach_probs = np.ones(self._num_players, dtype=np.float64)
+            self._full_update_average(self._game.new_initial_state(), reach_probs)
 
     def _lookup_infostate_info(self, info_state_key, num_legal_actions):
         """Looks up an information set table for the given key.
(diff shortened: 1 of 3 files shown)


colibri17:
@googlebot I fixed it.

colibri17:
@googlebot I fixed it.

lanctot:
Cool, thanks @colibri17 ! I will take a look early this week.

colibri17:
You're welcome! That's great @lanctot thank you.

open_spiel/python/algorithms/external_sampling_mccfr.py
```diff
@@ -0,0 +1,236 @@
...
+_REGRET_INDEX = 0
+_AVG_POLICY_INDEX = 1
+
+
+class ExternalSamplingSolver(object):
+    """
+    An implementation of external sampling MCCFR
+    """
+
+    def __init__(self, game, average_type='k_simple'):
```
lanctot, line 34: Can we make the average_type an enum defined at the top of this file?

open_spiel/python/algorithms/external_sampling_mccfr_test.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+from open_spiel.python import policy
+from open_spiel.python.algorithms import exploitability
+from open_spiel.python.algorithms import external_sampling_mccfr
+import pyspiel
+
+SEED = 39823987
+
+
+class OutcomeSamplingMCCFRTest(absltest.TestCase):
```
lanctot, line 31: Rename to ExternalSamplingMCCFRTest

open_spiel/python/algorithms/external_sampling_mccfr_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def test_outcome_sampling_leduc_2p_simple(self):
+        np.random.seed(SEED)
+        game = pyspiel.load_game("leduc_poker")
+        os_solver = external_sampling_mccfr.ExternalSamplingSolver(game, 'k_simple')
```
lanctot, line 36: es_solver, same below

lanctot (changes requested):

lanctot:
Overall, very nice! I only have a few minor comments. I'll import it once you fix these.

open_spiel/python/algorithms/external_sampling_mccfr.py
colibri17: I am not sure I get it. Do you mean define at the top of this file something like:
```
from enum import Enum
average_type = Enum('average_type', 'k_simple k_full')
```
And then having it called in the constructor as default parameter by using:
`def __init__(self, game, average_type=average_type.k_simple.name)`?

open_spiel/python/algorithms/external_sampling_mccfr_test.py
colibri17: Yes, good point. Forgot to change the names!

open_spiel/python/algorithms/external_sampling_mccfr_test.py
colibri17: Yes, good point. Forgot to change the method names!

open_spiel/python/algorithms/external_sampling_mccfr.py
lanctot: Yes but no worries I can change it upon import.

colibri17: Ok, will let it as it is then. Thanks

colibri17:
I did another push with the fixes. Let me know if everything is ok

lanctot (approved):

lanctot:
> I did another push with the fixes. Let me know if everything is ok

Looking good, importing now.

PR #417, opened 2020-10-22
Sequence-form LP implementation

michalsustr:
Depends on #393


michalsustr:
Closing for the same reasons as in #393

PR #418, opened 2020-10-22
fixing bracket in echo

pratapram:


pratapram:
@googlebot I signed it!

pratapram:
@googlebot I signed it!

PR #420, opened 2020-10-24
DeepCFR tree traversal bug

StochasticEntropy:
When traversing the game tree, the expected value under the current strategy has to be passed up the tree, not the maximum value of the current runout. This is mentioned in https://arxiv.org/abs/1811.00164, section 4: "When a terminal node is reached, the value is passed back up. In chance and opponent infosets, the value of the sampled action is passed back up unaltered. In traverser infosets, the value passed back up is the weighted average of all action values, where action a’s weight is σ^t(I, a)."

With the same parameters as in https://arxiv.org/abs/1901.07621, this gives a NashConv of about 0.3 after 100 iterations with 1500 traversals each, which is twice as much as shown in the paper. So if they used avg_exploitability = NashConv/2, the code seems comparable, otherwise there might still be some other bugs.

I also have a TF2-implementation that uses the skip-connection and layer norm as described in the paper, as well as applying the legal_action_mask in the network training, which has also been mentioned in [pull request 406](https://github.com/deepmind/open_spiel/pull/406). These changes don't give noticable improvements in leduc though.

I am also not sure what the plan with integrating TF2 implementations is. Should I provide the code, and if yes, how?

### DIFF
--- a/open_spiel/python/algorithms/deep_cfr.py
+++ b/open_spiel/python/algorithms/deep_cfr.py
@@ -305,17 +305,19 @@ class DeepCFRSolver(policy.Policy):
       for action in state.legal_actions():
         expected_payoff[action] = self._traverse_game_tree(
             state.child(action), player)
+      cfv = 0
+      for a_ in state.legal_actions():
+        cfv += strategy[a_] * expected_payoff[a_]
       for action in state.legal_actions():
         sampled_regret[action] = expected_payoff[action]
-        for a_ in state.legal_actions():
-          sampled_regret[action] -= strategy[a_] * expected_payoff[a_]
+        sampled_regret[action] -= cfv
       sampled_regret_arr = [0] * self._num_actions
       for action in sampled_regret:
         sampled_regret_arr[action] = sampled_regret[action]
       self._advantage_memories[player].add(
           AdvantageMemory(state.information_state_tensor(), self._iteration,
                           sampled_regret_arr, action))
-      return max(expected_payoff.values())
+      return cfv
     else:
       other_player = state.current_player()
       _, strategy = self._sample_action_from_advantage(state, other_player)


StochasticEntropy:
> @googlebot I signed it!

lanctot:
Thanks @StochasticEntropy ! 

A TF2 implementation with layer norm and skip connections would awesome and very welcome. I will bring it up with the team, but I think right now the best solution would be to put it in the same place and call it deep_cfr_tf2.py explaining the differences in the header. We will likely deprecated the old ones at some point, but we would have to give some advance notice. Once we do that we can rename things appropriately.

StochasticEntropy:
Great, I don't see anything else wrong with the code, could you check with Eric from https://arxiv.org/abs/1901.07621 if they used NashConv/2 in their figure 1?

StochasticEntropy:
deep_cfr_tf2.py is exactly the name I used for my local implementation. Google is everywhere.

lanctot:
Yes it is surely be NashConv/2 (= exploitability) as that's been more common in the poker community and among those authors.

StochasticEntropy:
I'll clean up my code and probably include tf.data and post it in a few days

StochasticEntropy:
Anyways, I would like to know if it's right, @EricSteinberger, can you tell us if you used NashConv or NashConv/2 in your Figure 1?

lanctot:
Why close this one? I have imported it already. You can make a separate one for the TF2 impl?

StochasticEntropy:
Sorry, I didn't mean to close anything

EricSteinberger:
Hi! Yes, we use the average exploitability of the players (i.e. NashConv/2).

StochasticEntropy:
Thank you for the quick feedback

PR #429, opened 2020-10-31
Add better support for simultaneous move games.

michalsustr:
This changes ExpectedReturns to use the policy for the requested player.
Because of this change, it propagates to multiple places that are implementing the Policy class.

### DIFF
--- a/open_spiel/algorithms/cfr.cc
+++ b/open_spiel/algorithms/cfr.cc
@@ -68,12 +68,13 @@ CFRAveragePolicy::CFRAveragePolicy(const CFRInfoStateValuesTable& info_states,
                                    std::shared_ptr<Policy> default_policy)
     : info_states_(info_states), default_policy_(default_policy) {}
 
-ActionsAndProbs CFRAveragePolicy::GetStatePolicy(const State& state) const {
+ActionsAndProbs CFRAveragePolicy::GetStatePolicy(
+    const State& state, Player player) const {
   ActionsAndProbs actions_and_probs;
-  auto entry = info_states_.find(state.InformationStateString());
+  auto entry = info_states_.find(state.InformationStateString(player));
   if (entry == info_states_.end()) {
     if (default_policy_) {
-      return default_policy_->GetStatePolicy(state);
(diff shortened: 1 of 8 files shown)


lanctot:
Hi @michalsustr, does this lead to CFR working on simultaneouse move games without the conversion to turn-based? If so, can you add a test to cfr_test that verifies this ?

michalsustr:
That would need to be a separate PR. The https://github.com/deepmind/open_spiel/blob/master/open_spiel/algorithms/cfr.cc#L427 would need to be rewritten to use legal actions of the specified player and we need to make sure you can get a `Child(joint_action)` corresponding to the fixed legal action of the player. I needed an utility like this and made it as part of #431

lanctot:
Ok is that why you are doing this? 

I am trying to figure out why we want to go down this route, rather than just converting the games to turn based when using CFR. 

This will change a lot of core code, and CFR has traditionally been applied to trun based games anyway. What is at the end of this road that is so appealing?

lanctot:
Oh this is needed for your infoset tree CFR, I guess?

michalsustr:
> Oh this is needed for your infoset tree CFR, I guess?

Yes exactly, I put it as a separate PR. 

I think in general this touches a more general point: most of the algorithms are (understandably) written for sequential games and the simultaneous move games are an "afterthought", as in this case. I think it is mostly because dealing with sim-move games is slightly more work, but not too much -- you just have to generally do two for loops (per player and per action) instead of one (per action for one player).

lanctot:
Yeah for me it's a balance between complexity, generality, and expected use. Making a CFR that natively handles simultaneous move games makes it more complicated for all use cases, but probably only <5% of users benefit. The other problem is then you're stuck handling the simultaneous move case all the time, which becomes a real hassle if you're not using it. And, for someone first learning CFR, it's an unnecessary blocker to understanding the main algorithm which is mostly used in turn-based games. 

Anyway, I'm fine with this as long as the more general CFR is the one you plan to add and we keep the basic one simple unless/until we have better justification for making it more complex.

PR #430, opened 2020-10-31
Fix running a matrix game with CFR

janrvdolf:
Loading a matrix game as turn-based game, running it with CFR, then computing exploitability caused a fatal error "UtilitySum unimplemented". This pull request implements the UtilitySum abstract method into MatrixGame class.


janrvdolf:
@googlebot I signed it!

michalsustr:
I might be wrong, but I think `Game::UtilitySum()` should match the `GameType::Utility` in the sense that for a zero-sum game it should return 0, for a constant-sum game the value of the constant, and for the other cases (`kGeneralSum`, `kIdentical`) there is no appropriate value to return, so it should fail. The `ComputeUtilitySum()` does a correct calculation, but might be misleading for the latter cases, as it doesn't fail. Is that correct @lanctot ?

janrvdolf:
Thanks for your comment, I get your point Michal. Then I can remove MatrixGame::ComputeUtilitySum I wrote, have to refactor "matrix_games.cc" so only zero-sum and constant-sum games implements the UtilitySum method and update value next to UpdateSum for zero-sum and constant-sum matrix games in "integration_tests/playthroughs/matrix_*.txt".

lanctot:
Thanks @janrvdolf , nice catch. Yes I agree with @michalsustr, in the case of general sum or identical payoff games it should fail.

Btw AFAICT your method sums over all the entries, but the utility sum refers only to each individual entry.

My suggestion is go further up and add it to NormalFormGame. If the game type is zero-sum then return 0. Otherwise if it is constant sum then generate a new initial state and apply joint action of all 0s to get the utils and return that sum. Otherwise, call SpielFatalError.

Yes, you will need to regenerate the playthroughs.

Also beware that when running CFR on nonzero-sum games you can use NashConv and divide by number of players to get exploitability.

janrvdolf:
@lanctot @michalsustr Thank you for your reaction. I updated the code according to your advice. Additionally, I had an issue running `regenerate_playthroughs.sh` with a path to the python script inside, so the last commit fixes it. 

The last thing I wanted to note is, I don't know if it is an issue, that if you choose during the installation not to install Hanabi, then the `regenerate_playthroughs.sh` fails. I thought it could skip Hanabi instead.

lanctot (approved):
Great, thanks! Importing now.

lanctot:
> The last thing I wanted to note is, I don't know if it is an issue, that if you choose during the installation not to install Hanabi, then the `regenerate_playthroughs.sh` fails. I thought it could skip Hanabi instead.

Oh, I agree. I would say that's is a bug. Can you open a new issue for it?

janrvdolf:
> > The last thing I wanted to note is, I don't know if it is an issue, that if you choose during the installation not to install Hanabi, then the `regenerate_playthroughs.sh` fails. I thought it could skip Hanabi instead.
> 
> Oh, I agree. I would say that's is a bug. Can you open a new issue for it? (There is no obvious solution for it, but maybe we can get to it eventually.)

Yes, I have opened a new issue for this.

PR #431, opened 2020-11-02
Add utility iterators for simultaneous move games.

michalsustr:
Adds class ActionView, which allows to have a more fine grained control over submitted joint actions for simultaneous-move games.


michalsustr:
Hm, maybe I should add some tests to this :)

open_spiel/utils/action_view.h
```diff
@@ -0,0 +1,71 @@
...
+// with simultaneous move nodes.
+struct ActionView {
+  const Player current_player;
+  const std::vector<std::vector<Action>> legal_actions;
+  // Collects legal actions at the specified state.
+  ActionView(const State& state);
+
+  // Provides an iterator over flattened actions where we fix an action
+  // for the specified player.
+  struct FixedActions {
```
elkhrt, line 36: It would be neater to have FixedActions / FixedActionsIterator and FlatJointActions / FlatJointActionIterator
rather than combine the two in the same class (also the iterators should probably be classes rather than structs - we don't want the members to have public visibility)

open_spiel/utils/action_view.h
```diff
@@ -0,0 +1,71 @@
...
+    FixedActions end() const;
+    FixedActions& operator++();
+    Action operator*() const;
+    bool operator==(const FixedActions& rhs) const;
+    bool operator!=(const FixedActions& rhs) const;
+  };
+  FixedActions fixed_action(Player player, int action_index) const;
+
+  // Provides an iterator over flattened actions. This is equivalent to calling
+  // SimMoveState::LegalFlatJointActions() and iterating over the result, but it
```
elkhrt, line 54: Can you rewrite LegalFlatJoinActions in terms of this? 
It would be preferable not to have two implementations of the same logic.

open_spiel/utils/action_view.cc
```diff
@@ -0,0 +1,130 @@
...
+}
+
+ActionView::ActionView(const State& state)
+    : current_player(state.CurrentPlayer()),
+      legal_actions(CollectActions(state)) {}
+
+// FlatJointActions
+
+ActionView::FlatJointActions ActionView::flat_joint_actions() const {
+  int prod = 1;
```
elkhrt, line 44: better name please

open_spiel/utils/action_view.cc
```diff
@@ -0,0 +1,130 @@
...
+  for (const std::vector<Action>& actions : legal_actions) {
+    if (actions.size() > 0) prod *= actions.size();
```
elkhrt, line 46: !actions.empty()

open_spiel/utils/action_view.cc
```diff
@@ -0,0 +1,130 @@
...
+  }
+  return FlatJointActions{prod};
+}
+
+ActionView::FlatJointActions ActionView::FlatJointActions::begin() const {
+  return *this;
+}
+ActionView::FlatJointActions ActionView::FlatJointActions::end() const {
+  return FlatJointActions{prod, prod};
+}
```
elkhrt, line 56: run clang format (or I can do this)

open_spiel/utils/action_view.cc
```diff
@@ -0,0 +1,130 @@
...
+bool ActionView::FlatJointActions::operator==(
+    ActionView::FlatJointActions other) const {
+  return pos == other.pos && prod == other.prod;
+}
+bool ActionView::FlatJointActions::operator!=(
+    ActionView::FlatJointActions other) const {
+  return !(*this == other);
+}
+Action ActionView::FlatJointActions::operator*() const { return pos; }
```
elkhrt, line 69: this should probably have a better name too

open_spiel/utils/action_view.cc
michalsustr: Please do.

michalsustr:
As Ed requested to use the action view in the sim nodes implementation I had to move it to the core, because otherwise it wouldn't link properly.

PR #432, opened 2020-11-02
Add a test for creation of a shared library.

michalsustr:
We can now have a shared library that includes third party dependencies packaged inside of it, so we can use things like sequence-form LPs without researchers worrying about linking with OrTools. 

This is useful for experiments, where you can just provide a singled shared library with all dependencies and copy it for example to a computation cluster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ matrix:
       - BUILD_SHARED_LIB="ON"
       - DEFAULT_OPTIONAL_DEPENDENCY="ON"
       - BUILD_WITH_ORTOOLS="ON"
+      - BUILD_WITH_ORTOOLS_DOWNLOAD_URL="https://github.com/google/or-tools/releases/download/v8.0/or-tools_ubuntu-18.04_v8.0.8283.tar.gz"
 
 script:
   - pwd
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,16 @@ matrix:
       - DEFAULT_OPTIONAL_DEPENDENCY="ON"
       - TRAVIS_USE_NOX=1
 
+  # Build and run tests for building a shared library with linkable
+  # third party dependencies in place.
+  - os: linux
+    dist: bionic # Ubuntu 18.04.2 LTS released on 26 April 2018
+    env:
+      - OS_PYTHON_VERSION=3.6
+      - BUILD_SHARED_LIB="ON"
+      - DEFAULT_OPTIONAL_DEPENDENCY="ON"
+      - BUILD_WITH_ORTOOLS="ON"
+
 script:
   - pwd
   - chmod +x install.sh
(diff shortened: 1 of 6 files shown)


michalsustr:
Not sure why, but the last travis failure was due julia, not the new ortools.

michalsustr:
Ok, probably Travis was just having a bad day, as it sometimes does :)

lanctot (approved):
Neat, thanks!

open_spiel/scripts/install.sh
```diff
@@ -150,6 +150,23 @@ if [[ ${BUILD_WITH_LIBTORCH:-"ON"} == "ON" ]] && [[ ! -d ${DIR} ]]; then
...
+# Add OrTools
+# This downloads the precompiled binaries available from the official website.
+# https://developers.google.com/optimization/install/cpp/
+DIR="open_spiel/ortools"
+if [[ ${BUILD_WITH_ORTOOLS:-"ON"} == "ON" ]] && [[ ! -d ${DIR} ]]; then
+  # You may want to replace this URL according to your system.
+  #
+  # Use version 8 at minimum, due to compatibility between absl library versions
+  # used in OpenSpiel and in OrTools.
+  DOWNLOAD_URL="https://github.com/google/or-tools/releases/download/v8.0/or-tools_ubuntu-18.04_v8.0.8283.tar.gz"
```
lanctot, line 162: Can we have an if Mac OS download X, else download Y here?

.travis.yml
```diff
@@ -60,6 +60,16 @@ matrix:
...
 
+  # Build and run tests for building a shared library with linkable
+  # third party dependencies in place.
+  - os: linux
+    dist: bionic # Ubuntu 18.04.2 LTS released on 26 April 2018
+    env:
+      - OS_PYTHON_VERSION=3.6
+      - BUILD_SHARED_LIB="ON"
+      - DEFAULT_OPTIONAL_DEPENDENCY="ON"
+      - BUILD_WITH_ORTOOLS="ON"
```
lanctot, line 71: Can we add something similar for MacOS? Should only require a few small changes to install.sh.

lanctot (changes requested):

.travis.yml
lanctot: Actually, never mind. I don't think I've tested the shared lib on MacOS yet, but will soon.

open_spiel/scripts/install.sh
lanctot: Oops, I realized I had never sent that last message, sorry...! 

Easy to do on my side... but easier for you to test directly on Travis by just doing more commits, so let's do that. 

Basically, I think all it takes is checking that BUILD_WITH_OR_TOOLS is something other than "OFF" and copy that to DOWNLOAD_URL. And then in travis.yml just use the Ubuntu 18.04 link. 

We should clarify this in the documentation (like the main CMakeLists too, and maybe include all the common links: MacOS, Ubuntu 18.04 and Ubuntu 20.04 so people can copy/paste.

Thanks for doing this.. sorry for the run-around!

michalsustr:
I rather made the download url a separate global variable, as I think it's better to have a consistent implicit type system (bool for `BUILD_WITH_XXX`) in the configs, rather than mixing them up (bool or optional<string>).

lanctot:
Ok great thanks!

lanctot (approved):

Issue #398, opened 2020-10-07
Thoughts on C++ optimization (LP solving)


lanctot:
Hi @andrschl, @michalsustr, @gabrfarina, 

I've recently been working on LP-solving from C++, using [OR-Tools](https://developers.google.com/optimization). I have a simple example that solves a matrix game (not yet released). It works like a charm internally, but I am having trouble externally. I managed to get it to compile and link against the binary distribution but it crashes when I run it, and in so doing I discovered the wonderful world of ABI incompatibilities between C++ standards, so that's not an option. When I try building from source I get a bunch of errors with cmake (see below: OR-Tools's CMake setup is marked as experimental, so this could just be something that is not supported yet).

I will include the code in the next update including my attempt to get it to work with CMake if you're curious. But, this has led me to consider other LP solvers. OR-Tools is very appealing because it checks off all the boxes: works internally and externally, and problems can be solved using some of the large commercial solvers. But it a lot of dependencies, so building from scratch could take some time and long-term could be hard to maintain. I've used GLPK in the past, and it worked quite well, but building the LP code on GLPK would mean it can't transfer seamlessly to e.g. CPLEX or Gurobi.

So I'm a bit torn.. I still want to try to get OR-Tools working, especially because they support a CMake build (!) which is great for us, but what I know for sure is that I can't spend much more time on this at the moment. I'm wondering what you have used in the past to do your LP solving, and was it from C++? Any recommendations on where to go from here? 

Here's the promised CMake error message:

```
CMake Error in /usr/local/google/home/lanctot/tmp/open_spiel/build/_deps/gflags-src/CMakeLists.txt:
  export called with target "gflags_nothreads_static" which requires target
  "open_spiel_core" that is not in any export set.


CMake Error in /usr/local/google/home/lanctot/tmp/open_spiel/build/_deps/gflags-src/CMakeLists.txt:
  export called with target "gflags_nothreads_static" which requires target
  "container" that is not in any export set.


CMake Error in /usr/local/google/home/lanctot/tmp/open_spiel/build/_deps/gflags-src/CMakeLists.txt:
  export called with target "gflags_nothreads_static" which requires target
  "flags" that is not in any export set.


CMake Error in /usr/local/google/home/lanctot/tmp/open_spiel/build/_deps/gflags-src/CMakeLists.txt:
  export called with target "gflags_nothreads_static" which requires target
  "flags_parse" that is not in any export set.


[...]

PR #442, opened 2020-11-17
Add infostate trees.

michalsustr:
This implements infostate trees following discussion of #398 with the extra ability of constructing depth-limited infostate trees.

The tree creation is extensively tested and I have a number of algorithms that work correctly on these structures (CFR, Sequence-form LP). 

Some things which should have (more) tests:

- [ ] Best response computation.
- [ ] Compute "gradients".
- [ ] InfostateTree methods in the sections of "Sequence operations" and "Decision operations"

Tagging @gabrfarina @elkhrt @Dugy as I'd like to kindly ask you to review the code. Thank you very much!

### DIFF
--- a/open_spiel/algorithms/infostate_tree.cc
+++ b/open_spiel/algorithms/infostate_tree.cc
@@ -694,13 +694,12 @@ std::pair<double, SfStrategy> InfostateTree::BestResponse(
     double max_value = init_value;
     SequenceId max_id = current;
     const InfostateNode* node = observation_infostate(current);
-    current = node->start_sequence_id();
-    while (current != node->end_sequence_id()) {
+    for (current = node->start_sequence_id();
+         current != node->end_sequence_id(); current.next()) {
       if (gradient[current] > max_value) {
         max_value = gradient[current];
         max_id = current;
       }
-      current.next();
     }
     if (init_value != max_value) {
       gradient[node->sequence_id()] += max_value;
--- a/open_spiel/algorithms/infostate_tree.cc
+++ b/open_spiel/algorithms/infostate_tree.cc
@@ -499,6 +499,14 @@ std::shared_ptr<InfostateTree> MakeInfostateTree(
 std::shared_ptr<InfostateTree> MakeInfostateTree(
     const std::vector<InfostateNode*>& start_nodes,
     int max_move_ahead_limit) {
+  std::vector<const InfostateNode*> const_nodes(
+      start_nodes.begin(), start_nodes.end());
+  return MakeInfostateTree(const_nodes, max_move_ahead_limit);
+}
+
+std::shared_ptr<InfostateTree> MakeInfostateTree(
+    const std::vector<const InfostateNode*>& start_nodes,
+    int max_move_ahead_limit) {
   SPIEL_CHECK_FALSE(start_nodes.empty());
   const InfostateNode* some_node = start_nodes[0];
   const InfostateTree& originating_tree = some_node->tree();
--- a/open_spiel/algorithms/infostate_tree.cc
+++ b/open_spiel/algorithms/infostate_tree.cc
@@ -1,0 +1,814 @@
+// Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "open_spiel/algorithms/infostate_tree.h"
+
+#include <limits>
+#include <memory>
+#include <string>
+#include <stack>
+#include <utility>
+#include <vector>
+
+#include "open_spiel/action_view.h"
+
+
+namespace open_spiel {
+namespace algorithms {
+
+InfostateNode::InfostateNode(
+    const InfostateTree& tree, InfostateNode* parent, int incoming_index,
+    InfostateNodeType type, const std::string& infostate_string,
+    double terminal_utility,
+    double terminal_ch_reach_prob, size_t depth,
+    std::vector<Action> legal_actions, std::vector<Action> terminal_history)
+    : tree_(tree), parent_(parent),
+      incoming_index_(incoming_index), type_(type),
+      infostate_string_(infostate_string),
+      depth_(depth),
+      terminal_utility_(terminal_utility),
+      terminal_chn_reach_prob_(terminal_ch_reach_prob),
+      legal_actions_(std::move(legal_actions)),
+      terminal_history_(std::move(terminal_history)) {
+
+  // Implications for kTerminalNode
+  SPIEL_DCHECK_TRUE(type_ != kTerminalInfostateNode || parent_);
+  // Implications for kDecisionNode
+  SPIEL_DCHECK_TRUE(type_ != kDecisionInfostateNode || parent_);
+  // Implications for kObservationNode
+  SPIEL_DCHECK_TRUE(
+      !(type_ == kObservationInfostateNode
+            && parent_ && parent_->type() == kDecisionInfostateNode)
+      || (incoming_index_ >= 0
+            && incoming_index_ < parent_->legal_actions().size())
+  );
+}
+
+const std::string& InfostateNode::infostate_string() const {
+  // Avoid working with empty infostate strings.
+  // Use Hasinfostate_string() first to check.
+  SPIEL_DCHECK_TRUE(has_infostate_string());
+  return infostate_string_;
+}
+
+bool InfostateNode::has_infostate_string() const {
+  return infostate_string_ != kFillerInfostate
+      && infostate_string_ != kDummyRootNodeInfostate;
+}
+
+double InfostateNode::terminal_utility() const {
+  SPIEL_CHECK_EQ(type_, kTerminalInfostateNode);
+  return terminal_utility_;
+}
+
+double InfostateNode::terminal_chance_reach_prob() const {
+  SPIEL_CHECK_EQ(type_, kTerminalInfostateNode);
+  return terminal_chn_reach_prob_;
+}
+
+const std::vector<Action>& InfostateNode::legal_actions() const {
+  SPIEL_CHECK_EQ(type_, kDecisionInfostateNode);
+  return legal_actions_;
+}
+
+const std::vector<std::unique_ptr<State>>& InfostateNode::corresponding_states()
+const {
+  return corresponding_states_;
+}
+
+const std::vector<double>& InfostateNode::corresponding_chance_reach_probs()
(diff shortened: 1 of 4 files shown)


open_spiel/algorithms/infostate_tree.cc
```diff
@@ -0,0 +1,814 @@
...
+  if (type_ == kTerminalInfostateNode) return "{}";
+
+  std::vector<std::string> certificates;
+  for (InfostateNode* child : child_iterator()) {
+    certificates.push_back(child->MakeCertificate());
+  }
+  std::sort(certificates.begin(), certificates.end());
+
+  std::string open, close;
+  if (type_ == kDecisionInfostateNode) {
```
Dugy, line 141: These might better be off as virtual function calls.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -0,0 +1,814 @@
...
+
+  if (is_leaf_node() && target_depth != current_depth) {
+    // Prepare the chain of dummy observations.
+    depth_ = target_depth;
+    std::unique_ptr<InfostateNode> node = Release();
+    InfostateNode* node_parent = node->parent();
+    int position_in_leaf_parent = node->incoming_index();
+    std::unique_ptr<InfostateNode> chain_head =
+        std::unique_ptr<InfostateNode>(new InfostateNode(
+            /*tree=*/tree_, /*parent=*/nullptr,
```
Dugy, line 167: Your IDE does not show these?

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -0,0 +1,814 @@
...
+  // 1. Compute counterfactual best response
+  // (i.e. in all infostates, even unreachable ones)
+  SequenceId current(0, this);
+  const double init_value = -std::numeric_limits<double>::infinity();
+  while (current.id() <= empty_sequence().id()) {
+    double max_value = init_value;
+    SequenceId max_id = current;
+    const InfostateNode* node = observation_infostate(current);
+    current = node->start_sequence_id();
+    while (current != node->end_sequence_id()) {
```
Dugy, line 698: This might be more readable as a `for` cycle, randomly adding a `continue` into the cycle would break it.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -0,0 +1,814 @@
...
+  if (tree->IsLeafSequence(id)) {
+    return true;
+  }
+
+  double actual_sum = 0.;
+  const InfostateNode* node = tree->observation_infostate(id);
+  for (SequenceId sub_seq : node->AllSequenceIds()) {
+    actual_sum += strategy[sub_seq];
+  }
+  if (fabs(actual_sum - expected_sum) > 1e-13) {
```
Dugy, line 797: This repeating pattern could be moved into a utility function `if (deviation(first, second) < 1e-13) \\\...`

open_spiel/algorithms/infostate_tree.h
```diff
@@ -0,0 +1,630 @@
...
+  // Build the tree.
+  void RecursivelyBuildTree(InfostateNode* parent, size_t depth,
+                            const State& state, int move_limit,
+                            double chance_reach_prob);
+  void BuildTerminalNode(InfostateNode* parent, size_t depth,
+                         const State& state, double chance_reach_prob);
+  void BuildDecisionNode(InfostateNode* parent, size_t depth,
+                         const State& state, int move_limit,
+                         double chance_reach_prob);
+  void BuildObservationNode(InfostateNode* parent, size_t depth,
```
Dugy, line 388: Maybe these repeating argument groups could be moved into a value struct.

open_spiel/algorithms/infostate_tree.h
```diff
@@ -0,0 +1,630 @@
...
+      std::vector<Action> legal_actions, std::vector<Action> terminal_history);
+  friend class InfostateTree;
+
+ public:
+  InfostateNode(InfostateNode&&) noexcept = default;
+  virtual ~InfostateNode() = default;
+
+  // -- Node accessors. --------------------------------------------------------
+  const InfostateTree& tree() const { return tree_; }
+  InfostateNode* parent() const { return parent_; }
```
Dugy, line 502: These might be moved into a common parent.

open_spiel/algorithms/infostate_tree_test.cc
```diff
@@ -0,0 +1,474 @@
...
+    for (int i = 0; i < 2; ++i) {
+      std::shared_ptr<InfostateTree> tree = MakeTree(iigs2, /*player=*/i);
+      SPIEL_CHECK_EQ(tree->root().MakeCertificate(), expected_certificate);
+    }
+  }
+  {  // Full Kuhn test.
+    std::shared_ptr<InfostateTree> tree = MakeTree("kuhn_poker", /*player=*/0);
+    std::string expected_rebalanced_certificate =
+      // Notice all terminals are at the same depth (same indentation).
+      "((" // Root node, 1st is getting a card
```
Dugy, line 117: How about using macros to concatenate to create these repetitive strings?

open_spiel/algorithms/infostate_tree_test.cc
```diff
@@ -0,0 +1,474 @@
...
+        "[((((({})))))((((({})))))]"
+      ")";
+    std::shared_ptr<InfostateTree> tree = MakeTree(
+        "kuhn_poker", /*player=*/0,
+        /*start_histories=*/{{0}, {2, 1, 0, 1}},
+        /*start_reaches=*/{1/3., 1/6.});
+    SPIEL_CHECK_EQ(tree->root().MakeCertificate(), expected_certificate);
+  }
+  {
+    std::string expected_certificate =
```
Dugy, line 232: This code looks like if could be part of a configuration file.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -82,13 +82,19 @@ const std::vector<Action>& InfostateNode::legal_actions() const {
   return legal_actions_;
 }
 
+size_t InfostateNode::corresponding_states_size() const {
+  return corresponding_states_.size();
```
Dugy, line 86: Because you seem to care about performance, consider moving one-liners into the header so that they could be inlined.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -491,12 +491,57 @@ std::shared_ptr<InfostateTree> MakeInfostateTree(
     const Game& game, Player acting_player,
     int max_move_limit) {
   return std::shared_ptr<InfostateTree>(new InfostateTree(
```
Dugy, line 493: Use `std::make_shared`, it shortens code like this. Also, it would allow you to search for the `new` keyword to check if everything uses smart pointers (skipping the low level parts that might need it).

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -491,12 +491,57 @@ std::shared_ptr<InfostateTree> MakeInfostateTree(
...
+std::shared_ptr<InfostateTree> MakeInfostateTree(
+    const std::vector<InfostateNode*>& start_nodes,
+    int max_move_ahead_limit) {
+  std::vector<const InfostateNode*> const_nodes(
+      start_nodes.begin(), start_nodes.end());
+  return MakeInfostateTree(const_nodes, max_move_ahead_limit);
+}
+
+std::shared_ptr<InfostateTree> MakeInfostateTree(
+    const std::vector<const InfostateNode*>& start_nodes,
```
Dugy, line 508: Are the pointers in the vector managed elsewhere? If yes, then it's OK, otherwise please use some smart pointer type.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -491,12 +491,57 @@ std::shared_ptr<InfostateTree> MakeInfostateTree(
...
+      if (!node) return false;
+      if (!node->is_leaf_node()) return false;
+      if (node->depth() != some_node->depth()) return false;
+      if (&node->tree() != &originating_tree) return false;
+    }
+    return true;
+  }());
+
+  std::vector<const State*> start_states;
+  start_states.reserve(start_nodes.size() * 8);
```
Dugy, line 524: What is 8? Shouldn't it be a `constexpr` value in some class? If it's just a guess how much space will be most likely enough (to avoid reallocations), then it's okay but maybe a comment would help.

open_spiel/algorithms/infostate_tree.cc
```diff
@@ -491,12 +491,57 @@ std::shared_ptr<InfostateTree> MakeInfostateTree(
...
+      start_states, chance_reach_probs,
+      originating_tree.infostate_observer_,
+      originating_tree.acting_player_,
+      max_move_ahead_limit));
 }
 
 std::shared_ptr<InfostateTree> MakeInfostateTree(
     const std::vector<const State*>& start_states,
-    const std::vector<float>& chance_reach_probs,
+    const std::vector<double>& chance_reach_probs,
```
Dugy, line 544: Are you sure you need double precision? A `float` has 4-5 decimals' precision, `double` usually makes sense only when error can accumulate, like summing some values, subtracting similar values etc.

open_spiel/algorithms/infostate_tree_test.cc
```diff
@@ -311,13 +309,60 @@ void CheckTreeLeaves(const InfostateTree& tree, int move_limit) {
...
+        terminal_continuation = terminal_continuation->child_at(0);
+      }
+      SPIEL_CHECK_EQ(terminal_continuation->type(), kTerminalInfostateNode);
+      SPIEL_CHECK_EQ(leaf_node->has_infostate_string(),
+                     terminal_continuation->has_infostate_string());
+      if (leaf_node->has_infostate_string()) {
+        SPIEL_CHECK_EQ(leaf_node->infostate_string(),
+                       terminal_continuation->infostate_string());
+      }
+      SPIEL_CHECK_EQ(leaf_node->terminal_utility(),
```
Dugy, line 348: The pattern on checking if `leaf_node` and `terminal_continuation` return the same value from some method call repeat here too many times. Consider if this isn't a situation where an overload of the equality operator wouldn't be better. In any case, you can make a lambda that would be called only as `checkEquality(&InfostateNode::terminal_utility);`.

open_spiel/algorithms/infostate_tree.cc
michalsustr: Had to do it this way because of friendliness.

open_spiel/algorithms/infostate_tree.cc
michalsustr: Yes they are managed by the another tree where they may come from (when we want to construct tree continuation based on a set of nodes)

open_spiel/algorithms/infostate_tree.cc
michalsustr: I think it's better because of LP construction.

open_spiel/algorithms/infostate_tree_test.cc
michalsustr: operator== is tricky to get right, so I wouldn't do it unless there is a strong use-case for it -- this test isn't, as it's not quite doing exactly ==

open_spiel/algorithms/infostate_tree.cc
Dugy: Aha, that is a good reason to do it the old way.

PR #445, opened 2020-11-23
Implement the Sheriff game

gabrfarina:
This draft pull request tracks progress on the implementation of the Sheriff game. One of the goals of this implementation is to be able to exactly match the game generator that was used to generate game instances for some recent papers on extensive-form correlated equilibria. This particular version of the game was introduced in [here](https://papers.nips.cc/paper/9122-correlation-in-extensive-form-games-saddle-point-formulation-and-benchmarks.pdf).

Things remaining:

- [x] Is it OK for the `ObservationString` to match the `InformationStateString`?
- [x] Compare game sizes against the original generator
- [x] Add game to table of games

r? @lanctot @michalsustr


lanctot:
Hi @gabrfarina, cool!

It's ok for ObservationString to match InformationStateString, but we normally reserve the observation to be something more "immediate" (in the case of imperfect info, just the new information such that a e.g. the sequence of observation could recover the info state). Is there a reason to have ObservationString defined? In this case, you can just leave it unimplemented and mark it as unimplemented in the game type (that would be my preference).

Also, I have to cut down the test time for battleship by making some of the tests optional, as they are taking a very long time (sometimes timing out on Travis, leading to tests failing). My plan is to add a flag to enable all tests, and have it be false by default. Can you do the same for Sheriff? (I guess the tree size comparison sweep is what is dominating the time? So maybe I will just disable those) I will do an update later today so you see what I did specifically.

gabrfarina:
> Is there a reason to have ObservationString defined? In this case, you can just leave it unimplemented and mark it as unimplemented in the game type (that would be my preference).

Ah yes, I'll do that!

> Also, I have to cut down the test time for battleship by making some of the tests optional, as they are taking a very long time 

I think perhaps it would be easiest to change the number of random tests here:

https://github.com/deepmind/open_spiel/blob/f0eda016880b269c36f00d6f595c18750b5e6233/open_spiel/games/battleship_test.cc#L59

Maybe we could go from 100 to something like 20?

gabrfarina:
I've added tests that make sure the game sizes match those of the original generator. I have kept the game sizes smalls to avoid excessive testing times as pointed out by @lanctot .

In the future, it would be a good idea to add tests checking individual payoffs. But given that all tests pass, for now I will mark this PR as ready for review.

lanctot:
Great timing! I'm hoping to issue a release in 2-3 weeks and was hoping Sheriff could make it in there.

I am busy with a few other PRs so it might be a day or so before I can look at this, but really cool to see it as a PR. Thanks @gabrfarina !

open_spiel/games/sheriff.cc
```diff
@@ -0,0 +1,392 @@
...
+  if (IsTerminal()) {
+    return {};
+  } else if (!num_illegal_items_) {
+    // This is the beginning of the game. The smuggles must decide how many
+    // illegal items to place in the cargo, which must be an integer in the
+    // range [0, conf.max_items]. The action id will correspond to the number
+    // of illegal items placed in the cargo.
+
+    action_ids.reserve(conf.max_items + 1);
+    for (uint32_t num_illegal_items = 0; num_illegal_items <= conf.max_items;
```
michalsustr, line 78: I'm fine with `uint32`, it's just I didn't see it used very much in the library, as usually `int`s are used for storing the settings. I guess it's ok according to the [style guide](https://google.github.io/styleguide/cppguide.html), which just specifies to not use `unsigned` qualifier.

open_spiel/games/sheriff.cc
```diff
@@ -0,0 +1,392 @@
...
+std::unique_ptr<State> SheriffState::Clone() const {
+  return std::make_unique<SheriffState>(*this);
+}
+
+std::string SheriffState::InformationStateString(Player player) const {
+  SPIEL_CHECK_TRUE(player >= 0 && player < NumPlayers());
+
+  std::string infostring = absl::StrCat("T=", MoveNumber(), " ");
+  if (player == kSMUGGLER) {
+    infostring = "num_illegal_items:";
```
michalsustr, line 174: The move number is ignored for the smuggler, is that on purpose?

michalsustr (commented):
I made a quick look, looks good to me, but I didn't check details of the implementation (if they match the game rules).

open_spiel/games/sheriff.cc
```diff
@@ -0,0 +1,392 @@
...
+
+SheriffState::SheriffState(
+    const std::shared_ptr<const SheriffGame> sheriff_game)
+    : State(sheriff_game), sheriff_game_(sheriff_game) {}
+
+Player SheriffState::CurrentPlayer() const {
+  if (!num_illegal_items_) {
+    // The smuggler still hasn't decided the number of illegal items to
+    // place in the cargo. The game has just begun.
+    return kSMUGGLER;
```
lanctot, line 47: Can you make the constants follow this format:

kSmuggler
kSheriff

(just not all caps)

open_spiel/games/sheriff.cc
```diff
@@ -0,0 +1,392 @@
...
+    // The action must represent the inspection feedback returned by the
+    // sheriff.
+    SPIEL_CHECK_EQ(CurrentPlayer(), kSHERIFF);
+    inspection_feedback_.push_back(
+        sheriff_game_->DeserializeInspectionFeedback(action_id));
+  }
+}
+
+// Facts about the game
+const GameType kGameType{
```
lanctot, line 243: We normally put these further up closer to the top of the file so people can easily find them, do you need it down here for any reason?

We also usually put this (and the REGISTERS_SPIEL_GAME, I think) in an anonymous namespace.

lanctot (changes requested):

open_spiel/games/sheriff.cc
gabrfarina: That's a good point! Let's see what the maintainers prefer, I don't mind moving to `int` and add checks for non-negativity where needed.

open_spiel/games/sheriff.cc
```diff
@@ -0,0 +1,392 @@
...
+#include "open_spiel/abseil-cpp/absl/strings/str_cat.h"
+
+namespace open_spiel {
+namespace sheriff {
+namespace {
+inline constexpr const Player kSMUGGLER = Player{0};
+inline constexpr const Player kSHERIFF = Player{1};
+
+template <typename T>
+void StrAppendVector(std::string* s, const std::vector<T>& v) {
```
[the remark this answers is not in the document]
gabrfarina: I don't have a strong preference, and it's a tradeoff. Though I would say that using stringstream means allocating unnecessary strings and copying them. At that point, if we accept the string copies I'd rather use something like `absl::Append(s, "[", absl::StrJoin(v, ","), "]")`.

The alternative is to leave the function above as it is in the anonymous namespace and avoid the copy. Not sure what's most ergonomic?

open_spiel/games/sheriff.cc
gabrfarina: Ah thanks, I had missed it. I will move `kGameType` in the anonymous namespace (together with the registration call. 

By the way, I don't mind applying this fix to Battleship as well in parallel later 🙂

open_spiel/games/sheriff.cc
gabrfarina: Ah, good catch! That was not done on purpose 😄

open_spiel/games/sheriff.cc
lanctot: I think uint32 is ok, thanks.

lanctot (approved):

PR #448, opened 2020-11-30
Make changes necessary for pip sdist

lanctot:
- setup.py: Add checks for C++ compiler and python-dev headers
- setup.py: Look for requirements.txt in current and parent directory
- setup.py & noxfile.py: Add flag to determine building and testing everything as opposed to just pyspiel
- Add requirements.txt to MANIFEST.in (read by setup.py)
- Change name to `open_spiel` and version to `0.2.0`
- Use compiler specified by CXX environment variable and drop down to default of clang++

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,3 +1,17 @@
+# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
 # Lint as: python3
 """An integration test building and testing open_spiel wheel."""
 import os
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,17 +1,3 @@
-# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -1,17 +1,3 @@
-# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
```
jblespiau, line 1: Should we keep that?

setup.py
```diff
@@ -1,17 +1,3 @@
-# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
```
jblespiau, line 1: same

setup.py
```diff
@@ -45,36 +31,70 @@ class BuildExt(build_ext):
...
 
   def run(self):
+    self._check_build_environment()
+    for ext in self.extensions:
+      self.build_extension(ext)
+
+  def _check_build_environment(self):
+    """Check for required build tools: CMake, C++ compiler, and python dev."""
     try:
       subprocess.check_call(["cmake", "--version"])
```
jblespiau, line 41: Do we want a minimal version? I remember some people had issues with old Cmakes, so maybe we can check the version is higher that some version you know works (I have not been following which version works or does not, maybe you know more.

Same for Clang

jblespiau (approved):

noxfile.py
lanctot: Whoops, yes!

setup.py
lanctot: Yes!

setup.py
lanctot: I thought of that originally as well, but I want to shy away from it, for a few reasons:

- Checking the versions is compiler-dependent, so to do this we'd have to force clang specifically.
- Everything should still work with any C++17 compiler, so we shouldn't strictly force clang++ (right now you can override it if the CXX environment variable is set), but set clang++ to the default
- I actually don't know what version we really _need_ but I know what works: Travis Ubuntu 18.04 has Clang 7.0.0 installed. So I modified the comment to say specifically what we need (a C++17 compliant compiler, but that we recommend Clang >= 7.0.0).

What do you think?

PR #450, opened 2020-12-01
add nfsp pytorch

Asugawara:
Hi @lanctot !
 I added implements NFSP in PyTorch.

[The experiment](https://github.com/deepmind/open_spiel/blob/10a55ff6d02d733f7d24c4bb1f5618c876e643d6/open_spiel/colabs/research_nfsp_tf_pt.ipynb) using Kuhn poker does not show a difference between PyTorch and TF.
(There is something wrong with Leduc poker experiment ???)

スクリーンショット 2020-12-01 20 48 41
スクリーンショット 2020-12-01 20 48 52
(episodes*10)


lanctot:
Hi @Asugawara, this is really cool!! I hope we can merge it before the release.

Ok, firstly: this could be due to variance or bad luck. I would not conclude that it is inconsistent just yet. Can you do 2-3 similar runs.. do they all look the same?

The curves look similar and seem to have the same trends. This could be due to network initialization. I remember when I was removing sonnet I found that there was a dpecific initialization value that we used in simple_nets and it made a siginificant difference (I was also trying to reproduce NFSP curves in Leduc!), you can see it here: https://github.com/deepmind/open_spiel/blob/d9a9b8f9f1f44143867217fc3f6ff2db71b174b0/open_spiel/python/simple_nets.py#L45. I am not sure what torch uses internally but maybe you can check and/or try this?

Tagging @jhtschultz too. If there's an export on NFSP reproducibility, it's him!! :)

Asugawara:
Thank you for your response!
First, I'm going to run leduc test a few times. (It'll take some time.)

Yes, I have understood simple_nets.Linear using truncated initialize. 
But Torch does not have one.
(ref: https://github.com/pytorch/pytorch/pull/32377 , 
https://github.com/pytorch/pytorch/issues/32293)

If a few times tests show the difference, I try to implement "truncated initialize”. ٩(ˊᗜˋ*)و

jhtschultz:
Nice addition! Yeah I agree that network initialization is the most likely explanation. And as @lanctot pointed out, there can be some variance between runs so it's possible there isn't even an issue. Will definitely take a closer look after you run it a few times.

Asugawara:
Thank you! @jhtschultz @lanctot 
Sorry for the delay in running a few times test! Please confirm the result.

These images which I attached are the results of the experiment.
Pytorch’s result is a little bit higher than TensorFlow.

スクリーンショット 2020-12-04 20 12 25

averaged over multiple results
スクリーンショット 2020-12-04 20 12 34

lanctot:
@Asugawara sounds a bit fishy to me. Did you look into the network initialization, or maybe you can try a custom one? 

Possibly it's DQN: maybe we can try it separately side-by side on a single-agent task like playing against a fixed strategy in a game. We've had issues with getting the illegal moves properly supported in the past.

Asugawara:
>Did you look into the network initialization, or maybe you can try a custom one?

No, I didn't investigate the network initialization. So my next action is investigating it.

>Possibly it's DQN: maybe we can try it separately side-by side on a single-agent task like playing against a fixed strategy in a game. 

OK! I try to run it! Thank you for your advice!

Asugawara:
Hi Dr. @lanctot ! 
I fexed above problem, this problem is due to my mistake.
Sorry!

Diffs
- [adding Sonnet layer (in dqn.py)](https://github.com/deepmind/open_spiel/pull/450/files#diff-50089fb447a0bb5404ffff1c99cbd2f874add5740fc4e95c81371738653c69ffR92)
- [fixing dtypes](https://github.com/deepmind/open_spiel/pull/450/files#diff-347f34063db107952cc5280fe2fbcf280c89b5841580dce7be9710cc8d379ed2R238)

スクリーンショット 2020-12-05 21 24 13

jhtschultz:
That's a great looking graph! Did the network initialization end up being necessary? Or was the gap entirely because of the Tensor vs. LongTensor?

lanctot:
Excellent!!!!!! Just in time for the release going out early this coming week. :)

Like @jhtschultz I am also curious what the mistake was? The network initialization thing was really not obvious to me when I was tryinf the same thing in June!

Asugawara:
Dr. @jhtschultz Dr. @lanctot 
Thank you for your resposes!

>Did the network initialization end up being necessary? Or was the gap entirely because of the Tensor vs. LongTensor?

Because of the LongTensor.
the below graph shows that the network initialization had little effect on the gap!

スクリーンショット 2020-12-05 22 55 59

>Just in time for the release going out early this coming week. :)
Yes!

lanctot:
Very nice. Good work!

However, what is a LonTensor and why would it affect the convergence? Also does this mean we should also change pytorch DQN and RCFR too?

jhtschultz:
LongTensors hold torch.int64 dtypes. Before the fix a LongTensor was being used to hold action probs which needed floats. PyTorch RCFR doesn't use any LongFloats, and DQN only uses one for actions, which are ints so should be fine.

Asugawara:
I think rounded probability affect the convergence. 

```
import numpy as np
import torch
action_prob = np.random.rand(3) 

torch.Tensor(action_prob)
>>> tensor([0.2497, 0.1564, 0.5688])

torch.LongTensor(action_prob)
>>>  tensor([0, 0, 0])
```

> Also does this mean we should also change pytorch DQN and RCFR too?

we shouldn't change them. They don't use LongTensor to probability!

Asugawara:
> PyTorch RCFR doesn't use any LongFloats, and DQN only uses one for actions, which are ints so should be fine.
Yes! Thank you for confirming!

jhtschultz:
One more thing...from those graphs I don't believe we can conclude that the initialization didn't have much of an effect. We know the LongTensor bug fix was necessary to close the performance gap, but we don't know that it was sufficient on its own. My guess is that the bug fix was sufficient, but to be sure we'd need one more experiment that implemented the bug fix and used the original networks. If you're up for running it I think it's worth it so we can be 100% certain.

lanctot:
BTW the current tests are failing because of a dependency problem, this PR should fix it: https://github.com/deepmind/open_spiel/pull/460. I'll try to get master fixed later tonight.

Asugawara:
>we'd need one more experiment that implemented the bug fix and used the original networks

I agreed with you.
And I run an additional test with bug_fix only. 
スクリーンショット 2020-12-06 19 00 38

I think we need both of them for reproducibility!

lanctot:
Oh no, I imported and merged here: https://github.com/deepmind/open_spiel/commit/ffe263b9b92d248f4f8f23855953e02167311a61 but I think it was missing the latest commits, sorry!

lanctot:
I will try to roll it back and re-import. @Asugawara, is all the code pushed to this PR now?

Asugawara:
>is all the code pushed to this PR now?
Yes!

>I will try to roll it back and re-import
Thank you and sorry for my rebasing. but no code change!

lanctot:
Wait are you saying that I don't need to roll back? Can you check the commit to master that I linked above to see if it has everything? Please confirm. If I missed something, I will roll back and re-import.

Asugawara:
OK, I see.
You won't rollback!

the commit to master that you linked havs 4 commits. (This PR has 4 commits!!)

lanctot:
Ok great, sorry for the confusion. Closing this PR now then. Thanks for checking all the cases!

Asugawara:
I couldn't have done it without you!! Thank you!

jhtschultz:
Thanks for running that last experiment! Looks like the bug fix improves performance by itself, and the initialization might give it an added boost. Can't say definitively from a small sample. Either way I agree it's good to include the initialization to be consistent with the tensorflow version. Thanks again and nice addition @Asugawara!

PR #451, opened 2020-12-01
Double Neural CFR (in progress)

StochasticEntropy:
This is an implementation of Double Neural CFR https://arxiv.org/abs/1812.10607

It is still a work in progress, the regret network is working fine, if I use the Policy Network methods from DeepCFR, i.e., training the policy network in the end, this gives NashConv < 0.3 after 100 iterations (like for DeepCFR, even slightly better, which I suspect is because of using CFR+), however the Double Neural Policy Network only achieves 0.4 if trained consecutively. There's not much information in the paper how they "helped" it to learn a more stable function like they did for the regret network where they scaled it with the known upper bound sqrt(t). But likely there's just a bug in the code, so any input is welcome.

But ultimately, I actually like the idea of using the Regret Network form the Double Neural paper since the regret network has to be trained on each iteration anyways, but then use the policy learning ideas from DeepCFR or SingleDeepCFR for the policy  since we don't need the intermediate results and in both cases everything can just be saved to disk during training.

Also, currently the algorithm uses external sampling, which likely is optimal for simple games like leduc anyways. Are there more interesting games, i.e., games with more structure that a NN could benefit more from in the OpenSpiel repository? (While still being somewhat tractable for playing around, like FHP and TEH)? 

And it's currently just using a simple MLP, from the figures in the papers it doesn't make much difference. If there's a way to get consecutive observations easily from the OpenSpiel API I could also add an RNN.

- [ ] Improve Policy Network Convergence
- [ ] add dn_cfr_test
- [ ] add robust sampling
- [ ] Code cleanup

### DIFF
--- a/open_spiel/python/algorithms/dn_cfr.py
+++ b/open_spiel/python/algorithms/dn_cfr.py
@@ -166,13 +166,14 @@ class PolicyNetwork(tf.keras.Model):
       x = layer(x)
       x = self.activation(x)
 
-    x = self.normalization(x)
+    #x = self.normalization(x)
     x = self.lastlayer(x)
     x = self.activation(x)
     x = self.out_layer(x)
     #x = tf.where(mask == 1, x, -10e20)
-    x = tf.keras.layers.ReLU()(x)
+    #x = tf.keras.layers.ReLU()(x)
     x = mask * x
+
     return x
 
 
@@ -221,12 +222,12 @@ class AdvantageNetwork(tf.keras.Model):
       x = layer(x)
       x = self.activation(x)
 
-    x = self.normalization(x)
+    #x = self.normalization(x)
     x = self.lastlayer(x)
     x = self.activation(x)
     x = self.out_layer(x)
+    #x = tf.keras.layers.ReLU()(x)
     x = mask * x
-
     return x
 
 
@@ -322,7 +323,10 @@ class DeepCFRSolver(policy.Policy):
     if self._save_strategy_memories:
       if os.path.isdir(self._save_strategy_memories):
         self._memories_tfrecordpath = os.path.join(
-                    self._save_strategy_memories, "strategy_memories.tfrecord")
+            self._save_strategy_memories, "strategy_memories.tfrecord")
+        self._strategy_reach_tfrecordpath = [os.path.join(
+            self._save_strategy_memories, f"strategy_reach_memories_{p}.tfrecord")
+            for p in range(self._num_players)]
       else:
         os.makedirs(os.path.split(self._save_strategy_memories)[0],
                                     exist_ok=True)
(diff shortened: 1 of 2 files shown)


lanctot:
This is one I have wanted for a while, very nice!!!

I would really like to get the opinion from the original authors. I know they have used OpenSpiel in the past, I will ask them to take a look.

Thanks!

lanctot:
Tagging DNCFR author @lihuiknight

StochasticEntropy:
I changed the policy weighing in the code to be scaled by t, which makes a huge difference in DeepCFR (in the first 100 iterations, in the beginning a lot is random obviously), for comparism, if I don't use that sample_weight in DeepCFR I'm going from 0.3 to 0.4

StochasticEntropy:
Probably if I was working for google and run it overnight for a trillion iterations I wouldn't see that ;P

lihuiknight:
@StochasticEntropy please refer to the latest version: https://openreview.net/pdf?id=ByedzkrKvH

do you use min-batch MCCFR? it will reduce the variance and will be very helpful.

StochasticEntropy:
I'm using a mini-batch of 1000

lihuiknight:
According to my experience,  the cumulative regret network is very important, please make more attention on this network's loss convergence.

StochasticEntropy:
What output activation do you use for your policy network?

StochasticEntropy:
Thanks :)

lihuiknight:
@StochasticEntropy  a better regret network will make a better final policy. 

For the activation of policy network, I use MSE in my original implementation, see the discussion in section 3.3.2. 

Also, I think you can use softmax if having a large buffer.

StochasticEntropy:
How can we learn a cumulative function if we use softmax?

lihuiknight:
@StochasticEntropy  For example, if there is a large buffer or reservoir to maintain the sampled information, you can use softmax to learn the final policy. Another intuitive idea is you can use the buffer or reservoir to maintain the learned regret networks.

StochasticEntropy:
Yes, that's what I'm currently doing, and then it works, but I'm trying to reproduce your paper results where you train the policy network after each iteration

StochasticEntropy:
I'm currently just doing a

input*(X Relu)**3 X*nbr_output_actions 

for the policy network

StochasticEntropy:
I know @lanctot you couldn't answer the policy problem 11 years ago, but there's still hope for you. Let's go, let's figure out a new policy training.

lihuiknight:
@StochasticEntropy OK, i see.  Typically, tuning the neural network is very important, i understand it needs a lot of efforts to make all the thing well. A reasonable roadmap is you can use a large buffer to save all the infosets' states, and then find a good hyperparameters and neural network to exactly keep track of the tabular CFR.  make sure you can obtain a very low exploitability. This procedure is very important. Then you use mini-batch MCCFR and use the similar hyperparameters to tune the regret network and policy network.

StochasticEntropy:
I'm a bit confused, isn't the idea of your paper that we don't need tabular CFR and buffers?

lihuiknight:
the above roadmap will help you select a better neural net and the corresponding hyperparamters in the Monte Carlo version.

StochasticEntropy:
Thanks.

StochasticEntropy:
So I think this will just stay a half double deep CFR. Best of both worlds :) ... until we understand how to train the policy ...

lihuiknight:
Yes. i see. Obtaining a low enough exploitability on large-scale IIG like tabular CFR via deep learning is very difficult task.  it needs a lot of effort in this direction. @lanctot @StochasticEntropy

StochasticEntropy:
You yourself wrote in your paper that the strategy training is theoretically a bit edgy (as is the regret training...), so do you think your Policy Training Method is better or worse than just saving everything to the end?

StochasticEntropy:
I think we somehow talked about different things. @lihuiknight thanks for trying to help.

ipa23:
@StochasticEntropy I see this PR was closed without merging.  What's the final status of your branch?  Does it work?

PR #456, opened 2020-12-04
Fix heap overflow in deep sea

gabrfarina:
Address sanitizer was reporting a heap overflow in deep sea. After a bit of digging I think I understand the bug, and I think this patch should be enough to fix it. However, I am not familiar with the game, so it would be better for someone to take a second look and confirm that no further action is required to fix this.

### DIFF
--- a/open_spiel/games/deep_sea.cc
+++ b/open_spiel/games/deep_sea.cc
@@ -128,7 +128,9 @@ std::string DeepSeaState::ObservationString(Player player) const {
   SPIEL_CHECK_GE(player, 0);
   SPIEL_CHECK_LT(player, num_players_);
 
-  std::string str(size_ * size_, '.');
+  // We need to account for the possibility that `player_row == size_` at
+  // terminal states, so that's why we add the +1.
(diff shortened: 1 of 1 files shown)


lanctot:
This is very surprising to me, I thought we were running nightly tests with sanitizers which should catch it. I will look into it.

lanctot:
@makroiss can you take a look and confirm that we need to allocate more space for these observations when called at terminal states?

lanctot:
Ok @makroiss answered me internally to say that the change makes sense. Does the playthrough test failing as a result make sense to you @gabrfarina ? (I only checked briefly on my phone, couldn't see the formatting too well.) I suppose this changes the observations at the terminals but that should be it? Possibly we just need to regenerated the playthroughs?

gabrfarina:
Ah yes, that makes sense! I was only running the test for the game waiting for it to be green under ASAN and forgot to check the other tests.

I will regenerate the playthrough.

lanctot:
> I will regenerate the playthrough.

It's ok, easy for me to do on my end during the import.

PR #458, opened 2020-12-04
Baseline in outcome mccfr

colibri17:
I modified the Python version of the outcome montecarlo counterfact regression (`outcome_sampling_mccfr.py`) adding the baseline variable. The code mirrors the C++ version.


lanctot:
Thanks @colibri17 ! Seems like the tests are failing.. did you use the same parameters as the C++ version? Can you think of any explanation for the different exploitability?

colibri17:
Hey there! As it regards the algorithms, yes the parameters are the same (i.e. epsilon exploration factor = 0.6) and the file should be the exact transpose of the C++ version. With respect to tests instead no, I did not change the Python tests values.

The tests seem to fail because values does not fall within given range (once above and once below). So in one case the algorithm performs better and in the other one performs worse. My suspect is the thresholds contained in the test file were chosen basing on the convergence rates we had in the previous version (that one built from your dissertation). Honestly, I do not know if we should expect the exact convergence rates with this new version (I'd say no, but I am not really an expert). If this is not the case, maybe the test thresholds should be updated, and copying the exact values from `outcome_sampling_mccfr_test.cc` might be a good proxy. 
Even in so doing some small adjustments to the thresholds might be required, because I guess that setting `SEED = 230398247` in Python does not give the same results we have in the C++ version (and so the games played during the test will be different).

However, before doing any of this, I think we should be sure that I did not add any bug in the code :) I am going to check again what I've done!

lanctot:
Baseline of zero is mathematically equivalent to outcome sampling, but indeed the numbers could be different due to numerical error (there will be more rounding errors in the baseline version), or as you point out due to the RNG being implemented differently in Python.

I am fine as long as they are roughly in the ballpark as what we should expect, but still want to be sure that it is correct.

Btw IIRC we are using small initial regrets rather than zero when we creating the infoset for the first time, because it has always helped in the sampling variants. Maybe you didn't replicate that on the python side?

I can take a closer look later during the week, but let me know if you think the numbers are roughly in range with outcome sampling on the c++ side (which also uses the baselines method) by averaging the exploitability at the end of a few longer runs (or even better looking at a few runs of each on a graph).

colibri17:
Yes, that's a good point, I had not replicated those on the python side. I will set them and try to run the tests again. Also, I will compare the histograms of some runs obtained from the python and c++ versions. They should have roughly the same distribution. It should not be too difficult to generate the plots. Will keep you updated, thanks for the feedback!

colibri17:
Hi @lanctot! I checked again the code but did non find any bug. Also, I generated some runs both for the C++ and Python version (where I used the same smaller initial regrets of `1 / 1e6`): 
- I did 1000 matches for kuhn_poker measuring Nash explotability after 10000 iterations. For the C++ code I used a seed of 42 and did not specified an initial random seed for Python (resulting in automatically assigning one from numpy). Here the results:
histogram_comparison_kuhn_poker

- I did 100 matches for leduc_poker measuring Nash exploitability after 10000 iterations. For the C++ code I used a seed of 42 and did not specified an initial random seed for Python. Here the results:
histogram_comparison_leduc_poker


I also did some trials with original initial regrets (`1 / 1e3`) but did not notice remarkable differents in the plots.


Surprisingly it seems the results obtained from Python are consistently and slightly better than those ones obtained from C++. This is pretty strange to me. I expected than on average the two versions had roughly the same distribution. Also I doubt that if we increased the number of matches different results are obtained, since both kuhn and leduc poker are quite small games.
The results are not really much different indeed, but I expected more similarity honestly. Is this difference just due to the different implementation of RNG / other randomness (e.g. different ways of sampling in chance nodes among Python and C++)? Or maybe 10000 iterations are not enough to have reliable results?

I also attached the scripts I used to generate and plot the data, so you can double check. 
[scripts.zip](https://github.com/deepmind/open_spiel/files/5660023/scripts.zip)

lanctot:
Hi @colibri17 , thanks for the detailed analysis! Can you clarify what the x and y axes are in your graphs?

colibri17:
On the x axis I set the exploitability measured at the 10000th iteration for the 1000 (or 100) matches, while on the y axis is reported the corresponding frequency. 

On the C++ side I set the seed at the very beginning of the script and then run 1000 (or 100) matches in a for loop. I guess the program keeps picking next values from the random sequence initialized with that seed.

lanctot:
Ok thanks. Hmm, yes that does look larger than I'd expect too. I will look into it, thanks.

lanctot:
Hi @colibri17, my first thought is that there is more numerical error happening in the baseline version (in C++) due to rounding from more divisions and multiplies. Is it easy for you to get a similar looking histogram for your new baseline version in Python? Does it look similar to the C++ one?

colibri17:
Well indeed the charts are both made with the constant baseline version both for Python and C++. If it can help I can run the histograms with the "old" version of Python (so the version of your dissertation), and try to compare this Python version with the new one. Also, it might be useful to compare all these three together (python-no-baseline, python-with-zero-baseline and c++-with-zero-baseline). Running the histograms is easy, so it should be quick.

lanctot:
Oh when you said the python version, you were talking about your new Python version in this PR (with constant baseline of 0)?  Sorry, I misunderstood-- I thought you were referring to the old Python version.

So now that's even more weird, right? I would expect these to look almost identical, especially since it's identical to the C++ one. Outcome sampling does have high variance, though. 

Yes it would be valiable to see what the old one gives because they should be functionally equivalent, but then my theory of the rounding causing the bug has to be wrong, so either the implementations are slightly different or there's something different happening between Python and C++.

lanctot:
> Also, it might be useful to compare all these three together (python-no-baseline, python-with-zero-baseline and c++-with-zero-baseline). Running the histograms is easy, so it should be quick.

It's up to you, but I think what you have already shown is enough for me to check a few places things might be different. It might be a few days before I respond, though.

colibri17:
Nope I was referring to the new one. Sorry my bad, I did not pointed out that explicitly! Yeah, exactly, I expected them to be almost identical as well. 

I made the same two charts by comparing all the 3 versions (c++-with-zero-baseline, python-with-zero-baseline, python-without-baseline):
histogram_comparison_kuhn_poker
histogram_comparison_leduc_poker
Hope it can help.

No worries, unfortunately I will be quite busy at work as well, but please let me know if you need something else.

lanctot:
Hi @colibri17 , ok I think I have the most likely explanation: I was inconsistent about my definition of "iteration" across languages. Oops :)

If you look here: https://github.com/deepmind/open_spiel/blob/b298e9f8b1e36fc3adbc36902181d565ce01f7d4/open_spiel/algorithms/outcome_sampling_mccfr.cc#L68

One iteration on the C++ side only updates for a single player (alternating the update player between iterations), whereas the Python version (and your new python version) does a for loop over the players. Can you verify this by changing that one function above so the C++ version also does a loop over the players, and rerunning the C++ graph?

I think the for loop over the players is more consistent with what I've always called an "iteration" (and I think with what I did with external sampling), though it is not quite consistent with what the RL researchers would call an "episode". Still, I think it's probably best to stick with what's consistent with the rest of the code and literature, so I'd say we should just change the C++ code to instead do the for loop. If people want to compare across sampling techniques, probably best to use raw time or nodes touched anyway.

lanctot:
Also could you increase the bounds for the tests (on both sides) based on your graphs? Choose the maximum value over your runs, and also feel free to add a comment citing this thread for explanation for where the values come from.

colibri17:
Hi @lanctot ! Thanks for your feedback. I did not spot that difference on the `RunIteration` method, I should have noticed while I was checking the code. I just focused on the `SampleEpisode` method. Good catch!

I am quite busy at work right now, but sure I'll make the changes as soon as I have some time. Should I run the graphs and take the maximum also for "liars dice" (which is in the tests of the C++ code but not on the Python side) and "Kuhn Poker with 3 players" (which is in the tests of Python but not on the C++ side)?

No worries. For me is a great way to learn and keep updated. I am glad I can help in improving the repository!

lanctot:
No rush! 

The exploitability for liar's dice is too slow on the Python side, that's why it is left out. That would be great of you to do for the C++ side for Liar's Dice, definitely a welcome improvement, but if you don't have time and the test passes now maybe we can just add a comment that different seeds might produce higher values. I'll leave it up to you!

colibri17:
Ok, I updated the C++ code by introducing a run of the method `SampleEpisode` for each player during one iteration. Thanks to this improvement, now the graphs produced for Kuhn Poker and Leduc Poker indeed are really similar.

I also changed the test thresholds taking the maximum exploitability I got from the multiple runs. This process was performed for games Kuhn Poker, Leduc Poker and Liars Dice. Here the maximum values I took from the multiple runs:
Maximum value cpp for leduc_poker: 2.756697
Maximum value python for leduc_poker: 3.06021554
Test threshold chosen for tests in both Python and C++ code: 3.07

Maximum value cpp for kuhn_poker 2 players: 0.127352
Maximum value python for kuhn_poker 2 players: 0.16043136
Test threshold chosen for tests in both Python and C++ code: 0.17

Maximum value python for kuhn_poker 3 players: 0.21390615
Test threshold chosen for tests in Python code: 0.22

Maximum value cpp for liars_dice: 1.449834
Test threshold chosen for tests in C++ code: 1.45

Finally, I show here the plots for all the multiple runs related to the 4 game types we have:

histogram_comparison_kuhn_poker
histogram_comparison_kuhn_poker_3p
histogram_comparison_leduc_poker
histogram_comparison_liars_dice

lanctot:
@colibri17  you are a star!!!! This is great work, thanks for putting in so much time.

I will try to import it this week but even if I do I think the next update will only be early January.

I encourage you to share some of the story or sumnarize it with links on the original issue thread as we will probably close it after this PR gets merged.

Thanks again and happy holidays!

colibri17:
Oops, I had left in the `CmakeList.txt` a reference to the code I used for tests. Now it works.
You're welcome :) Hope to have time to solve some other issues soon!

Yeah sure good point I will update the issue as well. 

Happy holidays to you too!

PR #468, opened 2020-12-18
[Fix typo] docs/developer_guide.md

yukarinoki:
Two corrections in  docs/developer_guide.md

```
docs/developer_guide.md#L17
*   `open_spiel/games`: Contains the games ++ implementations.
```
This should be C++, not ++.

```
docs/developer_guide.md#L27
## CPP and Python implementations.
```
This might be C++, not CPP


lanctot:
Thanks! Most of the team are on holidays now so we will import and merge in early January.

PR #496, opened 2021-02-10
UCI bot for playing chess

vofak:
Stockfish is a strong chess engine.
This pull requests creates an optional dependency on the stockfish cloned from the official stockfish github repository and implements a stockfish bot playing chess.


lanctot:
Nice!!!! Thanks, this is exciting :)

Seems like the tests failed.. it can't find the source files. But the install seemed to run github, maybe it's an off-by-one directory structure problem? Can you try ./install.sh from scratch on your end -- does it put it into the right place?

open_spiel/bots/stockfish/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+        stockfish/src/nnue/layers/affine_transform.h
+        stockfish/src/nnue/layers/clipped_relu.h
+        stockfish/src/nnue/layers/input_slice.h
+        stockfish_bot.cc
+        stockfish_bot.h
+        )
+
+target_include_directories (stockfish PUBLIC stockfish/src)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -DNNUE_EMBEDDING_OFF")
```
michalsustr, line 62: These seem like important compilation flags. Can you put this on top of the file? 
Is there a document which describes compilation flags that can be passed to Stockfish? Please add a link to them here, and possibly list the most important ones directly in the comment, with a short explanation. Thanks!

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "open_spiel/spiel_utils.h"
+
+#include "open_spiel/games/chess.h"
+
+namespace open_spiel {
+namespace stockfish {
+namespace {
+
+// FEN string of the initial position, normal chess
+const char* StartFEN = "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1";
```
michalsustr, line 34: I believe this is defined somewhere in chess board as well. Use the constant from there.

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+
+
+}  // namespace
+
+StockfishBot::StockfishBot(int argc, char **argv, int move_time) :
```
michalsustr, line 39: Can you make a constructor which passes Options? I think that is what is parsed with CommandLine::init

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+}
+
+void StockfishBot::RestartAt(const State& state) {
+  Restart();
+}
+
+
+Action StockfishBot::Step(const State& state) {
+
+  auto chess_state = dynamic_cast<const chess::ChessState *>(&state);
```
michalsustr, line 71: Use open_spiel::down_cast

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  Threads.start_thinking(pos, sf_states_, limits, false);
+
+  Threads.main()->wait_for_search_finished();
+
+  Thread* bestThread = Threads.get_best_thread();
+
+  // Send again PV info if we have a new best thread
+  if (bestThread != Threads.main())
+    sync_cout << UCI::pv(bestThread->rootPos, bestThread->completedDepth, -VALUE_INFINITE, VALUE_INFINITE) << sync_endl;
```
michalsustr, line 90: What is sync_out? Is it just synchronized stdout? Not sure if we need this debug info printed.
Maybe there is some verbose flag in the options?

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  sync_cout << "bestmove " << UCI::move(bestThread->rootMoves[0].pv[0], false);
+
+  std::cout << sync_endl;
+
+  Move best_move = bestThread->rootMoves[0].pv[0];
```
michalsustr, line 96: sure it's the best move? they are somehow sorted?

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  chess::Move move = MoveFromStockfishMove(best_move, chess_state->Board());
+
+  return chess::MoveToAction(move);
+}
+
+chess::Move StockfishBot::MoveFromStockfishMove(Move move, const chess::StandardChessBoard &board) {
```
michalsustr, line 103: We will need some better namespace control... it's hard to distinguish openspiel chess move from stockfish Move. I guess they didn't put their code into a namespace... Maybe you can put all of their header files into a namespace ::stockfish, something like this (I didn't try this):
```
namespace stockfish {
#include "open_spiel/bots/stockfish/stockfish/src/bitboard.h"
#include "open_spiel/bots/stockfish/stockfish/src/endgame.h"
#include "open_spiel/bots/stockfish/stockfish/src/position.h"
#include "open_spiel/bots/stockfish/stockfish/src/search.h"
#include "open_spiel/bots/stockfish/stockfish/src/thread.h"
#include "open_spiel/bots/stockfish/stockfish/src/tt.h"
#include "open_spiel/bots/stockfish/stockfish/src/uci.h"
#include "open_spiel/bots/stockfish/stockfish/src/syzygy/tbprobe.h"
#include "open_spiel/bots/stockfish/stockfish/src/types.h"
}
```
 and rename this namespace `open_spiel::stockfish` to `open_spiel::bots`. Then we will have nice chess::Move and stockfish::Move names, and stockfish:: will always refer to the original implementation.

open_spiel/bots/stockfish/stockfish_bot.cc
```diff
@@ -0,0 +1,137 @@
...
+      SpielFatalError("Conversion of move from stockfish to openspiel failed");
+    }
+  }
+
+  return chess::Move(*from_sq, *to_sq, board.at(*from_sq), *promotion_piece_type, type_of(move) == CASTLING);
+}
+
+void StockfishBot::InformAction(const State& state, Player player_id,
+                                Action action) {
+
```
michalsustr, line 130: Not implemented?

open_spiel/bots/stockfish/stockfish_bot.h
```diff
@@ -0,0 +1,57 @@
...
+void init();
+}
+
+namespace open_spiel {
+namespace stockfish {
+
+class StockfishBot : public Bot {
+ public:
+  explicit StockfishBot(int argc, char **argv, int move_time);
+  virtual ~StockfishBot() { Threads.set(0); };
```
michalsustr, line 38: Uhm. Not sure if this is good. Why it's here and are you sure it doesn't throw exceptions?

open_spiel/bots/stockfish/stockfish_bot_test.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "open_spiel/spiel.h"
+#include "open_spiel/spiel_bots.h"
+
+namespace open_spiel {
+namespace {
+
+uint_fast32_t Seed() { return absl::ToUnixMicros(absl::Now()); }
+
+void TestStockfish(Bot *opponent_bot, int argc, char **argv) {
+  std::mt19937 rng(Seed());
```
michalsustr, line 33: I think rng does this kind of seeding on it's own internally.

open_spiel/bots/stockfish/stockfish_bot_test.cc
```diff
@@ -0,0 +1,65 @@
...
+void TestStockfishAgainstStockfish(int argc, char **argv) {
+  auto opponent_bot = stockfish::MakeStockfishBot(argc, argv, 30);
+  TestStockfish(opponent_bot.get(), argc, argv);
+}
+
+}  // namespace
+}  // namespace open_spiel
+
+int main(int argc, char **argv) {
+  open_spiel::TestStockfishAgainstUniformRandom(argc, argv);
```
michalsustr, line 63: Don't pass argc/argv as these tests are not cmd-line parametrized. Use empty options instead.

open_spiel/bots/stockfish/stockfish_bot_test.cc
```diff
@@ -0,0 +1,65 @@
...
+
+  bots.push_back(stockfish::MakeStockfishBot(argc, argv, 30));
+  bot_ptrs.push_back(bots.back().get());
+
+  bot_ptrs.push_back(opponent_bot);
+
+  std::unique_ptr<open_spiel::State> state = game->NewInitialState();
+  auto returns = EvaluateBots(state.get(), bot_ptrs,
+                              absl::Uniform<int>(rng, 0, std::numeric_limits<int>::max()));
+  std::cout << returns << std::endl;
```
michalsustr, line 46: You can print out the state history for fun to see what actions were played :) state->History()

open_spiel/bots/stockfish/stockfish_bot.cc
michalsustr: It's hard for me to review this code without the namespaces as I don't know what is in OS and in SF

michalsustr (commented):
Awesome! :)

open_spiel/bots/stockfish/.gitignore
```diff
@@ -0,0 +1,2 @@
+# This directory is fetched during install, do not include in OpenSpiel.
```
jblespiau, line 1: I know it may seem a detail, but because of the way we sync between Google and Github, it's easier for us if we regroup all ignore into https://github.com/deepmind/open_spiel/blob/master/.gitignore

As you can see, there is a " # External git modules" section.

Could you please add any .gitignore into this file?

open_spiel/bots/stockfish/stockfish_bot.cc
jblespiau: (It will likely not lead to anything, but just in case, I asked in https://github.com/official-stockfish/Stockfish/issues/3350 whether stockfish could be into a namespace.)

ddobbelaere:
I don't want to spoil the party but...

Just saying that Stockfish is licensed under GPLv3. I see this project has Apache License v2.0, and according to http://www.apache.org/licenses/GPL-compatibility.html it's a one-way street.

lanctot:
> I don't want to spoil the party but...
> 
> Just saying that Stockfish is licensed under GPLv3. I see this project has Apache License v2.0, and according to http://www.apache.org/licenses/GPL-compatibility.html it's a one-way street.

But we're not copying any of the source into our repos: it's an optional external dependency, that is downloaded during the install script. That should be fine, right?

lanctot:
@ddobbelaere I just read the site you pointed us to. To the best of my understanding, there's nothing in there that would prevent the kind of setup this PR implements. Can you elaborate on why this would not be possible?

michalsustr:
Hmm I think @ddobbelaere is right:

> merely linking to [GPLv3 software] is considered by the GPLv3 authors to create a derivative work

We do _link_ with stockfish in this PR, as we include some header files in the bot... but I guess if we use STDIO to communicate and not _link_ against stockfish that is fine? The build can be contained in OpenSpiel which will make the particular CLI binary with which we can send messages and spawn it as a subprocess... Basically it's the same as if someone downloaded that binary. We use their API, but not their ABI :)

ddobbelaere:
Actually after reading more about the topic, I'm not sure anymore :)

Note that I don't want to play the d*ck here, the solution in the PR is the most elegant way to interface with Stockfish IMHO.
There's something called "compatible licenses", see https://www.gnu.org/licenses/gpl-faq.html#WhatDoesCompatMean and https://en.wikipedia.org/wiki/GNU_General_Public_License#Compatibility_and_multi-licensing

Please use Stockfish! :)

lanctot:
@michalsustr I would guess that means linking in the compiler sense, not "hyperlink" linking in a comment or #include of source files. If you are right, then strictly speaking we have to also remove the mention of SpielViz in our docs. That seems a bit too extreme to me (even for GPLv3). 

@ddobbelaere, no worries, we want to be sure we're interpreting the proper uses before importing this. We surely won't do anything against what is allowed by the licenses.

I will contact some people and get back to you both, thanks for raising this.

lanctot:
The more I think about it, that can't possibly be right because then no permissive-licensed code would be able to merely depend on GPLv3 code. That doesn't make any sense to me. Doesn't mean I'm not wrong, though.. :)

tewalds:
One way that would almost certainly pass legal scrutiny is to run stockfish in a separate process that exports a common protocol. There might be a chess specific protocol already, but if not, the Go Text Protocol (GTP) is pretty simple to implement.

https://github.com/deepmind/open_spiel/blob/master/open_spiel/examples/gtp.cc is an example that wraps the OpenSpiel C++ MCTS agent.

On the other side, there's already a GTPBot in OpenSpiel: https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/bots/gtp.py . Admittedly it's only in python for now, but it shouldn't be hard to translate into C++. 

This would incur a tiny bit of threading/communication overhead, but it shouldn't be too bad.

From a legal perspective, you could make the Stockfish GTP implementation open source under GPLv3, and host it on your own repo or even submit it to their repo.

ddobbelaere:
@tewalds the chess protocol you want is called UCI, as mentioned earlier: http://wbec-ridderkerk.nl/html/UCIProtocol.html

tewalds:
Ah right, even better. I doubt the internal AlphaZero has a UCI interface, though it wouldn't really matter either way as it's not public. The OpenSpiel AlphaZero implementation doesn't have a UCI interface, but it doesn't look like it'd be that hard, and would be pretty similar to the gtp.cc I linked above.

snicolet:
> One way that would almost certainly pass legal scrutiny is to run stockfish in a separate process that exports a common protocol.

Yes, that would be the way to go, using the UCI protocol. Good luck with your project, seems cool :-)

vofak:
Big thanks to @ddobbelaere for pointing out the legal issue.

I have updated the PR so that it now clones the stockfish repository and then builds and installs the binary. UciBot using the UCI then communicates with the installed stockfish. I tested the UCI with Stockfish and Ethereal and it seems to work fine.

I believe the legal issue is resolved then?

lanctot:
Great, thanks @vofak! I would still like to discuss with our legal team but I believe this indeed helps a lot!

lanctot:
Hi @vofak, I still see lots of references to Stockfish, and I believe downloading and installing Stockfish could still be considered "a link to GPL3 code". 

Could we please:

- Rename this bot to UCIBot
- Remove the BUILD_WITH_STOCKFISH option
- Remove all references to Stockfish specifically

Then we no longer need to make it a Stockfish-specific bot (it can be any chess bot that uses UCI), and we don't have to add any conditional flags, we can basically always build it into the main library.

I feel this is both (i) the safest option, (ii) cleaner, (iii) more general. WDYT?

vofak:
Hi @lanctot, I will rename the UciBot to UCIBot.

UCIBot can communicate with any UCI compliant chess engine. Only a path to an executable is needed.
The reason why I kept the stockfish as an optional dependency is that it provides an easy way for an OpenSpiel user to have a chess engine ready and available, and I could make an easy way to set options for this specific engine (because these options differ between engines). Of course, this is not that important for the functionality and I can easily delete everything related specificaly to Stockfish, if you think that it's safer.

lanctot:
I like it better too, just worried about the interpretation of "link" for determining what is allowed by GPL3. Doesn't the Stockfish binary get compiled along with OpenSpiel code? Let me have a chat with our legal team and see what we can do.

vofak:
The only connection to stockfish code is in install.sh. The script clones the stockfish repository and uses its makefile to build the binary and install it to /usr/local/bin. If the user already has stockfish installed, nothing is clones, built or installed.

The only reason to clone the repository (even a specific tag) and build from sources is to ensure compatibility of the uci options. These options can vary across uci compliant chess engines (and they tend to sometimes change with new versions of engines) and to make usage of stockfish in OpenSpiel as easy as possible, I created an easy way to set the available options for the version of stockfish installed in the script.

lanctot:
@vofak As I have said: I agree, and I prefer this version too, for all the reasons you list. 

But we will only pull it in once our legal team confirms that we are not violating any licensing requirements.

I can't believe that it's so difficult to find examples of what is considered a "link" for purposes of creating a derivative work. I am still surprised that your original PR was pronlematic. I would have thought that this case is common enough that it would be covered by an FAQ but there is suprisingly little information about what gets classified as a "link" and what doesn't. Please, if anyone can point us to something more  clear, that would help us out.

ddobbelaere:
I'm no layer, but I think as long as you don't touch the SF binary and communicate via UCI, there should be never any issue whatsoever. Rule of thumb is "does any of the program use same memory address space?". Clearly the answer is no and that case. I think "link" specifically points to linking (statically or dynamically) in technical meaning, in which case rule of thumb is no longer satisfied.

open_spiel/bots/stockfish/stockfish_bot.h
```diff
@@ -0,0 +1,86 @@
...
+    options_ = {};
+  }
+
+ private:
+  uci::Options options_;
+};
+
+std::unique_ptr<StockfishOptionsBuilder> MakeStockfishOptionsBuilder();
+
+std::unique_ptr<Bot> MakeStockfishBot(int move_time,
```
michalsustr, line 79: Please add last parameter a default path `std::string path_to_stockfish_binary = "stockfish"`

open_spiel/bots/uci/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_library(uci OBJECT
+        ../uci/uci_bot.cc
+        ../uci/uci_bot.h
+        )
```
michalsustr, line 4: add newline

open_spiel/bots/uci/uci_bot.cc
```diff
@@ -0,0 +1,296 @@
...
+#include <sys/wait.h>
+#include <iostream>
+#include <string>
+#include <unistd.h>
+#include <sys/ioctl.h>
+
+namespace open_spiel {
+namespace uci {
+
+UCIBot::UCIBot(const std::string &path,
```
michalsustr, line 26: rename `path` -> `bot_binary_path`

open_spiel/bots/uci/uci_bot.h
```diff
@@ -0,0 +1,79 @@
...
+namespace open_spiel {
+namespace uci {
+
+using Options = std::map<std::string, std::string>;
+
+class UCIBot : public Bot {
+ public:
+  explicit UCIBot(const std::string &path,
+                  int move_time,
+                  bool ponder,
```
michalsustr, line 32: Add an explanataion what is "ponder"

michalsustr: Also add some doc comment how you can construct the options for individual bots

open_spiel/scripts/install.sh
```diff
@@ -130,6 +130,16 @@ if [[ ${BUILD_WITH_ROSHAMBO:-"ON"} == "ON" ]] && [[ ! -d ${DIR} ]]; then
   git clone -b 'open_spiel' --single-branch --depth 1  https://github.com/jhtschultz/roshambo.git ${DIR}
 fi
 
+if [[ ${BUILD_WITH_STOCKFISH:-"ON"} == "ON" ]] && [[ ! -x `which stockfish` ]]; then
```
michalsustr, line 133: Add a small comment like 
```
# If build is enabled and stockfish binary is not found, clone stockfish sources and build 
# its binary, indepedent of the rest of the library -- we only use the standard UCI interface.
```

open_spiel/bots/stockfish/stockfish_bot.h
```diff
@@ -0,0 +1,86 @@
...
+namespace stockfish {
+
+enum StockfishAnalysisContempt {
+  kBoth,
+  kOff,
+  kWhite,
+  kBlack
+};
+
+class StockfishOptionsBuilder {
```
michalsustr, line 37: Add a comment with a link to manual that explains the configuration options.

open_spiel/bots/stockfish/stockfish_bot.h
```diff
@@ -0,0 +1,86 @@
...
+ public:
+  StockfishOptionsBuilder() { Reset(); };
+
+  StockfishOptionsBuilder *WithDebugLogFile(const std::string &debug_log_file);
```
michalsustr, line 41: (not necessary) but if you feel like it, you could adapt this macro https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/torch/csrc/api/include/torch/arg.h and add SPIEL_ARG :)

vofak:
@lanctot I have removed anything connected to stockfish specifically. The only thing left is now the UCIBot which can be used with any UCI engine. The result of this PR is not the one that it's name suggests but I hope there is no problem pulling it like this. If the legal team decides that cloning and installing stockfish is not a violation of their license, we can make a new PR.

lanctot:
@vofak wonderful, thanks. I have sent an email to our legal team. I hope that we can resolve this soon.

lanctot:
@vofak Apologies for the delay. 

I have changed the name of the PR to reflect your most recent changes.

There are now some conflicts with master. Could you please resolve those? I cannot import this until they are resolved.

Thanks again!

lanctot:
Hi @vofak, I am now ready to import this. However, I cannot until the conflicts are resolved. Can you resolve them?

lanctot:
Ah, I was able to do it myself -- perfect.

lanctot:
Hi @vofak @michalsustr , I have been unable to get a basic test working. I've opened an issue: https://github.com/deepmind/open_spiel/issues/681. So we will import this but will mark it as experimental until we can fix the test. I would appreciate if you could take a look once we merge it. Thanks.

PR #502, opened 2021-02-15
Fix typo "frp" to "for"

liyuqian:


lanctot:
Thanks, nice catch

liyuqian:
Thanks, @lanctot! I have an unrelated question: why is "information set" called "information state" in this repo?

lanctot:
Ah yes, because OpenSpiel is partly an attempt to bridge communities. I have many the last 10 years interacting with people who primarily research traditional RL and I found it easier to explain CFR using this terminology. 

In person I usually switch between them depending on who I an talking to :) I think my favorite in this topic is related counterfactual values and q-values, because it lets us compare the algorithms more directly.

open_spiel/algorithms/cfr.cc
```diff
@@ -423,7 +423,7 @@ void CFRSolverBase::GetInfoStatePolicyFromPolicy(
 // - alternating_player: Optionally only update this player.
 // - reach_probabilities: The reach probabilities of this state.
 // - current_player: Either a player or chance_player_.
-// - action_probs: The action probabilities to use frp this state.
```
goswamig, line 426: Testing

PR #506, opened 2021-02-19
Update julia version

### LINKED ISSUE #497: Problem with Julia API when installed in /usr/local
Hi @findmyway, we had to move our CI away from Travis and to Github Actions this week. The problem is that the Julia tests are now failing again, see here: https://github.com/jblespiau/open_spiel-1/runs/1595264912?check_suite_focus=true

The problem is similar to https://github.com/deepmind/open_spiel/issues/201 except the command I used before no longer works because I copied the library to the $HOME/packages/julias/julia-1.3.1. See here: https://github.com/deepmind/open_spiel/blob/32a3d303e0ce710f2ef70a3214d26640cde97bf7/open_spiel/scripts/install.sh#L197

Now the machines have Julia installed in /usr/local/julia1.5.3, and this might generally happen on the user side. So we'll probably need to do a sudo to copy the file. I think we can do this in our install script as long as we echo something like why we requiring sudo... but even if that works, it does not feel like an elegant solution. Do you have a better idea about how to fix this?

findmyway:
Close #497

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -31,7 +31,7 @@ jobs:
          # shared library with linkable third party dependencies in place.
         - os: ubuntu-20.04
           DEFAULT_OPTIONAL_DEPENDENCY: "ON"
-          BUILD_WITH_JULIA: "OFF"
+          BUILD_WITH_JULIA: "ON"
           TRAVIS_USE_NOX: 0
           BUILD_SHARED_LIB: "ON"
           BUILD_WITH_ORTOOLS: "ON"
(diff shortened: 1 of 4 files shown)


lanctot:
Great, thanks @findmyway .. just in time for an upcoming (small) release! 

Do you know when 1.6.0 is scheduled to release? I wonder if it makes sense to wait for it. Probably not -- we can just push this through to re-enable the tests and upgrade the version when it's released. WDYT?

findmyway:
> Great, thanks @findmyway .. just in time for an upcoming (small) release!
> 
> Do you know when 1.6.0 is scheduled to release? I wonder if it makes sense to wait for it. Probably not -- we can just push this through to re-enable the tests and upgrade the version when it's released. WDYT?

😄 I'm not sure when 1.6.0 will be released (It seems only [one blocking issue](https://github.com/JuliaLang/julia/milestone/36) left.) I think we don't need to wait for it. There is still more work left after fixing the github action issue (the binary builder, packaging pipeline). So I prefer to make another patch release later.

PR #509, opened 2021-02-20
Dark Hex Game added

BedirT:


lanctot (changes requested):
Two initial things: 

- I don't see dark_hex_test added to the CMakeLists.txt, please add it

- Please make the indentation style consistent with the rest of the code? ( in particular: don't indent nested namespaces)

Issue #499, opened 2021-02-13
python mccfr objects don't directly return Policies


VitamintK:
Is there a reason that the [MCCFR objects](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/external_sampling_mccfr.py) expose a method to get a callable policy (`callable_avg_policy()`) but not a way to get a [`Policy`](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/policy.py) object?

I think we could manually subclass from `Policy` and define `self.action_probabilities(state)` as `cfr_solver.callable_avg_policy(state)`, but is there a reason this isn't built into the MCCFR class already?

lanctot:
I _think_ this is really due to some legacy API from before we had policy objects; IIRC, we sent around callable "things" which represented policies, maybe to avoid storage of full policies? .. but I don't remember very clearly, @elkhrt do you remember... can you elaborate?

So we had to use this policy_from_callable helper to convert them when we made the change: https://github.com/deepmind/open_spiel/blob/6a9cba53cf3fe6ba2c7ed1f310e2e59ec7305484/open_spiel/python/algorithms/outcome_sampling_mccfr_test.py#L47

We should probably change these all indeed to just return policy objects. That would be make a good contribution, feel up to it? :)

colibri17:
Hey there! If @VitamintK has not started to work on it yet, I can contribute! Basically, the point is to make the method `callable_avg_policy` of outcome and external mccfr to return a policy object rather than a callable method, right?

VitamintK:
I have not started work on it, so feel absolutely free to do so.  Obviously @lanctot et al would know much better than me, but I imagine adding an `average_policy` method like in [non-monte carlo cfr](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/cfr.py) would be better for consistency with non-monte carlo cfr objects and backwards compatibility.  Not sure if you'd want to then delete the `callable_avg_policy` method though or keep it for backwards compatibility.

lanctot:
@colibri17 , that would be great! I would say let's aim to delete it and add the average_policy method to replace it. I would recommend doing these one instance at a time; for each one there may be many places in the code base that currently use the "callable_avg_policy" and it'll require switching them over to the average_policy. Maybe it'll be easier than I expect, in which one PR might do... but feel completely free to split these into one PR per instance.

colibri17:
[...]

PR #513, opened 2021-02-27
Average policy in mccfr

colibri17:
This pull requests is related to the issue https://github.com/deepmind/open_spiel/issues/499.

I created the class `AveragePolicy` which extends the base class `Policy` and implements the method `action_probabilities`. Within the solvers, the method `average_policy` initializes each time this class. @lanctot I mirrored what is done on the C++ side and this allows to return the average policy just for already explored infostates and using the random policy in the case of not explored infostates (as @VitamintK suggested). Also, I adapted the rest of the code accordingly. 
I finally compared the outputs of the tests of this new code version with the outputs of the tests of the previous one. The results are exactly the same. I report these here for future reference:
Outcome sampling:
Kuhn2P, conv = 0.010184780058560161
Kuhn3P, conv = 0.12295759857995159
Leduc2P, conv = 2.1751618625344777
External sampling:
Kuhn2P, conv = 0.7420193061644658
Kuhn2P, conv = 0.6532886929945988
Kuhn3P, conv = 1.2682464363671362
Kuhn3P, conv = 1.7098108280062134
Leduc2P, conv = 4.887844390143719
Leduc2P, conv = 4.813241731029391

As a side note: I tried to mirror the corresponding versions on the C++, but indeed between the Python and C++ versions there are some important differences. For example, I had to duplicate the class `AveragePolicy` for outcome and external sampling monte carlo, since there is no a single place to put these in Python (that would be `cfr.h`, `cfr.cc` in C++). I guess that these differences might be because the different Python counterfactual versions were developed in different moments. In my opinion a good issue might be to refactor the Python code for completely replicating the C++ side, even if this will require some effort.

@lanctot @VitamintK hope that this is what you had in mind! Please let me know if it fine to you :)


lanctot:
@colibri17, this looks great! Yes this is exactly what I was hoping for... really wonderful.

Right, I see the issue, not as easy to do here. I noticed cfr.py has some shared functions use by CFR-BR too but that doesn't seem like a good place either. 

Looks like in this case `AveragePolicy` is indeed the exact same class in both cases... how about we create a mccfr_utils.py or mccfr_helpers.py module and just put it in there, then both algorithms can just import that? There might be opportunity to move more stuff in there too but no need to do that right now.

colibri17:
Yes I think this is a good alternative. On the C++ side, methods, classes and structs for both pure CFR (cfr and cfr-br) and MCFFR (outcome sampling cfr and external sampling cfr) algorithms are placed within `cfr.h` and `cfr.cc`. But it seems to me that both in Python and in C++ the CFR algorithms have no methods or structures in common with the MCCFR ones, so it makes sense to split unrelated methods in separated modules.
Ideally we could think to refactor the code to make a super base class which contains the common structure shared by CFR and MCCFR algorithms (at the end CFR and MCCFR are both part of counterfactual algorithms). However, this is quite a lot of work and I think we would end up with complicated code, which is against the philosophy of the library of keeping things easy. 
These are my 2 cents! :)


Back to the pull requests, yes the `AveragePolicy` is exactly the same. 
Comparing the external and outcome sampling versions, there are also other methods which are exactly the same and could be moved in the `mccfr_utils.py` (look for example at `_add_avstrat`, `_add_regrets`, `_lookup_infostate_info`). For these we could create a base class containing these methods. 
If you agree, I can move `AveragePolicy` in a new module called `mccfr_utils.py `, and then open another pull request for moving `_add_avstrat`, `_add_regrets`, `_lookup_infostate_info` in `mccfr_utils.py `. Maybe, if this is the case, this new module could be just called `mccfr.py`?

lanctot:
Hi @colibri17 , yes that sounds perfect-- let's do that (common base class in mccfr.py to put AveragePolicy and the shared methods)!

colibri17:
Great, I am going to do the changes! Will move the shared methods in another pull request.

You're welcome! :)

lanctot:
Seems like our testing is broken on MacOS. Will fix it tomorrow and then import this. Thanks!

lanctot:
> Seems like our testing is broken on MacOS. Will fix it tomorrow and then import this. Thanks!

Nevermind, seems like you got unlucky and github actions CI failed. They are running properly now.

PR #421, opened 2020-10-25
Add Game - Dark Chess

vofak:
Adding a game [Dark chess](https://en.wikipedia.org/wiki/Dark_chess). It is an imperfect version of traditional chess. Since there are also some changes to the rules other then just the hidden information, this draft pull request only contains the perfect  information version of dark chess so far.

PR #517, opened 2021-03-02
Add DarkChess

michalsustr:
This is extension of #421 (that PR can be closed by this PR) -- I updated the code a bit and tested it. Now it is ready to be merged. The playthroughs are rather large, the dark_chess.txt is 2.6MB as there is a number of observation tensors that get printed out and the information state is stored as string concatenation of AOH. If you consider this a problem, maybe compression would be useful: a 7zip file is 18KB.

There is also a vizualization of the game playthrough in SpielViz:

image


vofak:
@googlebot I consent.

lanctot:
Hi guys, great! How is it related to https://github.com/deepmind/open_spiel/pull/421? Should we close 421?

michalsustr:
Yes you can close #421

BedirT:
The viz is amazing @michalsustr I didn't really know there was a GUI implementation on open_spiel. This could help a lot in my research actually. Could you guys direct me to where to check for implementing the GUI? @lanctot for the dark_hex I would love if we had the visuals (Ryan would more than me tbh :) )

michalsustr:
@BedirT Check out [SpielViz](github.com/michalsustr/spielviz), I'm happy to take in pull requests. It uses PyGTK. I think the easiest way is to make code that builds an SVG picture that is embedded in state view (like for Chess). This is also relatively portable, if you decide for example to make a web GUI with Javascript. Hit me up on e-mail or maybe even better open an issue on the repo, I'm happy to help.

michalsustr:
@vofak FYI DarkChess is merged.

brianhzhang:
Currently in the dark chess implementation, it seems to me that the 50-move counter and repetition counter are public information. In particular, this means that it is always common knowledge when a pawn move is made (but of course which pawn move is not public). Is this intentional? It reduces dramatically the amount of information that is private...

PR #520, opened 2021-03-02
Moved shared methods of outcome and external mccfrs into base class

colibri17:
I moved the shared methods of outcome and external mccfrs into the base class `_MCCFRSolverBase`, implemented in the module `mccfr.py`. Specifically, the methods moved are the following: `_lookup_infostate_info`, `_add_regret`, `_add_avstrat`, `average_policy`, `_regret_matching`.

Also, I noticed that the external sampling version had an initialization constant of `np.ones(num_legal_actions, dtype=np.float64) / 1000` for regrets and average strategy of a new infostate rather than `np.ones(num_legal_actions, dtype=np.float64) / 1e6` (method `_lookup_infostate_info`). I moved to `1/1e6`, as it is done on the C++ side.
As a result, the external mccfr test outputs slightly changed. I report here the results (comparison can be done with results reported [here](https://github.com/deepmind/open_spiel/pull/513#event-4389559891) ):
Kuhn2P, conv = 0.742222018437585
Kuhn2P, conv = 0.652778289351194
Kuhn3P, conv = 1.2682291839371136
Kuhn3P, conv = 1.7094910613902536
Leduc2P, conv = 4.887937425632959
Leduc2P, conv = 4.813561937379362

The outcome mccfr test outputs remained the same (as expected).

### DIFF
--- a/open_spiel/python/algorithms/cfr.py
+++ b/open_spiel/python/algorithms/cfr.py
@@ -114,7 +114,7 @@ def _update_average_policy(average_policy, info_state_nodes):
 
 
 class _CFRSolverBase(object):
-  r"""A base classe for both CFR and CFR-BR.
+  r"""A base class for both CFR and CFR-BR.
 
   The main iteration loop is implemented in `evaluate_and_update_policy`:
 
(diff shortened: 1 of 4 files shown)


lanctot:
Ok yes, 1e6 is better-- thanks!

PR #526, opened 2021-03-06
mccfr.AveragePolicy: make to_tabular() work

VitamintK:
`mccfr.AveragePolicy.to_tabular()` was erroring with `AttributeError: 'AveragePolicy' object has no attribute 'game'`.  This fixes it and adds 2 regression tests.


lanctot:
Great, thanks!

Issue #525, opened 2021-03-06
Problem with PolicyBot (Python) used together with AveragePolicy from MCCFR (C++)


agoryelov:
**Background:** I am using C++ External Sampling MCCFR with a custom Python game (Liar's Dice) as per the example at `mccfr_cpp_example.py`. PolicyBot is used to evaluate the model against a simple hard-coded Policy.

**Problem:** It seems that action probabilities dict returned from `solver.average_policy()` are shuffled which causes the PolicyBot to improperly choose an action.

I was able to fix this locally by using policy keys as legal actions:
```
(-) action = self._rng.choice(legal_actions, p=list(policy.values()))
(+) action = self._rng.choice(list(policy.keys()), p=list(policy.values()))
```

lanctot:
Thanks @agoryelov. I am trying to understand. Are you saying that the PolicyBot is making an assumption about how actions are ordered? Can you point us to exactly where that assumption is made and if our one-line solution fixes the problem, can you submit a PR?

agoryelov:
> Are you saying that the PolicyBot is making an assumption about how actions and probabilities are ordered?

Yeah, I believe that's what it is.

In `/open_spiel/open_spiel/python/bots/policy.py`, the PolicyBot uses `state.legal_actions()` to get a list of actions and `policy.action_probabilities().values()` to get an list of their probabilities. Since the `action_probabilities()` returns a dict, there is no guarantee that the order of the probabilities will be the same as the order of `state.legal_actions()`. 

Using the keys of the probabilities dict ensures the actions are in the correct order, but it would potentially allow the bot to make illegal actions. If that is an acceptable assumption, I can open a PR with this change. Alternatively, the bot can also validate whether action_probabilities has any illegal actions.

lanctot:
Yeah that sounds good, please submit a fix. 

There are currently nothing preventing users from constructing policies that assign positive probability to illegal actions, things will just fail if you do it. So it's already the user's responsibility so your fix would not allow anything that's not already allowed.

lanctot:
I just took a look at that class and was surprised to see that it uses the legal actions.. that class should not call legal actions at all, it should justt query the policy. Can you remove that and also modify the comment at the top?

Alternatively we can support both setups with a flag. The legal actions one should only consider legal actions (and reconstruct a policy by normalizing by the sum of probs over legal actions.. in this option we can also assume prob 0 for any actions not present in the dict.)

[...]

PR #527, opened 2021-03-07
Using policy keys instead of legal actions to sample

agoryelov:
Changes as discussed in #525


agoryelov:
@googlebot I signed it!

lanctot:
Hi @agoryelov after an internal review, we decided to just go with the original suggest and and just not do anything with legal moves in this class. It makes the class simpler and consistent with the C++ side. And people can add that logic if necessary, but essentially the policy should be well-formed and we should be able to assume that.

agoryelov:
Sounds good, glad to have this resolved either way

PR #529, opened 2021-03-09
Move GetOptimalPolicy for the game of Kuhn Poker to kuhn_poker namesp…

michalsustr:
Move GetOptimalPolicy for the game of Kuhn Poker to kuhn_poker namespace. Expose the same function via pybind.

The file `open_spiel/python/games/data.py` can be also deleted, along with its test (which can be moved to c++ side). I can do this in a subsequent PR.

### DIFF
--- a/open_spiel/algorithms/best_response_test.cc
+++ b/open_spiel/algorithms/best_response_test.cc
@@ -39,29 +39,6 @@ namespace {
 
 using InfostatesAndActions = std::vector<std::pair<std::string, Action>>;
 
-// The optimal Kuhn policy is taken directly from the Python implementation in
-// open_spiel/python/algorithms/exploitability_test.py. In it, the player bets
-// with probability alpha with the jack.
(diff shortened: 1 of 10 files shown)


michalsustr:
Not sure why the job `build_and_test / build (ubuntu-20.04, 3.8, 1, OFF, OFF, OFF) (pull_request)` failed... it complains the file `pybind11/games_kuhn_poker.cc` is missing, but it's in the repo, and other tests that build pyspiel worked... Maybe the test is acting up?

lanctot:
Restarted the test.

michalsustr:
Hmm weird, indeed it was missing! I definitely implemented it... but maybe forgot to add it... -- Update: I think it might be because 
`pybind11/` is in .gitignore and somehow it also matched this directory (though I guess it should only match the top-level pybind/) so I didn't add those files.

lanctot:
Very nice change, thanks for doing this!

PR #540, opened 2021-03-19
Add kriegspiel

vofak:
[Kriegspiel](https://en.wikipedia.org/wiki/Kriegspiel_(chess)) is an imperfect information chess variant, where players cannot see opponent pieces. Player is informed about the state of the game only from laconic messages from the referee.


lanctot:
Cool, a true classic, thanks!!

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+inline constexpr int NumPlayers() { return 2; }
+inline constexpr double LossUtility() { return -1; }
+inline constexpr double DrawUtility() { return 0; }
+inline constexpr double WinUtility() { return 1; }
+
+// See action encoding below.
+inline constexpr int NumDistinctActions() { return 4672; }
+
+// https://math.stackexchange.com/questions/194008/how-many-turns-can-a-chess-game-take-at-maximum
+inline constexpr int MaxGameLength() { return 17695; }
```
michalsustr, line 61: I think this is higher

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+#include <open_spiel/fog/observation_history.h>
+
+#include "open_spiel/abseil-cpp/absl/container/flat_hash_map.h"
+#include "open_spiel/abseil-cpp/absl/algorithm/container.h"
+#include "open_spiel/games/chess/chess_board.h"
+#include "open_spiel/games/chess.h"
+#include "open_spiel/spiel.h"
+#include "open_spiel/spiel_utils.h"
+
+// Kriegpsiel - imperfect information version of chess:
```
michalsustr, line 33: I think you said you follow the ICC rules, can you put a link here please?

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+  kFile = 1,
+  kRank = 2,
+  kLongDiagonal = 3,
+  kShortDiagonal = 4,
+  kKnight = 5
+};
+
+std::string CheckTypeToString(KriegspielCheckType check_type);
+
+std::pair<KriegspielCheckType, KriegspielCheckType> GetCheckType(const chess::ChessBoard &board);
```
michalsustr, line 82: Why a pair? Maybe a comment would be helpful

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+  void MaybeGenerateLegalActions() const;
+
+  absl::optional<std::vector<double>> MaybeFinalReturns() const;
+
+  // We have to store every move made to check for repetitions and to implement
+  // undo. We store the current board position as an optimization.
+  std::vector<std::pair<chess::Move, KriegspielUmpireMessage>> move_msg_history_;
+  // We store this info as an optimisation so that we don't have to compute it
+  // from move_msg_history for observations
+  std::optional<KriegspielUmpireMessage> last_public_msg{};
```
michalsustr, line 188: last_public_msg_

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+  std::optional<KriegspielUmpireMessage> before_last_public_msg{};
+  // Moves that the player tried and were illegal. We don't let player try them
+  // again on the same board because they are clearly still illegal;
+  std::set<chess::Move> illegal_tried_moves_;
```
michalsustr, line 192: Not sure about this, but I think having a vector and looping through it will be faster than a set, especially when the number of moves is small.

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+  // We store the start board for history to support games not starting
+  // from the start position.
+  chess::ChessBoard start_board_;
+  // We store the current board position as an optimization.
+  chess::ChessBoard current_board_;
+
+  // cached ActionObservationHistory for each player
+  std::vector<open_spiel::ActionObservationHistory> aohs_;
+
+  bool repetition_3_fold;
```
michalsustr, line 202: postfix_ for classes, nopostfix for structs: https://google.github.io/styleguide/cppguide.html

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+  }
+
+  if (rule_50_move && Board().IrreversibleMoveCounter() >= 50) {
+    return std::vector<double>{DrawUtility(), DrawUtility()};
+  }
+
+  return std::nullopt;
+}
+
+std::string DefaultFen(int board_size) {
```
michalsustr, line 719: Refactor and join with DC implementation

michalsustr (commented):
Some initial comments

open_spiel/games/kriegspiel.h
```diff
@@ -0,0 +1,276 @@
...
+
+std::string CheckTypeToString(KriegspielCheckType check_type);
+
+std::pair<KriegspielCheckType, KriegspielCheckType> GetCheckType(const chess::ChessBoard &board);
+
+struct KriegspielUmpireMessage {
+
+  bool illegal_ = false;
+  KriegspielCaptureType capture_type_ = KriegspielCaptureType::kNoCapture;
+  chess::Square square_ = chess::InvalidSquare();
```
michalsustr, line 88: kInvalidSquare, also at some other places... doesn't compile

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+                             chess::kEmptyPiece};
+
+    for (auto move_msg = state.MoveMsgHistory().rbegin();
+         move_msg != state.MoveMsgHistory().rend(); ++move_msg) {
+      if (move_msg->first.piece.color == color) {
+        last_move = move_msg->first;
+        break;
+      }
+    }
+    WriteMove(last_move, state.Board(), prefix + "last_move", allocator);
```
michalsustr, line 321: _last_move

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+    // 5 is maximum because we can't promote to a pawn.
+    WriteScalar((int8_t) move.promotion_type, 0, 5, prefix + "_promotion", allocator);
+  }
+
+  void WriteUmpireMessage(const KriegspielUmpireMessage &msg,
+                          const chess::ChessBoard& board,
+                          const std::string &prefix,
+                          Allocator *allocator) const {
+    int board_size = board.BoardSize();
+    WriteScalar(msg.capture_type_, 0, 2, "_capture_type", allocator);
```
michalsustr, line 244: "capture_type" or "public_capture_type"

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+    auto square_out = allocator->Get(prefix + "_captured_square",
+                                     {board_size, board_size});
+    for (int8_t y = 0; y < board_size; ++y) {
+      for (int8_t x = 0; x < board_size; ++x) {
+        const chess::Square square{x, y};
+        square_out.at(x, y) = square == msg.square_ ? 1.0f : 0.0f;
+      }
+    }
+    WriteScalar(msg.check_types_.first, 0, 5, "_check_one", allocator);
```
michalsustr, line 253: public_check_one , not sure if this is public or private

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+    WriteScalar(msg.check_types_.second, 0, 5, "_check_two", allocator);
```
michalsustr, line 254: likewise

open_spiel/games/kriegspiel.cc
```diff
@@ -0,0 +1,747 @@
...
+    WriteScalar((int8_t) msg.to_move_, 0, 2, "_to_move", allocator);
```
michalsustr, line 255: likewise

michalsustr:
A history "1841, 1869, 2453, 1285, 2425" causes an error `observer.h:72 m == data.size() m = 3, data.size() = 0`

michalsustr:
The same problem also happens for small kriegspiel: history "1257, 1882, 701" for "kriegspiel(board_size=4)"

open_spiel/games/kriegspiel.h
vofak: @michalsustr You're right. It is implemented in a way that attempting an illegal move is still considered a move. I can put a loose upper bound there. However, with the threefold repetition and 50 move rule turned off, the gamecan hypothetically go on forever.

michalsustr:
Can you submit PGN in a distinct pull request after this is merged? I believe it is not necessary for Kriegspiel implementation right away and it bloats the size of the PR. Thanks!

open_spiel/games/chess/chess_board.cc
```diff
@@ -484,8 +499,8 @@ void ChessBoard::GenerateLegalMoves(const MoveYieldFn &yield,
   }
 }
 
-void ChessBoard::GeneratePseudoLegalMoves(const MoveYieldFn &yield,
-                                          Color color) const {
+void ChessBoard::GeneratePseudoLegalMoves(const MoveYieldFn &yield,Color color,
+                                          bool ignore_enemy_pieces) const {
```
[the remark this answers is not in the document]
michalsustr: Also, I think this should be passed as a config flag to the chess board, in the same way king in check currently is.

michalsustr:
@lanctot I think the test failed because of #559 -- it worked fine on @vofak laptop. Can you please rerun the checks?

lanctot:
I suspect it's unrelated because it would not even have started if it was affected by the bug that #559 fixed (the build script would have failed with a parse error), whereas your test seems to be failing for real: 

```
ERROR: test_observations_raises_error_on_invalid_player (absl.testing.parameterized.EnforceAPIPartialTree_kriegspiel_Test)
absl.testing.parameterized.EnforceAPIPartialTree_kriegspiel_Test.test_observations_raises_error_on_invalid_player
----------------------------------------------------------------------
  File "/Users/runner/work/open_spiel/open_spiel/open_spiel/python/../integration_tests/api_test.py", line 336, in test_observations_raises_error_on_invalid_player
    v = state.observation_tensor(p)
pyspiel.SpielError: /Users/runner/work/open_spiel/open_spiel/open_spiel/games/kriegspiel.cc:245 val <= max
val = 46, max = 2

----------------------------------------------------------------------
Ran 454 tests in 626.677s

FAILED (errors=1)

183/188 Test #160: python_sampled_stochastic_games_test ..............   Passed  189.30 sec
184/188 Test #126: python_outcome_sampling_mccfr_test ................   Passed  351.28 sec
185/188 Test #158: python_pyspiel_test ...............................   Passed  207.10 sec
186/188 Test #157: python_policy_test ................................   Passed  214.00 sec
187/188 Test #136: python_value_iteration_test .......................   Passed  296.80 sec
188/188 Test #187: spiel_test ........................................   Passed  108.52 sec

99% tests passed, 2 tests failed out of 188

Total Test time (real) = 909.45 sec

The following tests FAILED:
	 58 - kriegspiel_test (Failed)
	 94 - python_api_test (Failed)
Errors while running CTest
nox > Command ctest -j12 --output-on-failure failed with exit code 8
nox > Session tests failed.
Error: Process completed with exit code 1.
```

I'll restart it anyway.

lanctot:
This is failing a test in a case where we are doing a check on an invalid player's turn. Likely all you are missing is a check for failure at the start of the function, and the logic of the function would cause it to go out of bounds or some data to not be well set.

michalsustr:
I just talked with @vofak and the missing `kriegspiel.txt` playthrough has a whopping 50MB, compressed with 7zip is only 100kB. A similar issue occurred in #517 only 10x smaller (5MB). I will make a PR that can read compressed playthrough files, WDYT @lanctot @elkhrt ?

lanctot:
My personal opinion is that when the playthrough is that big, there's really only questionable value of having it at all, and part of the value of playthroughs is to get a quick inspection of what the game looks like visually, which can't happen if it's compressed. Also it would add yet another dependency. So I'm favor of just leaving it out. I know Ed loves having them.. but, I really don't see the argument for them in these cases. This is just spam :)

michalsustr:
That makes sense, I am in favor of keeping unnecessary dependency bloat low. Another solution I thought of would be to limit the length of the replay to just 100 states or so -- this would be an easy universal solution. WDYT?

lanctot:
> Another solution I thought of would be to limit the length of the replay to just 50 states or so -- this would be an easy universal solution. WDYT?

Yeah I'd be happy with that. @elkhrt, wdyt?

michalsustr:
@vofak The names in the observation tensor seem off, shouldn't it be `public_illegal` instead of  `_illegal`?  Along with: `_illegal: [2], _capture_type: [3], _captured_square: [4, 4], _check_one: [6], _check_two: [6], _to_move: [2], _pawn_tries: [16]` (as seen in the playthroughs)

michalsustr:
@vofak Still some problems with MacOS... :/ Can you temporarily "hack" one of the consistently failiing test to get a history trace (State::HistoryString() ) so we can check what's going on for a particular history?

I'd like to have this in until the end of this week so I can build on top of it for RBC ...

vofak:
@michalsustr In the latest test, even the kriegspiel test fails . There you can see that the error occurs during generation of the observation tensor before the first move in the method WriteScalar with the argument max=2 (the error is that the passed value is larger than the max value). There is only one call with max=2, which is also the first time I try to pass an enum (enum KriegspielCaptureType : int8_t) as an integer value. The enum has values 0, 1 or 2 so there should really be no way that a value larger than 2 gets passed to the method.
Since the error occurs only on MacOS, might it be that the implicit cast from enum to integer works different there? Or is there something I am getting wrong about it? It is also weird that the first 3 games run without a problem and the fourth one fails before the first move. I admit I am lost.

lanctot:
This wouldn't surprise me especially with int8 (which I have found to be error prone) eso if you are casting enums. Could you try add an explicit cast? Also can you add an explicit range check and maybe and run in debug mode? Maybe we will get a stack trace.

I can also try finding it on my Mac if you let me know how to reproduce it.

open_spiel/games/kriegspiel.cc
```diff
@@ -223,7 +223,7 @@ class KriegspielObserver : public Observer {
                           Allocator *allocator) const {
     int board_size = board.BoardSize();
     WriteBinary(msg.illegal, prefix + "_illegal", allocator);
-    WriteScalar(msg.capture_type, 0, 2, prefix + "_capture_type", allocator);
+    WriteScalar(static_cast<int8_t>(msg.capture_type), 0, 2, prefix + "_capture_type", allocator);
```
lanctot, line 226: Can you please check that msg.capture_type is in the proper range at the start of this function?

open_spiel/games/kriegspiel.cc
```diff
@@ -232,8 +232,8 @@ class KriegspielObserver : public Observer {
         square_out.at(x, y) = square == msg.square ? 1.0f : 0.0f;
       }
     }
-    WriteScalar(msg.check_types.first, 0, 5, prefix + "_check_one", allocator);
-    WriteScalar(msg.check_types.second, 0, 5, prefix + "_check_two", allocator);
+    WriteScalar(static_cast<int8_t>(msg.check_types.first), 0, 5, prefix + "_check_one", allocator);
```
lanctot, line 235: Same question as above for these two values

lanctot (commented):
Can you please print an error message containing the history string if the values are outside of the expected range? I am happy to help find this on my MacOS, but I need more information.

lanctot:
Note also the error message:

```
ERROR: test_observations_raises_error_on_invalid_player (absl.testing.parameterized.EnforceAPIPartialTree_kriegspiel_Test)
absl.testing.parameterized.EnforceAPIPartialTree_kriegspiel_Test.test_observations_raises_error_on_invalid_player
```

This observation is being called on an /invalid player/. Possibly this problem occurs only when it's not a valid player observing?

lanctot:
Also, it is failing right here: https://github.com/deepmind/open_spiel/blob/93c2ebd9479389aff3a00be1c83ff70615f357e9/open_spiel/integration_tests/api_test.py#L336

If you put a try around that line, int he catch part you can print out the state's ToString and the history string to help figure out what is going on. Can you do that?

lanctot:
image

image

michalsustr:
Can you please summarize what were the problems and the solutions for future reference? Thanks!

elkhrt:
Importing this now
FWIW, I agree that very long playthroughs are of limited value, but I think the right solution is to have a playthrough which shows a short game played all the way to the end, rather than to cut the playthrough off mid-game.

PR #549, opened 2021-03-30
Creation of Poker subgames.

michalsustr:
This PR implements creation of (possibly randomized) subgames in full-deck game of Poker, i.e 4 suits, 13 cards.

I attempt to replicate the [libratus endgames](https://github.com/Sandholm-Lab/LibratusEndgames) although I am not entirely certain I understood the format correctly. If you happen to know exploitability of some small subgame (like subgame 4) that would be a good test to make.

It is now possible to create these subgames either by calling `LoadGame("universal_poker(...)")` with appropriately set parameters or by calling `leduc_poker::MakeRandomSubgame( ... params ...)`. See the tests for an example.

The code can be extended to general implementation (different settings than HUNL) in subsequent PR.

Feel free to update the nomenclature / other parts of the code, I am not so well-versed in Poker terminology :)

@lifrordi  can you take look?


michalsustr:
Oh, and I also had to change `basic_tests` to make checking of the legal action mask optional, as it doesn't work with large action numbers; however I couldn't decompose the problem to use smaller action numbers unless it would make the implementation significantly more complicated.

michalsustr:
One test is failing due to not finding the proper test file... I added this and thought it should work, as it should be absolute path?
https://github.com/deepmind/open_spiel/pull/549/files#diff-8c2ac33c5241a858de3fe1534cfe8cf3bd9ba7e9de52be3fe036181b05f3ad73R578

lifrordi:
Thanks, @ElnazDavoodi can you please have a look?

open_spiel/games/universal_poker.cc
```diff
@@ -466,33 +524,216 @@ std::unique_ptr<State> UniversalPokerState::Clone() const {
...
+                            std::vector<std::vector<uint8_t>>)> yield) {
+  // Taken and modified from CardSet::SampleCards()
+  int nbCards = num_hole_cards * num_players;
+
+  uint64_t p = 0;
+  for (int i = 0; i < nbCards; ++i) {
+    p += (1 << i);
+  }
+  // Enumerates all the uint64_t integers that with nbCards 1-bits.
+  // The final n is ignored. It is fine as long as the rank < 16.
```
ElnazDavoodi, line 582: there should be a check here to validate rank<16, if not appropriate error message should be thrown

open_spiel/games/universal_poker.cc
```diff
@@ -1050,6 +1320,68 @@ open_spiel::Action ACPCActionToOpenSpielAction(
...
+        cards.push_back(card);
+      }
+    }
+    logic::CardSet set(cards);
+    board_cards = set.ToString();
+  }
+  if (hand_reach.empty()) {
+    // FIXME open interval [0, 1) -> closed interval [0, 1]
+    // Can be done using a bijection of 2^24 int to float
+    std::uniform_real_distribution<double> dist(0.0, 1.0);
```
ElnazDavoodi, line 1375: how about this:
std::uniform_real_distribution<> dis(0.0, std::nextafter(1.0, DBL_MAX));

open_spiel/games/universal_poker.cc
```diff
@@ -1050,6 +1320,68 @@ open_spiel::Action ACPCActionToOpenSpielAction(
...
+      "potSize=%d,"
+      "boardCards=%s,"
+      "handReaches=%s"
+    ")";
+
+  if (pot_size == -1) {
+    // Multiples of 50, from 0 to 40000
+    // FIXME maybe there are other possible pot sizes?
+    std::uniform_int_distribution<int> dist(0, 800);
+    pot_size = dist(rng) * 50;
```
ElnazDavoodi, line 1350: maybe it's better to define 50 as variable, in case we want to try other values than 50, it's easier to find it and change the value.

open_spiel/games/universal_poker_test.cc
```diff
@@ -550,6 +551,98 @@ void TestFCHPA() {
...
+  // ...
+  check_index("Ad", "Ac", 1325);
+}
+
+std::string ReadSubgameReachProbs(const std::string& file_name) {
+  std::string dir = __FILE__;
+  dir.resize(dir.rfind("/"));
+  return file::ReadContentsFromFile(
+      absl::StrCat(dir, "/universal_poker/endgames/", file_name, ".txt"), "r");
+}
```
ElnazDavoodi, line 583: Suggestion: maybe it's more readable to define the value of reach probabilities for each subgame instead of reading it from a file!

ElnazDavoodi (commented):
Looks good to me, I just have some suggestions.

open_spiel/games/universal_poker_test.cc
michalsustr: I didn't want to do that because it's an array of 2652 values, a bit too much for a source file.

open_spiel/games/universal_poker.cc
michalsustr: Cards with rank >= 16 are not representable in CardSet, so there is nothing to check. I updated the comment.

open_spiel/games/universal_poker.cc
michalsustr: Thanks for the suggestion!

open_spiel/games/universal_poker.cc
michalsustr: Now that I think of it these should match the HUNL game string, I updated the comment and range of bets.

michalsustr:
I just noticed I made a mistake, the chance probs would implicitly convert from int to doubles, this would lead to a very hard-to-detect bug ... I think a good way to really make sure the code is correct is to compute subgame's game value in another codebase and check this is indeed correct (via sequence-form LP on the loaded game).

lanctot:
@michalsustr @ElnazDavoodi is this ready to import now?

michalsustr:
Yes.

PR #552, opened 2021-04-06
Create wheels.yml actions workflow

lanctot:
Use cibuildwheel to build source and binary distributions for pip releases


lanctot:
Testing this on my own fork, so will close here and make several changes in the coming weeks necessary to get this working.

PR #554, opened 2021-04-07
Gin Rummy updates

jhtschultz:
This is a draft PR to track progress on updates to gin rummy in preparation for the HIGC.

- [x] Add Observer.
- [x] Include perfect recall.
- [x] Make game size adjustable by parameterizing deck size and hand size.
- [x] Finalize observation format.
- [x] Add simple baseline bot.
- [x] Update playthroughs.
- [x] Add more thorough tests - particularly of observation tensors. (Accomplished using simple bot example. The bot relies on decoding the observation tensor, and has successfully played over 1M games without an error.)

Tagging @michalsustr


open_spiel/games/gin_rummy.cc
```diff
@@ -57,8 +57,194 @@ std::shared_ptr<const Game> Factory(const GameParameters& params) {
...
+        iig_obs_type_(iig_obs_type) {}
+
+
+  void WriteTensor(const State& observed_state, int player,
+                   Allocator* allocator) const override {
+    auto& state = open_spiel::down_cast<const GinRummyState&>(observed_state);
+    SPIEL_CHECK_GE(player, 0);
+    SPIEL_CHECK_LT(player, state.num_players_);
+
+    // TODO(Michal) Should we implement perfect recall?
```
michalsustr, line 85: If possible, yes. But AFAIK Gin can have various histories lengths, resulting in a lot of "zeros padding" right? Is there a tight upper bound on the length? If it is >>50 it's probably too long.

open_spiel/games/gin_rummy.cc
```diff
@@ -57,8 +57,194 @@ std::shared_ptr<const Game> Factory(const GameParameters& params) {
...
+    }
+  }
+
+  std::string StringFrom(const State& observed_state,
+                         int player) const override {
+    auto& state = open_spiel::down_cast<const GinRummyState&>(observed_state);
+    SPIEL_CHECK_GE(player, 0);
+    SPIEL_CHECK_LT(player, state.num_players_);
+
+    // TODO(Michal) Should we implement perfect recall?
```
michalsustr, line 117: I think we should do it for strings, as that is the canonical way of identifiying infostates (for tabular algs etc.)

open_spiel/games/gin_rummy.cc
```diff
@@ -665,6 +771,28 @@ GinRummyGame::GinRummyGame(const GameParameters& params)
       undercut_bonus_(ParameterValue<int>("undercut_bonus")) {
   SPIEL_CHECK_GE(knock_card_, 0);
   SPIEL_CHECK_LE(knock_card_, kDefaultKnockCard);
+  default_observer_ = std::make_shared<GinRummyObserver>(kDefaultObsType);
+  // TODO(Michal) Should we implement perfect recall?
```
michalsustr, line 775: See above.

open_spiel/games/gin_rummy.cc
```diff
@@ -665,6 +771,28 @@ GinRummyGame::GinRummyGame(const GameParameters& params)
...
+  //info_state_observer_ = std::make_shared<GinRummyObserver>(kInfoStateObsType);
+}
+
+// TODO(Michal) This is how it's implemented in dark chess. The commented out
```
michalsustr, line 779: I think the observer registration in Leduc is something (relatively) new, I am not sure what is the intended usage... Maybe @elkhrt can explain?

open_spiel/games/gin_rummy_test.cc
```diff
@@ -532,6 +533,32 @@ void OklahomaTest() {
...
+  Observation observation = Observation(*game, observer);
+
+  std::unique_ptr<open_spiel::State> state = game->NewInitialState();
+  std::vector<Action> initial_actions;
+  initial_actions = {1,  4,  5,  6,  17, 18, 19, 30, 31, 32, 2,  3,
+                     16, 29, 43, 44, 45, 7,  20, 33, 0,  52, 55, 1};
+  for (auto action : initial_actions) state->ApplyAction(action);
+  std::cout << state->ToString() << std::endl;
+
+  std::cout << observation.StringFrom(*state, 0) << std::endl;
```
michalsustr, line 552: Can be simplified to `for(int player = 0; player < 2; player++) {` ...

open_spiel/games/gin_rummy.cc
jhtschultz: There is no tight upper bound and this made me realize that the max game length needs to be enforced as there are pathological action sequences that can go on indefinitely. This is too long for the perfect recall tensor, but I did implement it for strings following your suggestion. I believe the observer functionality now matches that of dark chess.

open_spiel/games/gin_rummy.cc
jhtschultz: Done, see above.

open_spiel/games/gin_rummy_test.cc
jhtschultz: Fixed.

jhtschultz:
This most recent commit adds the ability to adjust the size of the game with the intention of providing smaller game sizes for research and testing - similar in spirit to smaller poker variants. This required some refactoring of the gin rummy utils, but fortunately little to no changes to the core game logic. The existing gin tests all pass, but alternate configurations have only been lightly tested.

In order to make the game truly adjustable, it would be nice to have the option of melds of size 2 (currently melds must be >= 3 cards). This would require more extensive changes, including changes to the action space, which is already somewhat complicated. Before going down that road, this seems like a good time to assess the current changes. @michalsustr I tagged you in some TODOs related to the observation tensor format. We should nail that down so it's stable for the competition.

open_spiel/games/gin_rummy.cc
michalsustr: Great!

michalsustr:
Is it important to have melds of size 2? I think having a deck with just 2 card suits and 4 ranks should be interesting enough, so the meld size wouldn't need to change.

open_spiel/games/gin_rummy.h
```diff
@@ -35,53 +35,71 @@
 //  "knock_card"      int    set a specific knock card     (default = 10)
 //  "gin_bonus"       int    bonus for getting gin         (default = 25)
 //  "undercut_bonus"  int    bonus for an undercut         (default = 25)
+//  "num_ranks"       int    num ranks in deck             (default = 13)
```
michalsustr, line 38: Please add input checks, like `num_ranks * num_suits` >= `hand_size * 2` (maybe I forgot +1 for the revealed card?) and others you can think of, that make up (implicit) invariants the code relies on and could crash (if supplied wrongly).

michalsustr (commented):
Looks nice! :) Great job! Sorry my turnaround takes a bit these days (NeurIPS coming...). 

> We don't even need to introduce CardSet, which I found actually presents some issues of its own (requires ACPC be installed, views ace as high when it's always low in gin).

I know you didn't like CardSet for these reasons, but I think it might be still worth implementing  (a variant of) it instead of VecInt, especially when we upper bound the implementation to work with standard card deck. I think it is fine not doing it now, maybe I will take a look into it sometime later after deadlines :)

open_spiel/games/gin_rummy.h
jhtschultz: Good call. I added the first check along with a few others. We no longer need the second one because I removed kMaxStockSize altogether. I realized it was only used to reserve 31 slots in the observation tensor. This was fine before the adjustable game size, but say you want to use a standard 52 card deck with 7 cards (a common recreational variant), then the max stock size goes up to 37. To handle this, we now simply reserve kDefaultNumCards (52) slots for stock size.

open_spiel/games/gin_rummy.h
```diff
@@ -35,53 +35,71 @@
...
+inline constexpr int kDefaultHandSize = 10;
+// TODO(jhtschultz) should kMaxStockSize be a tight upper bound for nonstandard
+// deck size?
 inline constexpr int kMaxStockSize = 31;  // Stock size when play begins
 inline constexpr int kWallStockSize = 2;
 inline constexpr int kDefaultKnockCard = 10;
 inline constexpr int kDefaultGinBonus = 25;
 inline constexpr int kDefaultUndercutBonus = 25;
+// Action constants
 inline constexpr int kDrawUpcardAction = 52;
```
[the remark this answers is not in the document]
jhtschultz: That's right. Added a comment here to make the action space layout clearer.

open_spiel/games/gin_rummy.h
```diff
@@ -176,14 +203,21 @@ class GinRummyState : public State {
...
 class GinRummyGame : public Game {
  public:
   explicit GinRummyGame(const GameParameters& params);
 
+  // TODO(Michal) Do we want the number of actions to change depending on game
+  // size params? Keeping it fixed means there will be unused action ints for
+  // small versions of the game, but there won't be many unused actions (< 52)
+  // and fixed action values for drawing a card, knocking, etc... is clearer
+  // and less likely to produce bugs.
   int NumDistinctActions() const override { return kNumDistinctActions; }
```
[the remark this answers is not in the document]
jhtschultz: Completely agree. Also see comment below...

open_spiel/games/gin_rummy.h
```diff
@@ -195,21 +229,35 @@ class GinRummyGame : public Game {
   std::unique_ptr<State> NewInitialState() const override {
     return std::unique_ptr<State>(
         new GinRummyState(shared_from_this(), oklahoma_, knock_card_,
-                          gin_bonus_, undercut_bonus_));
+                          gin_bonus_, undercut_bonus_, num_ranks_, num_suits_,
+                          hand_size_));
   }
+  // TODO(Michal) Do we want the obs tensor to change shape depending on game
```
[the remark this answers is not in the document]
jhtschultz: In accordance with this philosophy I made a change to the observation tensor. Previously, if you had a clairvoyant observer who could see both hands, the observation size increased by 52 to accommodate the second hand. Now, the observation tensor includes room for both hands in all cases. This makes the observation tensor size fixed under all game size variations and observer perspectives.

jhtschultz:
> Is it important to have melds of size 2? I think having a deck with just 2 card suits and 4 ranks should be interesting enough, so the meld size wouldn't need to change.

That's a good point. The goal is to have a simple version of the game to work with, not necessarily explore the entire space of small gin variants.

jhtschultz:
Added a simple gin bot which is quite useful for testing and to benchmark playing strength. I've played through a million games without any errors, so that should give us some amount of confidence that there are no major bugs. I made some minor cosmetic tweaks since running a lengthy test, so I'm going to run it again to be safe.

open_spiel/bots/gin_rummy/simple_bot.cc
```diff
@@ -0,0 +1,217 @@
...
+namespace open_spiel {
+namespace gin_rummy {
+
+void SimpleBot::Restart() {
+  knocked_ = false;
+  next_actions_ = {};
+}
+
+Action SimpleBot::Step(const State& state) {
+  std::vector<float> observation;
```
michalsustr, line 35: Why not use Observer? Then you don't need to update offsets (as much at least)

open_spiel/bots/gin_rummy/simple_bot.cc
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }  // Completed decoding observation.
+
+  auto legal_actions = state.LegalActions(player_id_);
+  // Next actions must be legal, in order from back to front.
+  if (next_actions_.size() > 0) {
+    Action action = next_actions_.back();
+    if (std::find(legal_actions.begin(), legal_actions.end(), action) ==
+        legal_actions.end()) {
+      std::cout << "Game state:" << std::endl;
```
michalsustr, line 91: cerr

open_spiel/bots/gin_rummy/simple_bot.cc
```diff
@@ -0,0 +1,217 @@
...
+    }
+    Action action = next_actions_.back();
+    SPIEL_CHECK_TRUE(std::find(legal_actions.begin(),
+        legal_actions.end(), action) != legal_actions.end());
+    next_actions_.pop_back();
+    return action;
+  } else if (!upcard.has_value()) {
+    // MoveType kDiscard
+    if (hand.size() != hand_size_ + 1) {
+      std::cout << "Game state:" << std::endl;
```
michalsustr, line 128: cerr

michalsustr:
Awesome, thanks @jhtschultz ! I think the playthroughs need to updated for the CI tests.

open_spiel/bots/gin_rummy/simple_bot.cc
jhtschultz: Ah, I built the first version of this bot a while ago and just didn’t switch it over to use Observer. How does this help us update offset less often though? Don’t we still get a tensor that we have to iterate over? And it's only through the python interface that we can access pieces by name right?

michalsustr: Yeah good point, I forgot it's not in C++ in master yet. Nevermind then!

open_spiel/bots/gin_rummy/simple_bot.cc
jhtschultz: Fixed

open_spiel/games/gin_rummy.h
```diff
@@ -35,53 +35,71 @@
...
 inline constexpr int kObservationTensorSize =
-    kNumPlayers          // Player turn
-    + kDefaultKnockCard  // Knock card
-    + kNumCards          // Player hand
-    + kNumCards          // Upcard
-    + kNumCards          // Discard pile
-    + kMaxStockSize      // Stock size
-    + kNumMeldActions;   // Opponent's layed melds
+    kNumPlayers             // Player turn
+    + kDefaultKnockCard     // Knock card
```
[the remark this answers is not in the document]
jhtschultz: Oops I thought I replied to this earlier! This is here because the observation tensor size is fixed and I had seen other games define a fixed tensor size here (backgammon). We could just as easily run this calculation inside of ObservationTensorShape(). Your call.

jhtschultz:
Made some minor tweaks and I'm finishing up a bot example which I'll add tomorrow. I was saving the playthroughs for last so we wouldn't have to redo them every time we made a change to the layout. I'll update those and we can do a final review.

open_spiel/games/gin_rummy.cc
jhtschultz: I removed the commented out code and followed the implementation in dark chess. If we encounter a use case for the observer registration I'll be happy to add it in.

jhtschultz:
Added the SimpleBot example. It lets us see how much of an edge SimpleBot has against a random bot — SimpleBot wins > 99% of the time. We can log the full gameplay for inspection, or output only game histories which can be used for supervised learning. This is a pretty solid test of the obs tensor and game implementation, and I'm continuing to run simulations.

Question: should SimpleBot be more specifically named SimpleGinRummyBot?

michalsustr:
> should SimpleBot be more specifically named SimpleGinRummyBot?

I'm in favor of `SimpleGinRummyBot`

> I was saving the playthroughs for last so we wouldn't have to redo them every time we made a change to the layout.

I think the code looks good, can you please save the final playthroughs so it can be merged? Thank you! :)

jhtschultz:
Should be good to go! Thanks for all the helpful feedback :)

lanctot:
Ok will import tomorrow morning. Thanks @michalsustr and @jhtschultz !

PR #560, opened 2021-04-14
Move libtorch download URL to global variables for friendlier setup.

michalsustr:
Thanks to @branislavhesko for pointing me out the local runfile installer!


lanctot:
Very nice!! Thanks your continued improvements to OpenSpiel, this is great.

PR #563, opened 2021-04-19
Make unnecessary Python dependencies optional

lanctot:


lanctot:
Now, _this_ is what I call "keeping it light":

```
pip >= 20.0.2
attrs >= 19.3.0
absl-py >= 0.10.0
six >= 1.15.0
```

😃

michalsustr:
Awesome!!

open_spiel/python/CMakeLists.txt
```diff
@@ -8,6 +8,56 @@ if(Python3_VERSION VERSION_LESS "3.6.0")
     "Python found ${Python3_VERSION} < 3.6.0")
 endif()
 
+# Detext the Python ML frameworks.
```
michalsustr, line 11: Detect.

open_spiel/scripts/python_extra_deps.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
michalsustr, line 1: Have you considered specifying min versions rather than equality? This should make the optional dependencies easier to use.

requirements.txt
```diff
@@ -1,36 +1,12 @@
 # The single source of truth for the OpenSpiel pip dependencies.
```
michalsustr, line 1: Not a single source of truth anymore.

open_spiel/scripts/python_extra_deps.sh
michalsustr: I guess the reasoning in requirements.txt still applies? Maybe it would appropriate to move it here?

open_spiel/scripts/global_variables.sh
```diff
@@ -101,3 +101,11 @@ export BUILD_WITH_ORTOOLS="${BUILD_WITH_ORTOOLS:-"OFF"}"
 # https://github.com/google/or-tools/releases/download/v8.0/or-tools_debian-10_v8.0.8283.tar.gz
 # https://github.com/google/or-tools/releases/download/v8.0/or-tools_MacOsX-10.15.7_v8.0.8283.tar.gz
 export BUILD_WITH_ORTOOLS_DOWNLOAD_URL="${BUILD_WITH_ORTOOLS_DOWNLOAD_URL:-"https://github.com/google/or-tools/releases/download/v8.0/or-tools_ubuntu-18.04_v8.0.8283.tar.gz"}"
+
+# Used to determine whether to include the Python ML frameworks in the tests.
```
michalsustr, line 105: If someone uses a fresh python install and wants to, I don't know, run AZ MCTS with pytorch, these won't be installed. I think it might be a good idea to update the documentation to mention this (something like "put ON instead of AUTO to install these extra deps")

open_spiel/scripts/python_extra_deps.sh
lanctot: I have changed requirements.txt to be lower bounds on the version. These are now entirely optional in the sense that if the user wants to use an algorithm or tool that depends on them, it is tgeir responsibility to install it separately.  These specific versions are never installed except on github actions CI for testing purposes, do they exist as recommendations to the user (not requirements)

requirements.txt
lanctot: Same reason as above, the packages in python_extra_deps are truly no longer official dependencies. The idea is that official dependencies in this file reflects only what is necessary to run the core python API.

open_spiel/scripts/global_variables.sh
lanctot: Thanks I will update the docs. However, as stated above, even setting these to ON will not install the frameworks. The user has to do that now.

open_spiel/scripts/python_extra_deps.sh
michalsustr: Makes sense! Maybe exactly this comment could be added to this file.

open_spiel/scripts/global_variables.sh
michalsustr: Ah, right... the ON install is done in travis script, I thought that was install.sh, my bad.

open_spiel/scripts/python_extra_deps.sh
lanctot: Yeah agree, this conversation made me realize this has to be better documented. Will do. Thanks!

lanctot: I probably will rephrase "single source of truth" because that feels too strong regardless.

requirements.txt
lanctot: I probably will rephrase "single source of truth" anyway because that feels too strong now.

open_spiel/python/CMakeLists.txt
lanctot: Fixed in most recent commit, thanks!

PR #564, opened 2021-04-19
Ensure move_number_ is decremented by UndoAction().

jhtschultz:
https://github.com/deepmind/open_spiel/blob/df657bccc51c9fab3e3eec941abaf1a1c6139d84/open_spiel/spiel.h#L546-L548

`--move_number_` is not called in most game implementations of UndoAction(). This PR tests for this by adding a line of code to TestUndo() in basic_tests.cc, and addresses the issue by adding `--move_number_` to every game that doesn't already implement it (except efg_game, which is a special case).

It's worth noting that not every game that implements UndoAction() tests it. We could add RandomSimTestWithUndo() to these game tests. It's slow test, but running even just a single simulation is a decent test of the UndoAction function. If there's a desire to add this test I'll gladly update the PR. Cheers!

### DIFF
--- a/open_spiel/games/backgammon.cc
+++ b/open_spiel/games/backgammon.cc
@@ -510,6 +510,7 @@ void BackgammonState::UndoAction(int player, Action action) {
   }
   turn_history_info_.pop_back();
   history_.pop_back();
+  --move_number_;
 }
 
 Action BackgammonState::TranslateAction(int from1, int from2,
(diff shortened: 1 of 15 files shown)


lanctot:
> It's worth noting that not every game that implements UndoAction() tests it. We could add RandomSimTestWithUndo() to these game tests. It's slow test, but running even just a single simulation is a decent test of the UndoAction function. If there's a desire to add this test I'll gladly update the PR. Cheers!

I agree.. but I only saw this after we imported the PR. I will do a github update tomorrow morning; if you're still motivated to do this as a separate PR I'd gladly take you up on the offer.

PR #566, opened 2021-04-21
Add UndoAction() tests.

jhtschultz:
Follow up to https://github.com/deepmind/open_spiel/pull/564. Adds RandomSimTestWithUndo() for all games which implement UndoAction(). This is potentially a slow test so we only run a single simulation, but this should still suffice as a basic test of UndoAction().

UPDATE: Apparently it's a worthwhile test because multiple games failed! For different reasons too. Opening this PR so there's a record of the error messages.


lanctot:
Nice catches! Yeah it's an amazingly tricky function to get right. I fixed a bug in breakthrough's undo a few weeks back.. was present for over two years ;) thanks for doing this!

lanctot:
@jhtschultz can you resolve the conflicts and merge with master? Might be nice to get this in and the fixes to the undo (or removals) in time for the release.

jhtschultz:
Merged. Looks like there are 5 games that are failing: catch, cliff walking, dark hex, deep sea, and tiny bridge.

PR #567, opened 2021-04-21
X-Y board size options added for game Hex

BedirT:
I have modified the board_size functionality to be able to give row and column sizes separately. This was needed for a certain type of research questions.


elkhrt:
Thanks for the contribution - integrating it now.
I had to change it a bit to make sure it doesn't break existing users.

PR #583, opened 2021-05-10
Add cibuildwheel workflow for building pip wheels, and pip tests

lanctot:


.github/workflows/wheels.yml
```diff
@@ -0,0 +1,74 @@
...
+          python --version
+          chmod +x install.sh
+          ./install.sh
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade setuptools
+          python -m pip install --upgrade -r requirements.txt -q
+          source ./open_spiel/scripts/python_extra_deps.sh
+          python -m pip install --upgrade $OPEN_SPIEL_PYTHON_JAX_DEPS $OPEN_SPIEL_PYTHON_PYTORCH_DEPS $OPEN_SPIEL_PYTHON_TENSORFLOW_DEPS $OPEN_SPIEL_PYTHON_MISC_DEPS
+          python -m pip install twine build
+          python -m pip install cibuildwheel==1.10.0
```
henryiii, line 58:
```suggestion
          python -m pip install cibuildwheel==1.12.0
```
?

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Build sdist
+        run: |
+          python -m build -s
+          twine check dist/*.tar.gz
+          
+      - name: Build bdist_wheel
+        run: python -m cibuildwheel --output-dir wheelhouse
+
+      - name: Install and test bdist_wheel
+        run: ./open_spiel/scripts/test_wheel.sh
```
henryiii, line 68: FYI, you can tell cibuildwheel to do the testing via variables too; if you do that, you can remove most of the install step.

.github/workflows/wheels.yml
lanctot: Ah yes, thanks!

.github/workflows/wheels.yml
lanctot: Thanks for the suggestions!

Do you mean it can do the testing from within the same virtual environments that have built the wheels? That would be ideal, yes! Do you have an example you can point me to on how to run arbitrary commands within them?

henryiii: https://cibuildwheel.readthedocs.io/en/stable/options/#test-command - multiple commands should be joined with && on all platforms.

lanctot: That's awesome, thanks for the pointer! I will take a look once I figure out how to fix the current build issue.

henryiii: FYI, cibuildwheel 2.0 will add a config mode which is really nice for adding things like this (see https://github.com/scikit-build/cmake-python-distributions/pull/139/commits/bc80ddb3bc13f1d28d9a1e28b36ba9ce6f92cba9 for a great example of it). 2.0.0a4 is fully usable as long as you don't activate the new "build" backend instead of pip (ironically broken due to a `pypa/pip` bug, not a `pypa/build` bug).

henryiii:
Ohh, curious, looks like the new Path caster introduced in https://github.com/pybind/pybind11/pull/2730 is breaking support for anything older than macOS 10.15? Guessing you are using master pybind11?

henryiii:
Yep, bug in master pybind11. Either revert to using an older commit, or wait till a fix is made (hopefully soon).

lanctot:
Actually in OpenSpiel we are using the smart_holder branch, would it still explain it?

henryiii:
Pretty sure @rwgk's team keeps smart_holder branch up to date.

rwgk:
> Actually in OpenSpiel we are using the smart_holder branch, would it still explain it? Edit: tagging @rwgk who we've been working with. This is not urgent so will wait, thanks for your help!

Yes, I'm generally merging PRs against master very quickly into the smart_holder branch (I did so this time).

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+          python --version
+          chmod +x install.sh
+          ./install.sh
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade setuptools
+          python -m pip install --upgrade -r requirements.txt -q
+          source ./open_spiel/scripts/python_extra_deps.sh
+          python -m pip install --upgrade $OPEN_SPIEL_PYTHON_JAX_DEPS $OPEN_SPIEL_PYTHON_PYTORCH_DEPS $OPEN_SPIEL_PYTHON_TENSORFLOW_DEPS $OPEN_SPIEL_PYTHON_MISC_DEPS
+          python -m pip install twine build
+          python -m pip install cibuildwheel==1.12.0
```
henryiii, line 62: Why not 2.0.1? Also, you can use the action if you'd like instead of manually installing it - that's easier to keep up to date (say with dependabot).

If you use version 2, you can specify your settings (most of the CIBW_* variables) in your pyproject.toml, which also makes it much easier to run locally.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Build sdist
+        run: |
+          python -m build -s
```
henryiii, line 65:
```suggestion
          pipx run build --sdist
```

You don't have to install build, you can just run it with pipx. And long-options are usually preferred when you aren't typing them interactively. Up to you, though.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+      OPEN_SPIEL_BUILD_WITH_ACPC: ON
+      OPEN_SPIEL_BUILD_WITH_HANABI: ON
+      OS_PYTHON_VERSION: 3.9
+      # While testing, try specific versions
+      # CIBW_BUILD: cp36-* cp37-* cp38-* cp39-*
+      CIBW_PROJECT_REQUIRES_PYTHON: ">=3.6"
+      CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+      CIBW_BUILD: ${{ matrix.CIBW_BUILD }}
+      CIBW_SKIP: cp27-* pp*
+      CIBW_BEFORE_BUILD: python -m pip install --upgrade cmake
```
henryiii, line 37: These should be specified in pyproject.toml normally, then everyone who builds this from an SDist will benefit. Though I see you are running `.install.sh` and `./open_spiel/scripts/python_extra_deps.sh` - at least for Python, the environment is isolated, you will not be using things installed via pip inside the wheel build process. It's not even running on the same version of Python that you are running cibuildwheel with.

Now it might be picking up build produces from the environment, but those must be done inside `CIBW_BEFORE_ALL` or `tool.cibuildwheel.before-all` (TOML). For example, on Linux, these are built on the host rather than in the docker container if you run them in a step.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+          CIBW_ENVIRONMENT: "CXX=$(which g++) OPEN_SPIEL_BUILDING_WHEEL='ON' OPEN_SPIEL_BUILD_WITH_ACPC='ON' OPEN_SPIEL_BUILD_WITH_HANABI='ON'"
+          CIBW_BUILD: cp36-manylinux_x86_64 cp37-manylinux_x86_64 cp38-manylinux_x86_64 cp39-manylinux_x86_64
+        - os: macOS-10.15
+          CIBW_ENVIRONMENT: "OPEN_SPIEL_BUILDING_WHEEL='ON' OPEN_SPIEL_BUILD_WITH_ACPC='ON' OPEN_SPIEL_BUILD_WITH_HANABI='ON'"
+          CIBW_BUILD: cp36-macosx_x86_64 cp37-macosx_x86_64 cp38-macosx_x86_64 cp39-macosx_x86_64
+    env:
+      OPEN_SPIEL_BUILDING_WHEEL: ON
+      OPEN_SPIEL_BUILD_WITH_ACPC: ON
+      OPEN_SPIEL_BUILD_WITH_HANABI: ON
+      OS_PYTHON_VERSION: 3.9
```
henryiii, line 30: The outer environment shouldn't matter, this should only be in CIBW_ENVIRONMENT / `tool.cibuildwheel.environment`.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+      # While testing, try specific versions
+      # CIBW_BUILD: cp36-* cp37-* cp38-* cp39-*
+      CIBW_PROJECT_REQUIRES_PYTHON: ">=3.6"
```
henryiii, line 33: This should be set in setup.cfg in the standard python_requires setting, cibuildwheel will read it from there. And cibw 2.0 only supports 3.6+ anyway.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+          which g++
+          g++ --version
+          python --version
+          chmod +x install.sh
+          ./install.sh
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade setuptools
+          python -m pip install --upgrade -r requirements.txt -q
+          source ./open_spiel/scripts/python_extra_deps.sh
+          python -m pip install --upgrade $OPEN_SPIEL_PYTHON_JAX_DEPS $OPEN_SPIEL_PYTHON_PYTORCH_DEPS $OPEN_SPIEL_PYTHON_TENSORFLOW_DEPS $OPEN_SPIEL_PYTHON_MISC_DEPS
```
henryiii, line 60: None of this _should_ affect the cibuildwheel run. If it does, that's probably a bug - you don't want stuff build with this version of GCC, etc, inside manylinux2014, that has it's own compiler, etc. If you need it, put it in the before build variable. Join with `&&` in environment variable mode or use a list in TOML for multiple commands.

.github/workflows/wheels.yml
lanctot: Ok, thanks again for all the great feedback. 

I have now run a basic test suite within the Docker image by the `CIBW_TEST_COMMAND` and larger tests externally on just the two platforms. 

Our release has been delayed for some time waiting only on this bdist wheel, and there are a few other reasons that 
I need to push it forward. Unfortunately, I won't have the time to address everything since I have quite limited time to spend on this. So some of these points will have to be deferred (most notably a move to the TOML -- I am unable to invest the time into it right now). I appreciate the help nonetheless, and will look into it for future releases.

.github/workflows/wheels.yml
lanctot: > Why not 2.0.1? 

Because I copied this mostly from examples I found and have not kept up with the versions/releases :) 

> Also, you can use the action if you'd like instead of manually installing it - that's easier to keep up to date (say with dependabot). Or you can use pipx, but the syntax for pinning is (currently) a little ugly.
> If you use version 2, you can specify your settings (most of the CIBW_* variables) in your pyproject.toml, which also makes it much easier to run locally.

Ok good to know. I'll look into it for the next release....... or solicit a contribution from an enthusiatic third party (nudge nudge wink wink, I'll say no more! :))

.github/workflows/wheels.yml
lanctot: Applied this suggestion in one of my latest commits, thanks!

.github/workflows/wheels.yml
lanctot: Ah, in this case it does, actually. The `install.sh` needs these to clone the dependencies.

.github/workflows/wheels.yml
lanctot: Done, thanks!

.github/workflows/wheels.yml
lanctot: I have added comments to explain why these are still needed. Short version: I have two set of tests now: basic and full. 

The 'basic' tests are run within the docker image via `CIBW_TEST_COMMAND` as per your recommendation. These have to be a subset of all the tests because unfortunately we need >= 3.7 for some of the machine learning library dependencies (but I want the basic OpenSpiel wheels to still be 3.6-compliant).

The 'full' tests are then run afterward (outside the docker images), by installing the wheel and running the full suite of tests (including all the Tensorflow, PyTorch, and JAX tests). So these full tests require the extra installs.

PR #584, opened 2021-05-10
Add default values for GameInfo in python through the binding.

TheoCabannes:
This enables defining a python non constant-sum-utility games without defining utility_sum.

### DIFF
--- a/open_spiel/python/pybind11/pyspiel.cc
+++ b/open_spiel/python/pybind11/pyspiel.cc
@@ -261,11 +261,10 @@ PYBIND11_MODULE(pyspiel, m) {
   py::class_<GameInfo> game_info(m, "GameInfo");
   game_info
       .def(py::init<int, int, int, double, double, double, int>(),
-           py::arg("num_distinct_actions") = 0,
-           py::arg("max_chance_outcomes") = 0,
-           py::arg("num_players") = 0, py::arg("min_utility") = 0,
-           py::arg("max_utility") = 0, py::arg("utility_sum") = 0,
-           py::arg("max_game_length") = 0)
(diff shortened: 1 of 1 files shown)


lanctot:
Thank @TheoCabannes,

I am trying to see how this enables simultaneous move games without explicittly defining player_sum? Are you working on one? If so, would great to have an example in the repos alongside the others, would you be willing to contribute it?

Also maybe these defaults should be invalid values as defaults so that we can put asserts somewhere to ensure that the necessary ones are in the valid ranges.

Tagging @elkhrt who wrote the revamp of the python games, would like his opinion on this.

TheoCabannes:
@lanctot my description was a little bit wrong, sorry about that. I have changed it.
This enables defining non constant-sum-utility games without defining utility_sum. In non constant-sum-utility games, utility_sum does not exists, and I think that it should have a default values.

I used the virtual functions in Game to define the default values: https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/spiel.h#L742

Screen Shot 2021-05-10 at 11 03 35 AM
Screen Shot 2021-05-10 at 11 03 44 AM

TheoCabannes:
I am working on one simultaneous game without fix utility_sum. I am happy to contribute when my implementation of the game is ready.

elkhrt (commented):
I'm fine with defaulting utility_sum to zero, but please remove the other defaults.

TheoCabannes:
I added a commit to only change utility_sum default value

lanctot:
> I added a commit to only change utility_sum default value

Thanks. Importing this now.

PR #585, opened 2021-05-10
Update description of max_game_length in spiel.h.

TheoCabannes:
Use description of MaxGameLength().

Currently the description of MaxGameLength() is more exhaustive than the description of max_game_length.
This makes the code a little bit hard the understand. I wonder if it would make sense to only describe each variable once and refer to the description in the other docstrings.


lanctot:
Great, thanks!

PR #586, opened 2021-05-11
Fix str concatenation issue

TheoCabannes:
Game in rl_environment can either be a string or a pyspiel game. In case it is a pyspiel game then it cannot be concatenate with string.

Let me know if small PR like this one are annoying. In this case I stop making small ones, or make a big one with several small fixes.


lanctot:
Not annoying at all. Thanks for the fixes!

lanctot:
It is better that each fix is self-contained, makes them much easier to review and discuss separately.

PR #593, opened 2021-05-13
Fix cmake equality bug

ssokota:
cmake EQUAL only returns true if inputs are equal and are valid numbers

### DIFF
--- a/open_spiel/CMakeLists.txt
+++ b/open_spiel/CMakeLists.txt
@@ -34,20 +34,20 @@ if(NOT BUILD_TYPE)
 endif()
 message("${BoldYellow}Current build type is: ${BUILD_TYPE}${ColourReset}")
 
-if(${BUILD_TYPE} EQUAL "Debug")
+if(${BUILD_TYPE} STREQUAL "Debug")
   # Basic build for debugging (default).
   # -Og enables optimizations that do not interfere with debugging.
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -Og")
 endif()
 
-if(${BUILD_TYPE} EQUAL "Testing")
+if(${BUILD_TYPE} STREQUAL "Testing")
   # A build used for running tests: keep all runtime checks (assert,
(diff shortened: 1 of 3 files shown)


ssokota:
@googlebot I signed it!

lanctot:
Thanks!

lanctot:
Hmm, seems to be failing tests (nit compiling in abseil). Not sure why, I guess you didn't see these?

lanctot:
Looks to me like `-O2` triggers a bug in abseil? That'd be quite a surprise to me but I have no other explanation.

lanctot:
Confirmed: https://github.com/abseil/abseil-cpp/issues/832

We will have to increase our abseil version.

lanctot:
@ssokota, can you change the version tag on this line: https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/scripts/install.sh#L87

from `20200923.1` to `20200923.3`? I'd like to use this PR to trigger tests to be sure everything else still works with that update.

lanctot:
Ok looking good. Oddly thread_test now fails but only on Ubuntu 18.04.

lanctot:
I would like to get this in ASAP. For now, can you comment out that test for now, here: https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/utils/CMakeLists.txt#L80

(and add a comment that it fails on Ubuntu 18.04 using the Testing build type). We can fix it and re-enable later.

lanctot:
Now the Julia test fails. Any ideas, @findmyway ?

findmyway:
That seems to be the old famous glibc mismatch error. You can disable it first. Let me test it locally and see how to fix it gracefully.

lanctot:
I think these are unrelated to @ssokota 's change. I had to make them on master too for the sync to github today, and I haven't even imported the change yet. 

The Julia test is now failing on master too: https://github.com/deepmind/open_spiel/actions/runs/849347899

Something must have changed on the machines that runt he tests on github. I'm temporarily disabling the julia test until we can fix it.

PR #596, opened 2021-05-17
Fix GLIBCXX_3.4.29 not found error in the Julia wrapper

findmyway:
It seems this version is not shipped with Julia@v1.6.1. The only way left is to make a symlink or remove libstdc++.so.6 from the prebuilt julia binary.

- [ ] Check whether making a symlink works or not.

### DIFF
--- a/open_spiel/julia/CMakeLists.txt
+++ b/open_spiel/julia/CMakeLists.txt
@@ -14,11 +14,5 @@ install(TARGETS spieljl
     LIBRARY DESTINATION lib
 )
 
-# This is failing now too (in addition to thread_test). Temporarily disabling.
-# See https://github.com/deepmind/open_spiel/pull/593 and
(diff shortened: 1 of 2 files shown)


lanctot:
Can you pull changes from master and try again?

lanctot:
Hi @findmyway,

I had to disable the julia test this morning: https://github.com/deepmind/open_spiel/commit/cb327b1d662e3552416c9e10af61cf994ae6a746

Can you add it back in this branch?

findmyway:
Yes, sure. I'm working on it.

lanctot:
IIRC we needed a copy command previously, see https://github.com/deepmind/open_spiel/commit/885dd00c1ac0839f641a8e9c2c68bee5e7f16f4d

lanctot:
I've disabled the JAX NFSP test, importing this now, thanks for the fix!

PR #598, opened 2021-05-17
Update developper guide to be consistent with https://github.com/deepmind/open_spiel/commit/7b659bbdf54e63da4eea98feebee3ec64f7c449e#diff-5fb49e5993f125efe1e959a56b6ee5349f25f260359ac48268934a31bd7a8555

TheoCabannes:


lanctot:
Can you try pull changes from master and committing again?

TheoCabannes:
I think that my branch is up-to-date. It is only changing one md file

lanctot:
> I think that my branch is up-to-date. It is only changing one md file

I asked because we had some issues with master this morning and it took a few commits to fix them (so you may have branched in between these). 

Oddly the failure was in a JAX nfsp test that I had not seen before. I reran them and it's fine, but I wonder now if we have a nondeterministic failure somewhere. We'll probably catch it at a future point.

Thanks for the PR!

PR #589, opened 2021-05-12
Test of simultaneous python game and conversion to turn based with dummy dynamic routing game

TheoCabannes:
The goal of this PR is:
- to create a test for the following issues:
   - https://github.com/deepmind/open_spiel/issues/588
   - https://github.com/deepmind/open_spiel/pull/587
   - https://github.com/deepmind/open_spiel/pull/584
- to add example of simultaneous game solved with conversion to sequential games with imperfect information with python.

Thank you a lot for considering my contribution.
The game is a dummy version of [A mean field route choice game](http://www.professeurs.polymtl.ca/jerome.le-ny/docs/proceedings/2018_CDC_MFGrouteChoice_Salhab.pdf) by R. Salhab, J. Le Ny and R. P. Malhamé, 2018 IEEE CDC.
This implementation targets to implement a "more realistic" dynamic routing game than this paper (the notion of cost and travel time in the paper is not linked to the actual graph traversal time). This implementation also targets to solve the game with an optimal control approach using mean field games. Hopefully it can become a part of open spiel.

PR #599, opened 2021-05-17
Need some help on solving dummy dynamic routing game implementation with CFR

TheoCabannes:
I am creating this PR to get some help on my current work on the implementation of a dynamic routing game within open spiel. This PR is an updated version of https://github.com/deepmind/open_spiel/pull/589.

Currently I have 2 errors and I need some help:
1. Some tests in GameSimTest are failing because of unknown parameters 'players'.
Screen Shot 2021-05-17 at 2 28 20 PM

2. The sequential game ask the legal actions for player 3 while they are only 2 players: 
Screen Shot 2021-05-17 at 7 06 43 PM

I also have other issues but they might be consequences of the issues above (one python issue: https://github.com/TheoCabannes/open_spiel/issues/10 and one issue with MCCFR with external or internal sampling).

The tests are not passing but this is also why I am doing this PR.


TheoCabannes:
I have updated the PR description because I was able to solve some issues with the two previous commits.

elkhrt:
Thanks for helping us with these rough edges. Python games are still new, and you're the first person to use them for real. Fixes below - please let me know how you get on.

1. Parameters to games are specified in a dict. You need the following changes:
```
    parameter_specification={"players": pyspiel.GameParameter(3)})
    # ...
    num_players=params.get("players", 3),
    # ...
    super().__init__(_GAME_TYPE, game_info, params)
```
Plus remove the constructor argument. We could certainly make this a bit cleaner.

2. There's a bug which causes Python games to be prematurely garbage-collected. Please replicate the workaround at https://github.com/deepmind/open_spiel/blob/cb327b1d662e3552416c9e10af61cf994ae6a746/open_spiel/python/games/iterated_prisoners_dilemma.py#L65-L77 (including assigning the TODO to author11).

Pausing now since I see you've made some changes.

TheoCabannes:
Thank you a lot @elkhrt.
For the point 1, I do not fully understand the fix. Can you be a little bit more precise?
For the point 2, thank you I have done it [here](https://github.com/deepmind/open_spiel/pull/599/commits/7f785aaea5c46340d7e11b65e99a37dc8fe9f4c0).

I figure out that one of my mistake was due to the fact that **if no players have legal_actions then the game should be in a terminal state**. I was currently setting the game to be finish if the vector of actions passed to _apply_actions was a list of 0. But this should be done before calling _apply_actions as it is tested that the legal_actions (for the RolloutModeIncrementCurrentPlayer) are not empty [here](https://github.com/deepmind/open_spiel/blob/cb327b1d662e3552416c9e10af61cf994ae6a746/open_spiel/tests/basic_tests.cc#L397). This should be fixed on this game.
Locally I only have the mistake with CFR algorithm now. I will update the PR later to also show the errors with MCCFR algorithm.

I was not able to apply CFR to iterated_prisoners_dilemma neither. So I guess this might be normal.
I will find if the documentation is consistent with [this test](https://github.com/deepmind/open_spiel/blob/cb327b1d662e3552416c9e10af61cf994ae6a746/open_spiel/tests/basic_tests.cc#L394) when converting a simultaneous game to a turn based game.

elkhrt:
For (1), you are passing the num_players parameter as an argument to the constructor. That's not how it works. You need instead to get the `players` param from the `params` dict. I listed three lines that need to change. They are lines 61, 123, 129 in your most recent version.

TheoCabannes:
Thank you a lot @elkhrt. I have updated the PR. Now it should have only one error which is a test I have created.
This test is trying to apply CFR on the sequential version of the dynamic routing game.

Screen Shot 2021-05-18 at 12 05 17 AM

Can you please provide me some help in how to solve my game with any algorithm? @perolat told me to use CFR first, but it does not seem to be working.

elkhrt:
This is failing because you haven't defined observations for the states, which means they are all treated as being the same state, even though they have different legal actions (which is of course an error).
Simple fix is to implement `string_from` in `NetworkObserver`:
```
    def string_from(self, state, player):
        return state.history_str()
```
Eventually you might prefer something more readable, and to distinguish observations (imperfect recall) from information states (perfect recall).

TheoCabannes:
Thank you a lot @elkhrt. This solves my issues so I am converting this PR to draft and I will let you know if I find other issue.

TheoCabannes:
> For (1), you are passing the num_players parameter as an argument to the constructor. That's not how it works. You need instead to get the `players` param from the `params` dict. I listed three lines that need to change. They are lines 61, 123, 129 in your most recent version.

@elkhrt or @lanctot, is it possible to have heterogeneous players. I am working on a implementation of a dynamic routing game (not this dummy one with 2 links network) and I would like to have players (or vehicles) with different origin and destination. I do not see how I can do that with the params dict. Can you help me?

TheoCabannes:
Closing the PR as most of my questions have been solved. Thank you a lot! I will now implement travel time on a link using chance nodes.

TheoCabannes:
> 1. Parameters to games are specified in a dict. You need the following changes:
> 
> ```
>     parameter_specification={"players": pyspiel.GameParameter(3)})
>     # ...
>     num_players=params.get("players", 3),
>     # ...
>     super().__init__(_GAME_TYPE, game_info, params)
> ```
> 
> Plus remove the constructor argument. We could certainly make this a bit cleaner.

@elkhrt it seems that with https://github.com/deepmind/open_spiel/commit/88529098f8ddc24e0928fe491130241109ec128f `pyspiel.GameParameter` does not exists anymore:
Screen Shot 2021-05-24 at 5 05 04 PM
What should I do instead of using `pyspiel.GameParameter(3)` now?

Thank you for your help!

lanctot:
Hi @TheoCabannes, you should just use python values directly now. See this commit for details/examples of changes: https://github.com/deepmind/open_spiel/commit/38ed691a9535dd5501eb5e2f292515e5a8e53430.

Issue #588, opened 2021-05-12
Inconsistent documentation in case no possible actions


TheoCabannes:
In case a player has no possible actions, the documentation is not clear.

ApplyActions states that the player should have the action kInvalidAction: https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/spiel.h#L557.
Generate_playthrough.py seems to work when the player has kInvalidAction as action.

LegalActions seems to say that the player should have a empty list: https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/spiel.h#L229
spiel.cc returns an error in case kInvalidAction is used https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/spiel.cc#L317
This makes CFR not working on a game I am developing in my own fork: https://github.com/TheoCabannes/open_spiel/issues/7

Can you help me? Is what I am saying true?
Can CFR works with games where some players do not have valid actions?

lanctot:
Hi @TheoCabannes,

Sure we can help but I have to understand your use case first.

Is it a simultanrous move game? If so are you loading it as a sequential game using turn_based_simultaneous_game?

Sounds you are indeed hitting some uncovered use cases of our API.

TheoCabannes:
Indeed, I am creating a simultaneous move game and I am loading it as a sequential game using turn_based_simultaneous_game. I am also currently writing it with python.

I am currently hitting several issues, so I will probably create a small use case in order to enable debugging my current issues.
Hopefully, the small use case can later be converted in a test.

Thank you a lot for your help!

lanctot:
If a simultaneous game, are you saying you have nodes where some players have actions and others don't have any?

Can we clearly exclude the case that it is a player's turn and they have zero legal moves?

lanctot:
Absolutely we will definitely make sure it is properly handled and covered in a test afterward. Thanks a lot for your help in following up.

TheoCabannes:
Yes, in some nodes some players have actions and other not.
When converting into turn based it created nodes where there is no legal moves.
Currently, one fix is to use kInvalidAction in this case and comment the line https://github.com/deepmind/open_spiel/blob/53eeb6127578f91e88a6c0451983d10e3509446d/open_spiel/spiel.cc#L317

[...]

PR #601, opened 2021-05-17
Check if legal actions is empty for one player.

TheoCabannes:
This PR is done following the discussion on https://github.com/deepmind/open_spiel/issues/588#issuecomment-842404649.
I used a map in python instead of a for loop (as it was done in https://github.com/deepmind/open_spiel/commit/96bd3e1e65d2c36947b4e45a7033bf775b8bbbe7) after reading [this blog](https://www.geeksforgeeks.org/python-map-vs-for-loop/).

### DIFF
--- a/open_spiel/python/examples/example.py
+++ b/open_spiel/python/examples/example.py
@@ -83,9 +83,9 @@ def main(_):
       # Simultaneous node: sample actions for all players.
       random_choice = lambda a: np.random.choice(a) if a else [0]
       chosen_actions = [
-           random_choice(state.legal_actions(pid))
-           for pid in range(game.num_players())
-       ]
+          random_choice(state.legal_actions(pid))
+          for pid in range(game.num_players())
+      ]
       print("Chosen actions: ", [
           state.action_to_string(pid, action)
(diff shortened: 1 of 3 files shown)


open_spiel/python/examples/example.py
```diff
@@ -81,10 +81,11 @@ def main(_):
...
       # Simultaneous node: sample actions for all players.
-      chosen_actions = [
-          random.choice(state.legal_actions(pid))
-          for pid in range(game.num_players())
-      ]
+      def get_random_action(player):
+        legal_actions_player = state.legal_actions(player)
+        return (0 if not legal_actions_player else
+                random.choice(legal_actions_player))
+      chosen_actions = map(get_random_action, range(game.num_players())
```
elkhrt, line 88: The original version (list comprehension) is preferred to `map` in Python.
The blog you linked is about map vs for loop which is a different thing. List comprehensions are (almost) always a better choice than map.

You can just leave the original code and define:
```
random_choice = lambda a: np.random.choice(a) if a else [0]
```

TheoCabannes: Resolve comment by the last 2 commits.

PR #603, opened 2021-05-17
Add a test when all players have no legal actions.

TheoCabannes:
And update the documentation.
This is done after working on [this PR](https://github.com/deepmind/open_spiel/pull/599).

In implementation of simultaneous game, one should anticipate in apply_actions that no players will have legal actions in the next step in order to set the game to be terminated. This should be done because if no players have legal actions and the game is not terminated then in the sequential version of the simultaneous game, the code will try to get the action of the player num_player which is not a valid player. The test is checking that in the sequential version of the simultaneous game, num_player is not returned as a valid player when no player have valid actions.

I am not sure that my description is clear, but hopefully it should help.


lanctot:
Great, thanks!

We will try to import and sync more frequently during this high activity period to help with the turn-around.

PR #605, opened 2021-05-18
Use legal_chance_outcomes bindings to simplify _legal_actions in python.

TheoCabannes:
It seems to be easier to only have to implement chance_outcomes and not _legal_actions when chance nodes are used in python.
The code in C++ seems to be done with this philosophy but the python binding does not enable this.
This PR adds the python binding to enable using legal_change_outcomes() and change the two python examples.

### DIFF
--- a/open_spiel/python/pybind11/pyspiel.cc
+++ b/open_spiel/python/pybind11/pyspiel.cc
@@ -296,9 +296,7 @@ PYBIND11_MODULE(pyspiel, m) {
            (std::vector<int>(State::*)(int) const) & State::LegalActionsMask)
       .def("legal_actions_mask",
            (std::vector<int>(State::*)(void) const) & State::LegalActionsMask)
-      .def("legal_chance_outcomes",
-           (std::vector<open_spiel::Action>(State::*)(void) const) &
-               State::LegalChanceOutcomes)
+      .def("legal_chance_outcomes", &State::LegalChanceOutcomes)
       .def("action_to_string", (std::string(State::*)(Player, Action) const) &
                                    State::ActionToString)
       .def("action_to_string",
--- a/open_spiel/python/pybind11/pyspiel.cc
+++ b/open_spiel/python/pybind11/pyspiel.cc
@@ -296,7 +296,6 @@ PYBIND11_MODULE(pyspiel, m) {
            (std::vector<int>(State::*)(int) const) & State::LegalActionsMask)
       .def("legal_actions_mask",
            (std::vector<int>(State::*)(void) const) & State::LegalActionsMask)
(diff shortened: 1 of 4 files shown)


lanctot:
Github Actions seems to be having trouble today.. failing with inexplicable "internal errors". If you manage to find a log for your test runs, please point me to it.

I've had to restart the jobs twice already after the sync to master a few hours ago. Until we can get a log I'm guessing it's a problem on their end.

TheoCabannes:
This one seems to be working: https://github.com/deepmind/open_spiel/pull/605/checks?check_run_id=2610070209

lanctot:
> This one seems to be working: https://github.com/deepmind/open_spiel/pull/605/checks?check_run_id=2610070209

Yeah it's the same with the runs running on master from the github sync. Some of them work, but the ones that don't fail with a "Github Actions internal error" (no log of the job itself), so it's hard to figure out what's going on. I noticed the same with the ones failing in this PR (can't see the log for the ones that failed).

lanctot:
(Normally it posts a full log so you can see what happened)

lanctot:
Hi @TheoCabannes, GitHub Actions seems ok now: if you pull and merge changes from master (or just disable the julia_test in julia/CMakeLists.txt) and commit, it should rerun the tests. I can import it once we see it passing the tests.

TheoCabannes:
@lanctot I created two commits that cancel each other in order to relaunch the tests.

TheoCabannes:
Tests are passing.

open_spiel/python/pybind11/pyspiel.cc
```diff
@@ -296,6 +296,9 @@ PYBIND11_MODULE(pyspiel, m) {
            (std::vector<int>(State::*)(int) const) & State::LegalActionsMask)
       .def("legal_actions_mask",
            (std::vector<int>(State::*)(void) const) & State::LegalActionsMask)
+      .def("legal_chance_outcomes",
+           (std::vector<open_spiel::Action>(State::*)(void) const) &
```
elkhrt, line 300: This cast is only needed in the case that there's an overloaded method name. It can be omitted here.

TheoCabannes: Thanks! I have updated the PR

elkhrt:
This is a nice simplification, but we can make it even simpler - change PyState::LegalActions to call LegalChanceOutcomes at chance nodes. That means the Python-side _legal_actions will only be called for real players, so the chance logic can be removed from it entirely.
(And also means we don't need to expose the legal_chance_outcomes function)

TheoCabannes:
Thank you a lot. I have implemented your suggestion. Tests are passing on my local machine.

PR #608, opened 2021-05-19
Fix #606

### LINKED ISSUE #606: Julia test failing
Hi @findmyway, seems like there are still some problems with the Julia test so I've disabled it again. See: https://github.com/deepmind/open_spiel/runs/2613878461?check_suite_focus=true.

I _think_ currently the line with the copy of the libstdc++.so is only triggered when julia is installed manually. On Github Actions, Julia 1.6.1 is installed in /usr/local/julia. So I wonder if there's a similar patch for this issue in that case?

findmyway:
Fix #606 

I need to have a think of it first. The correct way is to find the `Julia_LIBRARY_DIR ` first like https://github.com/JuliaInterop/libcxxwrap-julia/blob/master/FindJulia.cmake#L81

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -75,10 +75,15 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: julia-actions/setup-julia@v1
+    - name: Ad-hoc fix
+      if: ${{ matrix.os == 'ubuntu-20.04' }}
+      run: |
+        # workaround for https://github.com/deepmind/open_spiel/issues/606
+        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 $(julia --startup-file=no -E 'using Libdl;print(abspath(joinpath(Libdl.dlpath("libjulia"), "..", "julia")))')
     - name: Install
       run: |
         pwd
-        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 $(julia -e 'print(joinpath(Sys.BINDIR, "..", "lib", "julia"))')
         chmod +x install.sh
         ./install.sh
     - name: Build and test
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -80,7 +80,7 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
         # workaround for https://github.com/deepmind/open_spiel/issues/606
-        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 $(julia --startup-file=no -E 'using Libdl;print(abspath(joinpath(Libdl.dlpath("libjulia"), "..", "julia")))')
+        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 /opt/hostedtoolcache/julia/1.6.1/x64/lib/julia 
     - name: Install
       run: |
         pwd
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -80,7 +80,7 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
         # workaround for https://github.com/deepmind/open_spiel/issues/606
-        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 /opt/hostedtoolcache/julia/1.6.1/x64/lib/julia 
+        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 $(julia --startup-file=no -e 'using Libdl;print(abspath(joinpath(Libdl.dlpath("libjulia"), "..", "julia")))')
     - name: Install
       run: |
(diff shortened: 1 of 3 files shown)


findmyway:
Hi @lanctot ,

To avoid polluting the `install.sh` with `sudo`. I simply make the copy in the github action workflow. I estimate it should be fixed in the next major release of Julia. Then we can remove this step.

lanctot:
Ok, but I think you'll need to uncomment the Julia test too :)

Here: https://github.com/deepmind/open_spiel/blob/1de866492591fb5fcd174665db6266f80ced7f98/open_spiel/julia/CMakeLists.txt#L17

findmyway:
Thanks for your remind, it seems ` /usr/lib/x86_64-linux-gnu/libstdc++.so.6` only contains glibcxx up to `GLIBCXX_3.4.28` 😢

lanctot:
Can you try moving that test with all the dependencies enabled to `ubuntu-21.04` in actions.yml?

findmyway:
It seems we need to wait for https://github.com/actions/virtual-environments/issues/3359#issue-886426749

lanctot:
Seems like github actions is acting up again, similar to yesterday. I will restart the job.

lanctot:
> It seems we need to wait for [actions/virtual-environments#3359 (comment)](https://github.com/actions/virtual-environments/issues/3359#issue-886426749)

Hmm.. are you sure this will update libc++? We use clang++/libc++ rather than g++/libstdc++.

I wonder if a newer version of clang/libc++ would work that we can install (or comes installed on Ubuntu 21.04).

Can you summarize the problem? I vaguely remember a discussion from last year, one of the julia or cxxwrap devs said the temp solution we chose would be a problem long term. Can we go back now and do the better thing?

lanctot:
GitHub Actions down again took several hours to fix yesterday: https://github.com/actions/virtual-environments/issues/2179

lanctot:
Hmm, my bad-- it's not github actions, it's because ubuntu-21.04 isn't defined yet. I know that  they do have a `ubuntu-latest` (which I think is 20.10). Can you try that one instead?

lanctot:
Hi @findmyway, I did a bunch of searches and it seems like the copy is still the temporary fix for this. 

I think it hasn't worked for you because you're copying it to the wrong place. I tried running the command you have in actions.yml and it didn't point to the right place.

How about we just hard-code it for GithubActions (to something like /usr/local/julia-1.6.1/lib/julia ?) and restore the copy in the install script for custom installations in $HOME.

lanctot:
Actually the location seems to be `/opt/hostedtoolcache/julia/1.6.1/x64/bin/julia`.

How about using:
```
$(which julia)../../lib/julia
```
?

lanctot:
Oh wait, that will be a symbolic link. You need to follow that link to the real binary, then add `/../../lib/julia` :)

....or maybe just try a hard-coded `/opt/hostedtoolcache/julia/1.6.1/x64/lib/julia` just to see if it'll work first

.github/workflows/actions.yml
```diff
@@ -75,6 +75,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: julia-actions/setup-julia@v1
+    - name: Ad-hoc fix
+      if: ${{ matrix.os == 'ubuntu-latest' }}
+      run: |
+        # workaround for https://github.com/deepmind/open_spiel/issues/606
+        sudo cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 $(julia --startup-file=no -E 'using Libdl;print(abspath(joinpath(Libdl.dlpath("libjulia"), "..", "julia")))')
```
lanctot, line 83: Can you try copying to `/opt/hostedtoolcache/julia/1.6.1/x64/lib/julia` instead ?

findmyway:
> Can you summarize the problem? I vaguely remember a discussion from last year, one of the julia or cxxwrap devs said the temp solution we chose would be a problem long term. Can we go back now and do the better thing?

The main reason is that prebuilt Julia binary ships with a specific version of `libstdc++.so`. And some time ago, only `GLIBCXX_3.4.26` (or even older I guess?) is required for `libopenspieljl`. But now even with Julia@v1.6, the `libstdc++.so` is still left behind the required version, (and I'm really not sure why it requires `GLIBCXX_3.4.29` now, but I believe this is the key issue we have).

To use the latest `libstdc++.so`, either build the Julia from source, so that it will use the default `libstdc++` on the system, or load the library on demand (this work is still ongoing and it's unclear when it will be resolved)

.github/workflows/actions.yml
findmyway: The result should be the same. But I'll give it a try.

lanctot:
It works, ok I will import it now and make a few minor changes. Thanks for fixing!

findmyway:
That's ... my bad. 😿 
I wrongly used `-E` instead of `-e` parameter here.

lanctot:
This has now been merged.

PR #612, opened 2021-05-21
fix bug in oshi_zumo where returns are sometimes inverted

VitamintK:
In oshi_zumo, there was a bug where pushing the wrestler completely off the board counted as a loss instead of a win.  (With `Alesia`, this means that all returns were inverted, so always playing `0` was a dominant strategy.)

To repro:
```
game = pyspiel.load_game('oshi_zumo',
     {
         "coins": pyspiel.GameParameter(2),
         "size": pyspiel.GameParameter(1),
         "horizon": pyspiel.GameParameter(2)
      })
game = pyspiel.convert_to_turn_based(game)

print('off the board:')
state = game.new_initial_state().child(1).child(0).child(1).child(0)
print(state)
print(state.returns(), '\n')

print('still on the board:')
state = game.new_initial_state().child(0).child(0).child(1).child(0)
print(state)
print(state.returns())
```

prints
```
off the board:
Coins: 0 2, Field: #...W

[-1.0, 1.0] 
```

and

```
still on the board:
Coins: 1 2, Field: #..W#

[1.0, -1.0]
```

### DIFF
--- a/open_spiel/algorithms/tabular_best_response_mdp_test.cc
+++ b/open_spiel/algorithms/tabular_best_response_mdp_test.cc
@@ -145,7 +145,7 @@ void OshiZumoGameTests() {
       "oshi_zumo(coins=10,size=3,min_bid=1)");
   TabularBestResponseMDP tbr1(*game, uniform_policy);
   TabularBestResponseMDPInfo br_info = tbr1.NashConv();
-  SPIEL_CHECK_FLOAT_NEAR(br_info.nash_conv, 1.997891313932980, kFloatTolerance);
+  SPIEL_CHECK_FLOAT_NEAR(br_info.nash_conv, 1.988311287477953, kFloatTolerance);
 
   TabularBestResponseMDP tbr2(*game, first_action_policy);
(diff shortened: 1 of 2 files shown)


VitamintK:
I also changed the value for `OshiZumoGameTests` in `tabular_best_response_mdp_test.cc` to the updated value.  I don't know how to verify that the value of the test is correct, so I just blindly changed it, but I see no reason that it would be wrong.

lanctot:
Yikes! Nice catch, thanks.

For the NashConv value did you just run the current best response code (the non-mdp one) to get the value? That is what I did. If you didn't can you check it?

If the values differ then there is likely still a bug in handling simultaneous move games in the mdp BR and you can just comment it out like I did for Goofspiel. But I am curious if those two different BRs agree on the value.

VitamintK:
>  It was actually "misere oshi-zumo"
😆 
> For the NashConv value did you just run the current best response code (the non-mdp one) to get the value? 
Just did this and confirmed the value is correct.

Issue #617, opened 2021-06-01
Cannot generate play-through for `python_mfg_crowd_modelling`


TheoCabannes:
In https://github.com/TheoCabannes/open_spiel/tree/use-init-in-mfg-modules, if I try to run `./open_spiel/scripts/generate_new_playthrough.sh python_mfg_crowd_modelling` then I got the following error:
```python3
Traceback (most recent call last):
  File "open_spiel/python/examples/playthrough.py", line 72, in <module>
    app.run(main)
  File "/Users/theophile/miniconda3/envs/dev/lib/python3.7/site-packages/absl/app.py", line 303, in run
    _run_main(main, args)
  File "/Users/theophile/miniconda3/envs/dev/lib/python3.7/site-packages/absl/app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "open_spiel/python/examples/playthrough.py", line 63, in main
    FLAGS.game, actions, alsologtostdout=FLAGS.alsologtostdout)
  File "/Users/theophile/miniconda3/envs/dev/lib/python3.7/site-packages/open_spiel-0.3.1-py3.7-macosx-10.9-x86_64.egg/open_spiel/python/algorithms/generate_playthrough.py", line 129, in playthrough
    observation_params_string)
  File "/Users/theophile/miniconda3/envs/dev/lib/python3.7/site-packages/open_spiel-0.3.1-py3.7-macosx-10.9-x86_64.egg/open_spiel/python/algorithms/generate_playthrough.py", line 358, in playthrough_lines
    default_observation.set_from(state, player)
  File "/Users/theophile/miniconda3/envs/dev/lib/python3.7/site-packages/open_spiel-0.3.1-py3.7-macosx-10.9-x86_64.egg/open_spiel/python/mfg/games/crowd_modelling.py", line 281, in set_from
    self.dict["t"][state.t] = 1
IndexError: index 10 is out of bounds for axis 0 with size 10
```
I am willing to help debugging this issue, but it might be easier with some help. I will open a PR to follow the issue.

PR #618, opened 2021-06-01
Create python MFG module and import it in automatic tests.

TheoCabannes:
The tests for this PR will fail because I cannot generate a play-through for `python_mfg_crowd_modelling` (https://github.com/deepmind/open_spiel/issues/617). Therefore `integration_tests/playthrough_test.py` fails.
It would be great to enable generating playthrough for python mean field games as it should help for developers (like me) that would like to implement a mean field game in python.


TheoCabannes:
Closing this PR as it might be address in next sync with https://github.com/deepmind/open_spiel/issues/617

PR #620, opened 2021-06-03
removing the game call to GoofSpiel

balloch:
The call to game GoofSpiel breaks the tutorial. In cell "id": "orVhRjWBsBWx", it explicitly states about likelihood of betting with a a king, jack, or queen, which is clearly alluding to Kuhn Poker, not goofspiel. Removing the game change makes the tutorial run fine, and all outputs make sense


balloch:
@googlebot I signed it!

PR #625, opened 2021-06-08
Enable getting policy when no legal actions

TheoCabannes:
This is linked to https://github.com/deepmind/open_spiel/issues/588


TheoCabannes:
Actually, I should check the notebook because it seems there is a lot of changes.

PR #628, opened 2021-06-11
Distinguish between new and old states in distribution

TheoCabannes:
When using distribution, I have found the code hard to read. Also the code might give weird results in case observation_string has not been well implemented.
To address these issues by:
- adding the new state to a temporary dictionary before updating the main dictionary of strings
- adding an assertion about the sum of the probabilities of each states
- adding an assertion about the string representation of the new states not being similar than the old string representation

### DIFF
--- a/open_spiel/python/mfg/algorithms/distribution.py
+++ b/open_spiel/python/mfg/algorithms/distribution.py
@@ -79,9 +79,7 @@ class DistributionPolicy(distribution.Distribution):
             if new_mfg_state_str not in new_distribution_over_states:
               new_listing_states.append(new_mfg_state)
             new_distribution_over_states[
-                new_mfg_state_str] += prob * self.distribution_over_states[
-                    mfg_state.observation_string(
-                        pyspiel.PlayerId.DEFAULT_PLAYER_ID)]
+                new_mfg_state_str] += prob * self.value(mfg_state)
 
       elif self.player_id_from_states(
           listing_states) == pyspiel.PlayerId.MEAN_FIELD:
@@ -96,8 +94,7 @@ class DistributionPolicy(distribution.Distribution):
           new_listing_states.append(new_mfg_state)
           new_distribution_over_states[new_mfg_state.observation_string(
               pyspiel.PlayerId.DEFAULT_PLAYER_ID
-          )] = self.distribution_over_states[mfg_state.observation_string(
-              pyspiel.PlayerId.DEFAULT_PLAYER_ID)]
+          )] = self.value(mfg_state)
 
       else:
         assert self.player_id_from_states(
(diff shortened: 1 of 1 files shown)


TheoCabannes:
This PR is particularly important to me, because a MFG instantiation should distinguish in their string observer the current player state, the current time and the current player type for the distribution to work correctly (i.e. probabilities sum to 1). However, a default string observer might be `history_str()` which does not distinguish the states with different player types.
With this PR, implementing `history_str()` for the game observer will raise `"Some new states have the same string representation of old states."` which seems to be an important feature to me.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -65,16 +65,20 @@ def evaluate(self):
 
     while not self.is_terminal_from_states(listing_states):
       new_listing_states = []
+      new_distribution_over_states = collections.defaultdict(float)
```
RaphaelMarinier, line 68: With this new data structure for the new distribution, self.distribution_over_states no longer needs to be a default dict (which will also be safer to ensure that we don't ask for states that are not tracked, and magically get a proba of 0).

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -65,16 +65,20 @@ def evaluate(self):
...
 
       if self.player_id_from_states(listing_states) == pyspiel.PlayerId.CHANCE:
         for mfg_state in listing_states:
           for action, prob in mfg_state.chance_outcomes():
             new_mfg_state = mfg_state.child(action)
             new_mfg_state_str = new_mfg_state.observation_string(
                 pyspiel.PlayerId.DEFAULT_PLAYER_ID)
-            if new_mfg_state_str not in self.distribution_over_states:
+            # as a state can be the child of two different parent states, we do
```
RaphaelMarinier, line 76: Comments should be proper sentences. Start with an uppercase.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -90,26 +94,39 @@ def evaluate(self):
...
                 new_mfg_state_str] += prob * self.distribution_over_states[
                     mfg_state.observation_string(
                         pyspiel.PlayerId.DEFAULT_PLAYER_ID)]
+
+      assert all(state_str not in self.distribution_over_states
+                 for state_str in new_distribution_over_states), (
+        "Some new states have the same string representation of old states.")
+      self.distribution_over_states.update(new_distribution_over_states)
+      sum_state_probabilities = sum(map(self.value, new_listing_states))
+      assert abs(sum_state_probabilities - 1) < 1e-4, (
```
RaphaelMarinier, line 122: This is no longer true in the multi-population case (the value of the sum will be the number of populations).
Let me see if we can push the multi-pop support to github so that you can update your PR.

RaphaelMarinier (changes requested):

open_spiel/python/mfg/algorithms/distribution.py
lanctot: > Let me see if we can push the multi-pop support to github so that you can update your PR.

Updated. See this commit: https://github.com/deepmind/open_spiel/commit/ce3c25a788319a58076732cdf28c7ef5651b2ae7. Thanks @RaphaelMarinier!

TheoCabannes:
Because of the large change in master due to multipopulation MFG, I will close this PR and re-open a new one.

PR #631, opened 2021-06-13
Some suggested updates to improve MFG code

TheoCabannes:
Feel free to remove the PR if the "improvements" are not improvements.
- Fix Pylint no-else-return, dangerous-default-value and super-with-arguments warning
- Use property in crowd modelling python
- Enable empty legal actions in best response value
- parametrized game info in crowd modelling python

### DIFF
--- a/open_spiel/python/mfg/algorithms/best_response_value.py
+++ b/open_spiel/python/mfg/algorithms/best_response_value.py
@@ -85,11 +85,12 @@ class BestResponse(value.ValueFunction):
     assert state.current_player() == pyspiel.PlayerId.DEFAULT_PLAYER_ID, (
       "The player id should be 0")
     state_legal_actions = state.legal_actions()
-    max_q = 0
     if state_legal_actions:
       max_q = max(
           self.eval_state(state.child(action))
           for action in state_legal_actions)
+    else:
+      max_q = self.eval_state(state.child(0))
     self._state_value[state_str] = state.rewards()[0] + max_q
     return self._state_value[state_str]
 
--- a/open_spiel/python/mfg/algorithms/best_response_value.py
+++ b/open_spiel/python/mfg/algorithms/best_response_value.py
@@ -35,7 +35,7 @@ class BestResponse(value.ValueFunction):
       root_state: The state of the game at which to start. If `None`, the game
         root state is used.
     """
-    super(BestResponse, self).__init__(game)
+    super().__init__(game)
     if root_state is None:
       root_state = game.new_initial_state()
     self._root_state = root_state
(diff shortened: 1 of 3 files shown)


open_spiel/python/policy.py
```diff
@@ -268,7 +265,13 @@ def policy_for_key(self, key):
...
       ```
     """
-    return self.action_probability_array[self.state_lookup[key]]
+    try:
+      return self.action_probability_array[self.state_lookup[key]]
+    except KeyError:
+      # In case they are no legal actions in a state, then the state
+      # will not be added to action_probability_array in the initialization,
+      # therefore 0 should be played.
+      return {0: 1}
```
TheoCabannes, line 274: I will update this part of the code once https://github.com/deepmind/open_spiel/pull/625/files get merged.

open_spiel/python/mfg/algorithms/best_response_value.py
```diff
@@ -82,13 +82,16 @@ def eval_state(self, state):
       self._state_value[state_str] = (
           state.rewards()[0] + self.eval_state(new_state))
       return self._state_value[state_str]
-    else:
-      assert state.current_player() == 0, "The player id should be 0"
+    assert state.current_player() == pyspiel.PlayerId.DEFAULT_PLAYER_ID, (
```
RaphaelMarinier, line 85: (Not true anymore with multi-pop, we'll try to push the support for it so that you can sync).

open_spiel/python/mfg/algorithms/best_response_value.py
```diff
@@ -82,13 +82,16 @@ def eval_state(self, state):
...
+      "The player id should be 0")
+    state_legal_actions = state.legal_actions()
+    max_q = 0
+    if state_legal_actions:
```
RaphaelMarinier, line 88: Will there be a case with no legal actions?

open_spiel/python/mfg/games/crowd_modelling.py
```diff
@@ -72,8 +72,14 @@ class MFGCrowdModellingGame(pyspiel.Game):
   """
 
   # pylint:disable=dangerous-default-value
-  def __init__(self, params: Mapping[str, Any] = _DEFAULT_PARAMS):
-    super().__init__(_GAME_TYPE, _GAME_INFO, params)
+  def __init__(
+    self, params: Mapping[str, Any] = None, game_info: Mapping[str, Any] = None
+  ):
+    if game_info is None:
+      game_info = _GAME_INFO
```
RaphaelMarinier, line 79: I think that it would be better to dynamically build here the game_info that you pass to the super class, based on the parameters. You just need to fill max_chance_outcomes, max_game_length correctly based on the parameters.

TheoCabannes: Thanks. Will do.

open_spiel/python/mfg/algorithms/best_response_value.py
TheoCabannes: Yes, for my implementation of routing game, one player might have no action (no routing choice) at a specific time step.

open_spiel/python/mfg/algorithms/best_response_value.py
TheoCabannes: Thanks

TheoCabannes:
Because of the large change in master due to multipopulation MFG, I will close this PR and re-open a new one.

PR #634, opened 2021-06-17
Pytorch Deep CFR

rainwangphy:
I only change the deep CFR file in Pytorch.

The main changes are:
1. Adding a new implementation of MLP with new init, following what did in Pytorch NFSP
2. Replacing the init of MLP in deep CFR codes with new (due to the difference of the two implementations of MLP)

Hope this can work.


rainwangphy:
@googlebot I signed it!

lanctot:
Thanks @rainwangphy! (For more detail, see: https://github.com/deepmind/open_spiel/issues/633)

PR #638, opened 2021-06-21
Distinguish old and new states in distribution and add assertions

TheoCabannes:
Assert that the sum over the distribution is always equal to the number of players/populations.

I figure out that they are no tests for multipopulation MFG in python.

### DIFF
--- a/open_spiel/python/mfg/algorithms/best_response_value.py
+++ b/open_spiel/python/mfg/algorithms/best_response_value.py
@@ -75,13 +75,8 @@ class BestResponse(value.ValueFunction):
       return self._state_value[state_str]
     elif state.current_player() == pyspiel.PlayerId.MEAN_FIELD:
       dist_to_register = state.distribution_support()
-      def get_state_probability(str_state):
-        try:
-          return self._distribution.value_str(str_state)
-        except ValueError:
-          return 0
       dist = [
-          get_state_probability(str_state)
+          self._distribution.value_str(str_state, 0.)
           for str_state in dist_to_register
       ]
       assert abs(sum(dist) - self.game.num_players()) < 1e-4, (
--- a/open_spiel/python/mfg/algorithms/best_response_value.py
+++ b/open_spiel/python/mfg/algorithms/best_response_value.py
@@ -75,10 +75,18 @@ class BestResponse(value.ValueFunction):
       return self._state_value[state_str]
     elif state.current_player() == pyspiel.PlayerId.MEAN_FIELD:
       dist_to_register = state.distribution_support()
+      def get_state_probability(str_state):
+        try:
+          return self._distribution.value_str(str_state)
+        except ValueError:
+          return 0
       dist = [
-          self._distribution.value_str(str_state)
+          get_state_probability(str_state)
           for str_state in dist_to_register
       ]
+      assert abs(sum(dist) - self.game.num_players()) < 1e-4, (
+            "Sum of probabilities of all possible states should be the number "
+            f"of population, it is {sum(dist)}.")
       new_state = state.clone()
       new_state.update_distribution(dist)
       self._state_value[state_str] = (
--- a/open_spiel/python/mfg/algorithms/distribution.py
+++ b/open_spiel/python/mfg/algorithms/distribution.py
@@ -33,6 +33,7 @@ class DistributionPolicy(distribution.Distribution):
     """
     super(DistributionPolicy, self).__init__(game)
     self._policy = policy
+    self.game = game
     if root_state is None:
       self._root_states = game.new_initial_states()
     else:
@@ -61,7 +62,7 @@ class DistributionPolicy(distribution.Distribution):
     # Maps state strings to floats. For each group of states at a
     # given timestep, given player ID and given population, these
     # floats represent a probability distribution.
-    self.distribution = collections.defaultdict(float)
+    self.distribution = {}
     for state in listing_states:
       self.distribution[state.observation_string(
           pyspiel.PlayerId.DEFAULT_PLAYER_ID)] = 1.
@@ -87,10 +88,18 @@ class DistributionPolicy(distribution.Distribution):
             pyspiel.StateType.MEAN_FIELD):
         for mfg_state in listing_states:
           dist_to_register = mfg_state.distribution_support()
-          dist = list(map(self.value_str, dist_to_register))
-          assert (sum(dist) - len(self._root_states)) < 1e-4, (
-            "Sum of probabilities of all possible states should be the number of "
-            f"population, it is {sum(dist)}.")
+          def get_probability_for_state(str_state):
+            try:
+              return self.value_str(str_state)
+            except ValueError:
+              return 0
+          dist = [
+            get_probability_for_state(str_state)
+            for str_state in dist_to_register
+          ]
+          assert (sum(dist) - self.game.num_players()) < 1e-4, (
+            "Sum of probabilities of all possible states should be the number "
+            f"of population, it is {sum(dist)}.")
           new_mfg_state = mfg_state.clone()
           new_mfg_state.update_distribution(dist)
           new_listing_states.append(new_mfg_state)
@@ -113,12 +122,14 @@ class DistributionPolicy(distribution.Distribution):
               new_listing_states.append(new_mfg_state)
             new_distribution[new_mfg_state_str] += prob * self.value(mfg_state)
 
-      assert all(state_str not in self.distribution
-            for state_str in new_distribution), (
-          "Some new states have the same string representation of old states.")
+      for state_str in new_distribution:
+        if state_str in self.distribution:
+          raise ValueError(
+            f'{state_str} has already been seen in distribution.')
       self.distribution.update(new_distribution)
-      sum_state_probabilities = sum(map(self.value, new_listing_states))
-      assert abs(sum_state_probabilities - len(self._root_states)) < 1e-4, (
+      sum_state_probabilities = sum(
+        [self.value(state) for state in new_listing_states])
+      assert abs(sum_state_probabilities - self.game.num_players()) < 1e-4, (
         "Sum of probabilities of all possible states should be the number of "
         f"population, it is {sum_state_probabilities}.")
       listing_states = new_listing_states
@@ -128,8 +139,9 @@ class DistributionPolicy(distribution.Distribution):
         mfg_state.observation_string(pyspiel.PlayerId.DEFAULT_PLAYER_ID))
 
   def value_str(self, mfg_state_str):
-    v = self.distribution.get(mfg_state_str)
-    if v is None:
+    try:
+      return self.distribution[mfg_state_str]
+    except KeyError as e:
       # Check this because self.distribution is a default dict.
-      raise ValueError(f'Distribution not computed for state {mfg_state_str}')
-    return v
+      raise ValueError(
+        f'Distribution not computed for state {mfg_state_str}') from e
--- a/open_spiel/python/mfg/algorithms/distribution.py
+++ b/open_spiel/python/mfg/algorithms/distribution.py
@@ -133,10 +133,25 @@ class DistributionPolicy(distribution.Distribution):
         mfg_state.observation_string(pyspiel.PlayerId.DEFAULT_PLAYER_ID))
 
   def value_str(self, mfg_state_str, default_value = None):
+    """Return probability of the state encoded by mfg_state_str.
+
+    Args:
+      mfg_state_str: string description of the state. This should be created
+        using observation_string.
+      default_value: in case the state has not been seen by the distribution, to
+        avoid raising a value error the default value is returned if it is not
+        None.
+    Returns:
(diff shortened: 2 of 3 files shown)


open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -68,33 +68,34 @@ def evaluate(self):
 
     while not self.is_terminal_from_states(listing_states):
       new_listing_states = []
+      new_distribution = collections.defaultdict(float)
```
RaphaelMarinier, line 71: I commented on this in the initial PR (https://github.com/deepmind/open_spiel/pull/628/commits/d058d5da4e3da6acf13dc9f110e227bb27911800):
With this new data structure for the new distribution, self.distribution no longer needs to be a default dict (which will also be safer to ensure that we don't ask for states that are not tracked, and magically get a proba of 0).

Can you also check if there are other comments I made in the initial PR that are not addressed?

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -68,33 +68,34 @@ def evaluate(self):
...
+            new_distribution[new_mfg_state_str] += prob * self.value(mfg_state)
 
       elif (self.type_from_states(listing_states) ==
             pyspiel.StateType.MEAN_FIELD):
         for mfg_state in listing_states:
           dist_to_register = mfg_state.distribution_support()
-          dist = [
-              self.distribution[str_state] for str_state in dist_to_register
-          ]
+          dist = list(map(self.value_str, dist_to_register))
```
RaphaelMarinier, line 90: I would keep the list comprehension style:
dist = [self.value_str(state) for state in dist_to_register]
map() is generally discouraged over list comprehension:

https://google.github.io/styleguide/pyguide.html#271-definition
https://google.github.io/styleguide/pyguide.html#210-lambda-functions

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -68,33 +68,34 @@ def evaluate(self):
...
+          assert (sum(dist) - len(self._root_states)) < 1e-4, (
```
RaphaelMarinier, line 91: self._root_states -> self.game.num_players() might be clearer.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -105,11 +106,21 @@ def evaluate(self):
...
+            if new_mfg_state_str not in new_distribution:
               new_listing_states.append(new_mfg_state)
-            self.distribution[new_mfg_state_str] += prob * self.distribution[
-                mfg_state.observation_string(
-                    pyspiel.PlayerId.DEFAULT_PLAYER_ID)]
+            new_distribution[new_mfg_state_str] += prob * self.value(mfg_state)
+
+      assert all(state_str not in self.distribution
+            for state_str in new_distribution), (
+          "Some new states have the same string representation of old states.")
```
RaphaelMarinier, line 118: Can an example be printed, without complicating the code too much?

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -105,11 +106,21 @@ def evaluate(self):
...
+      self.distribution.update(new_distribution)
+      sum_state_probabilities = sum(map(self.value, new_listing_states))
```
RaphaelMarinier, line 120: Use list comprehension / generator.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -105,11 +106,21 @@ def evaluate(self):
...
+      assert abs(sum_state_probabilities - len(self._root_states)) < 1e-4, (
```
RaphaelMarinier, line 121: len(root_states) -> num_players() as above.

RaphaelMarinier (changes requested):
Regarding multi-population tests, they are coming, they are in review on our side.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Sorry for not mentioning that. In the current code distribution is still a default dict, therefore I used also a default dict for new distribution. Also new distribution is never used to get probability (but only to update distribution) so they should not be any issue with magical 0 proba.

TheoCabannes: Also value_str check if the key exists.

TheoCabannes: Should I still move to a dict and not a default dict? Also should I update self.distribution to be a dict too?

RaphaelMarinier: Having new_distribution be a defaultdict is good I think (otherwise, "new_distribution[new_mfg_state_str] += prob * self.value(mfg_state)" would be more complicated).

My proposal is to have "self.distribution" be a normal dict, since we no longer need the "default dict" property for that one (and we could remove the manual check in value_str()).

TheoCabannes: Thanks for the clarification, I misunderstood your first remark. That makes sense!

TheoCabannes: I have one question I did not manage to answer by myself: when calling the probability of each state in the distribution support, the new implementation will require that each state in the distribution support has been seen already, which might not be the case.
For example in the routing game, a possible state might be to be on link A->B. But maybe as nobody has reached the link A->B yet, the distribution has not seen A->B yet and this will raise a KeyError.
Does it make sense? Do you have any idea about potential solution (other than casting KeyError in line 88/90?

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Done. I had to add the game as an attribute of the distribution.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Thanks for the pointers!

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Done.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Done.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Done as above.

TheoCabannes (commented):
Thanks for the review

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, game, policy, root_state=None):
     """
     super(DistributionPolicy, self).__init__(game)
     self._policy = policy
+    self.game = game
```
RaphaelMarinier, line 36: Not needed, the parent class (distribution.Distribution) should already have this field.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -128,8 +139,9 @@ def value(self, mfg_state):
         mfg_state.observation_string(pyspiel.PlayerId.DEFAULT_PLAYER_ID))
 
   def value_str(self, mfg_state_str):
-    v = self.distribution.get(mfg_state_str)
-    if v is None:
+    try:
+      return self.distribution[mfg_state_str]
+    except KeyError as e:
       # Check this because self.distribution is a default dict.
```
RaphaelMarinier, line 145: Please update the comment.

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -87,10 +88,18 @@ def evaluate(self):
             pyspiel.StateType.MEAN_FIELD):
         for mfg_state in listing_states:
           dist_to_register = mfg_state.distribution_support()
-          dist = list(map(self.value_str, dist_to_register))
-          assert (sum(dist) - len(self._root_states)) < 1e-4, (
-            "Sum of probabilities of all possible states should be the number of "
-            f"population, it is {sum(dist)}.")
+          def get_probability_for_state(str_state):
```
RaphaelMarinier, line 91: Ah, I did not realize that we needed to fallback to a default 0 value here, but indeed we do.
I would not use exceptions for that, since it's not something exceptional (and catching exception can be slow).
I would call directly self.distribution.get(str_state, 0.) here.

See also:
https://google.github.io/styleguide/pyguide.html#24-exceptions

RaphaelMarinier (changes requested):
Last request: could you run the benchmark (https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/mfg/algorithms/benchmark.py) before and after your change, to check that your change does not affect the speed too much?
Thanks!

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Thanks

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Thanks for catching that.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: THanks

TheoCabannes: I would have to update https://github.com/deepmind/open_spiel/blob/104deea0f931f960b362a38e498410385ac5e81b/open_spiel/python/mfg/algorithms/best_response_value.py#L79

open_spiel/python/mfg/algorithms/best_response_value.py
```diff
@@ -75,10 +75,18 @@ def eval_state(self, state):
       return self._state_value[state_str]
     elif state.current_player() == pyspiel.PlayerId.MEAN_FIELD:
       dist_to_register = state.distribution_support()
+      def get_state_probability(str_state):
+        try:
+          return self._distribution.value_str(str_state)
+        except ValueError:
```
RaphaelMarinier, line 81: Also here, don't use the exception mechanism for something that can happen under normal conditions.
I would modify the Distribution API to allow for an argument with a default value when the state is not in the distrib, so that you could call value_str(str_state, 0.)

RaphaelMarinier (changes requested):

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -132,9 +132,11 @@ def value(self, mfg_state):
...
-      raise ValueError(
-        f'Distribution not computed for state {mfg_state_str}') from e
+  def value_str(self, mfg_state_str, default_value = None):
+    if default_value is None:
+      try:
+        return self.distribution[mfg_state_str]
+      except KeyError as e:
+        raise ValueError(
+          f'Distribution not computed for state {mfg_state_str}') from e
+    return self.distribution.get(mfg_state_str, 0.)
```
RaphaelMarinier, line 142: 0. -> default_value

open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -132,9 +132,11 @@ def value(self, mfg_state):
...
         mfg_state.observation_string(pyspiel.PlayerId.DEFAULT_PLAYER_ID))
 
-  def value_str(self, mfg_state_str):
-    try:
-      return self.distribution[mfg_state_str]
-    except KeyError as e:
-      raise ValueError(
-        f'Distribution not computed for state {mfg_state_str}') from e
+  def value_str(self, mfg_state_str, default_value = None):
+    if default_value is None:
```
RaphaelMarinier, line 136: Can you add a docstring, saying that an exception will be raised if the value is not found and default_value is None?

open_spiel/python/mfg/algorithms/best_response_value.py
TheoCabannes: Thanks

RaphaelMarinier (changes requested):

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Thanks for catching that.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: Done

TheoCabannes:
Here are the results of benchmark before the changes:
Before_changes
Here are the results of benchmark after the changes:
After_changes

RaphaelMarinier:
Thanks. We are importing the change.

lanctot:
This has now been merged into master via https://github.com/deepmind/open_spiel/commit/610dfdf2c6cab600601560b4a15a8df5864aeb14

Thanks @TheoCabannes !

PR #639, opened 2021-06-21
Suggestion for mfg python code

TheoCabannes:
- fix pylint errors: super with arguments.
- fix pylint errors: arguments-differ.
- fix pylint errors: invalid-name.
- fix pylint errors: method could be a function.
- fix pylint errors: no else return
- fix pylint errors: useless-object-inheritance
- do not have a default game info in crowd modelling python as Raphael suggested.
- use property x and t in crowd modelling python

### DIFF
--- a/open_spiel/python/mfg/algorithms/distribution.py
+++ b/open_spiel/python/mfg/algorithms/distribution.py
@@ -27,14 +27,6 @@ def type_from_states(states):
   return types[0]
 
 
-def is_terminal_from_states(states):
-  """Get is_terminal of a list of states and assert they are the same."""
-  is_terminal_from_states_ = [state.is_terminal() for state in states]
-  for is_terminal in is_terminal_from_states_:
-    assert is_terminal_from_states_[0] == is_terminal
-  return is_terminal_from_states_[0]
-
-
 class DistributionPolicy(distribution.Distribution):
   """Computes the distribution of a specified strategy."""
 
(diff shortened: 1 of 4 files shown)


open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -19,6 +19,22 @@
...
 
+def type_from_states(states):
+  """Get node type of a list of states and assert they are the same."""
+  types = [state.get_type() for state in states]
+  for state_type in types:
+    assert types[0] == state_type, f'types: {types}'
+  return types[0]
+
+
+def is_terminal_from_states(states):
```
RaphaelMarinier, line 30: We could actually either:
- get rid of this, and instead use type_from_states(states) == pyspiel.StateType.TERMINAL on line 71, or
- replace the body of this function with "return type_from_states(states) == pyspiel.StateType.TERMINAL"

open_spiel/python/mfg/games/crowd_modelling.py
```diff
@@ -143,18 +144,16 @@ def _legal_actions(self, player):
     """Returns a list of legal actions for player and MFG nodes."""
     if player == pyspiel.PlayerId.MEAN_FIELD:
       return []
-    elif player == 0 and player == self.current_player():
+    if player == 0 and player == self.current_player():
```
RaphaelMarinier, line 147: DEFAULT_PLAYER_ID?

RaphaelMarinier (changes requested):
Looks good overall.

open_spiel/python/mfg/games/crowd_modelling.py
TheoCabannes: Done.

open_spiel/python/mfg/algorithms/distribution.py
TheoCabannes: I chose open 2.

RaphaelMarinier (approved):
LG

PR #643, opened 2021-06-25
Enable no legal actions in get all states

TheoCabannes:
This enables game where a player might have no action at a specific turn but can have an action later.


TheoCabannes:
Based on discussion with Ed and Julien, legal_actions should always return non empty array (except for some players in simultaneous games).

PR #644, opened 2021-06-25
Do not appen state with 0.0 probability in FP

TheoCabannes:
Due to the fact that floating points can not represent value very small, probability of a state can be 0.0. In this case a DivideByZero error will happen. To prevent that we should check that the probability of a state is not 0.

For example if you want to model radioactive decay law (with mean field), at one point the quantity of atoms will be so small that floating point that represents it will be 0.0.


TheoCabannes:
I discuss with Julien Perolat, and this should be taken into account internally.

TheoCabannes:
This has been done by https://github.com/deepmind/open_spiel/commit/6a813471a3875cf736ed84c31220288e06eae5d0

PR #646, opened 2021-06-28
Add assertion about legal_actions() being non empty.

TheoCabannes:
[NOT URGENT][TESTS]
Adding a test that legal_actions() is not empty when generating play-through.


TheoCabannes:
Will review test in random simulation before submitting this PR if needed.

TheoCabannes:
This is a to do in MFGs https://github.com/deepmind/open_spiel/blob/bbfb0259b5d26c1e3f05000f1f4a05b7ad60393e/open_spiel/games/mfg/crowd_modelling_test.cc#L79 and should be done in basic_tests.cc

Issue #49, opened 2019-09-12
Question: information state targetting 


michalsustr:
I'm trying to rewrite OOS (http://mlanctot.info/files/papers/aamas15-iioos.pdf) and specifically information state targeting.

Part of it is to write a function which takes two ActionObservation sequences and returns if they are compatible (it's possible to get from the current infoset to the target infoset).

In OpenSpiel, neither the `InformationState` / `InformationStateAsNormalizedVector` provide "incremental" information: it's a domain-dependent representation of the whole AO sequence. 
`Observation` or `ObservationAsNormalizedVector` do these "incremental" observations: however how is the observation of "no observation" encoded in phantom games? Empty vector?

Also, do you think it would make sense to store the ActionObservation for each player in State (similarly to how history_ is stored)? It is not possible to do a currentState->parentState lookup, so to retrieve the whole AO for arbitrary state you'd have to make targetted tree traversal from the root to this state. 

I'm asking this specifically in the context of OOS, where you will be given a target infoset (or State, in the Bot API), but you might not know what it's ActionObservations are (if you didn't visit this infoset before and cached it). The result is that you cannot target this infoset efficiently.

michalsustr:
> That seems costly to have turned on by default (since the observations are potentially large vectors)

Did you consider other representation of observations, as the one for neural networks? Maybe something game specific, for example GoofspielObservation which would contain the actual information (win/lose/draw in a round, nature card and possibly cards revealed) from which other representations could be built (such as for neural networks?). These game-specific observations can be much much smaller (probably 1 or 2 orders of magnitude compared to one-hot encoding of doubles :) ). 

>  Support an optional way to turn it on, but have it switched off by default.

I'd be careful about putting more mutability/statefullness into State :) I suggest these kinds of settings should be `const`-ed, i.e. there can be a RootStateFactory(bool capabilityX, bool capabilityY, ...) which sets them up, and disallows user to turn them off later, so that we don't have some states which do have AOs and some which don't. 

I thought about subclassing for type safety, i.e. `StateWithCapabilityX`, but then all other methods will have to return `StateWithCapabilityX`, and make rest of the code type incompatible. This would indeed require templating... _everything_ ... 

This seems like rather a common design problem, so maybe someone with more C++ experience can help? @Dugy

> I can add a bunch of methods that return the collections and/or const refs to them.

[...]

PR #648, opened 2021-06-28
Add mean field node to basic tests

TheoCabannes:
This should enable using basic tests in mean field games.

Currently the tests are failing for:
- crowd_modelling_2d_test (Child aborted) (I do not know why).
- crowd_modelling_test.py and predator_prey_test.py due to `pyspiel.SpielError: MaxChanceNodesInHistory() is not implemented`. I do not understand why this issue is not raised in other python tests that call `random_sim_test`.
I am submitting this PR to get some help on it.

### DIFF
--- a/open_spiel/python/mfg/games/predator_prey.py
+++ b/open_spiel/python/mfg/games/predator_prey.py
@@ -98,7 +98,7 @@ class MFGPredatorPreyGame(pyspiel.Game):
     self.geometry = get_param("geometry", params)
     game_info = pyspiel.GameInfo(
         num_distinct_actions=_NUM_ACTIONS,
-        max_chance_outcomes=max(self.size, _NUM_CHANCE),
+        max_chance_outcomes=max(self.size * self.size, _NUM_CHANCE),
         num_players=num_players,
         min_utility=-np.inf,
         max_utility=+np.inf,
--- a/open_spiel/python/mfg/games/predator_prey.py
+++ b/open_spiel/python/mfg/games/predator_prey.py
@@ -116,6 +116,10 @@ class MFGPredatorPreyGame(pyspiel.Game):
     """
     return MFGPredatorPreyState(self)
 
+  def max_chance_nodes_in_history(self):
+    """Maximun chance nodes in game history."""
+    return self.horizon
+
   def new_initial_state_for_population(self, population):
     """State corresponding to the start of a game for a given population."""
     return MFGPredatorPreyState(self, population)
--- a/open_spiel/python/mfg/games/predator_prey.py
+++ b/open_spiel/python/mfg/games/predator_prey.py
@@ -118,7 +118,7 @@ class MFGPredatorPreyGame(pyspiel.Game):
 
   def max_chance_nodes_in_history(self):
     """Maximun chance nodes in game history."""
-    return self.horizon
+    return self.horizon + 1
 
   def new_initial_state_for_population(self, population):
     """State corresponding to the start of a game for a given population."""
--- a/open_spiel/python/mfg/games/predator_prey.py
+++ b/open_spiel/python/mfg/games/predator_prey.py
@@ -43,12 +43,14 @@ _DEFAULT_HORIZON = 10
 _NUM_ACTIONS = 5
 _NUM_CHANCE = 5
 _DEFAULT_REWARD_MATRIX = np.array([[0, -1, 1], [1, 0, -1], [-1, 1, 0]])
+_DEFAULT_NUM_PLAYERS = 3
 
 _DEFAULT_GEOMETRY = Geometry.SQUARE
 
 _DEFAULT_PARAMS = {
     "size": _DEFAULT_SIZE,
     "horizon": _DEFAULT_HORIZON,
+    "players": _DEFAULT_NUM_PLAYERS,
     # The reward matrix is represented as a string containing a
     # space-separated list of values.
     # Its size defines the number of populations in the mean field game.
@@ -87,7 +89,7 @@ class MFGPredatorPreyGame(pyspiel.Game):
     self.horizon = get_param("horizon", params)
     flat_reward_matrix = np.fromstring(
         get_param("reward_matrix", params), dtype=np.float, sep=" ")
-    num_players = int(round(np.sqrt(len(flat_reward_matrix))))
+    num_players = get_param("players", params)
     if len(flat_reward_matrix) != num_players**2:
       raise ValueError(
           f"Reward matrix passed in flat representation does not represent a "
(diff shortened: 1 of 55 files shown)


RaphaelMarinier:
Regarding crowd_modelling_2d_test, I have a high suspicion that this is caused by a bug when computing the observation tensor. Those bugs were fixed in the 1D game [here](https://github.com/deepmind/open_spiel/commit/052d88ca99063834cf8fe70bec3a1bdaa2d5ed45), but are still present in the 2D version. I just fixed them internally.

Regarding MaxChanceNodesInHistory, I'll have a look why you get this error with `MaxChanceNodesInHistory()`.

lanctot:
Hello @TheoCabannes, if you check the logs it seems that the errors have the same root cause:

```
MaxChanceNodesInHistory() not implemented.
```

It is referring to this function: https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/spiel.h#L859

Do MFGs have chance nodes? If so, we should implement the function for the MFGs. If not, can we we do a similar `return 0` here?

TheoCabannes:
Hello @lanctot, thank you a lot.
I have identify this function. I have also identify the binding in python https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/python/pybind11/pyspiel.cc#L351

However, what I do not understand is that `MaxChanceNodesInHistory() not implemented.` does not seem to be an issue for [python games](https://github.com/deepmind/open_spiel/tree/master/open_spiel/python/games): https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/python/games/iterated_prisoners_dilemma_test.py#L59 (btw just spotted a nit in the name of the test method `test_game_from_cc` should probably be `test_random_sim_game`) while `MaxChanceNodesInHistory() not implemented.` is raised in the mfg python games (c.f. the tests). 
I do not know if there is some binding issue in mfg python games for `MaxChanceNodesInHistory()` or if `MaxChanceNodesInHistory()` is never called for python game in random sim but is called for mfg python games in [RamdonSimulation](https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/tests/basic_tests.cc#L258).

**To make a long story short, I was not able to understand why I got this issue in the tests (and why it is not an issue for 'regular' python games). I think this might be an issue in the binding.**

TheoCabannes:
> Do MFGs have chance nodes? If so, we should implement the function for the MFGs. If not, can we we do a similar `return 0` here?
MFGs have chance nodes, should `max_chance_nodes_in_history` functions be implemented in python mfgs? Also it seems that `MaxChanceNodesInHistory` is not implemented for C++ mfgs (therefore I suspect that there is a default implementation for MFGs in C++ that might not exists in python).

lanctot:
I spoke to @perolat, he says we should implement that function in the MFG games. 

There may be other tests that fail after that, and we can handle them one-by-one as the tests fail.

Can you take charge of this @TheoCabannes  ?

lanctot:
There is a basic/empty/stub implementation in the base class, but it needs to be overridden by each game individually like the other methods. 

The error is getting triggered within the base class's default implementation, here: https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/spiel.h#L863

lanctot:
> MFGs have chance nodes. Should `max_chance_nodes_in_history` functions be implemented in python mfgs?

Yes, if we want to use the standard tests, they have to be. Same with the C++ MFGs, they each have to override `Game::MaxChanceNodesInHistory`.

And there may be other functions in this category once these are fixed. But probably best to catch them now before there are too many MFGs.

TheoCabannes:
Sure I can look at that. I will ask on mfg support chat room for more direction.

However I do not understand why the tests are passing for '#49 - crowd_modelling_test', MaxChanceNodesInHistory is not implemented in C++ crowd modelling but the test is passing.

lanctot:
> However I do not understand why the tests are passing for '#49 - crowd_modelling_test', MaxChanceNodesInHistory is not implemented in C++ crowd modelling but the test is passing.

Ah, yes it is implemented in that game (in the header file because it's small): https://github.com/deepmind/open_spiel/blob/9e4f6d219244dcd3deddf1ca75ab1113580bc73c/open_spiel/games/mfg/crowd_modelling.h#L134

TheoCabannes:
Thank you. My bad, I did not see it. Sorry about that.

lanctot:
> Thank you. My bad, I did not see it. Sorry about that.

No worries! It's good, because it provides an example. Also, because it is passing, maybe that means this is the only function that needs to be done before all the MFGs can move to the standard tests.

TheoCabannes:
I have added max_chance_nodes_in_history but the tests are failing on my computer due to https://github.com/deepmind/open_spiel/issues/657

TheoCabannes:
Converting to drafts as more bugs seems to exist. I will do another PR for crowd modelling C++ and another one for MaxChanceNodesInHistory binding.

lanctot:
Ok, great. But feel free to reuse this PR for everything related to these.

TheoCabannes:
@RaphaelMarinier what should be `max_chance_nodes_in_history` for the `MFGPredatorPreyGame`?
I put `horizon` but the tests are failing (you should be able to see the tests results soon).

Also, for MFGPredatorPreyGame test with random sim I changed the signature of `RandomSimTest` to :
```
void RandomSimTest(
    const Game& game, int num_sims,
    bool serialize = true, bool verbose = true, bool mask_test = true,
    const std::function<void(const State&)>& state_checker_fn
        = &DefaultStateChecker,
    int mean_field_population = -1);
```

lanctot:
I believe @RaphaelMarinier is on holiday. @perolat is telling me it should be like the other games, i.e. `horizon + 1`.

open_spiel/python/mfg/games/predator_prey.py
```diff
@@ -116,6 +116,10 @@ def new_initial_state(self):
     """
     return MFGPredatorPreyState(self)
 
+  def max_chance_nodes_in_history(self):
+    """Maximun chance nodes in game history."""
+    return self.horizon
```
lanctot, line 121: Try horizon + 1?

TheoCabannes:
Closing this PR as https://github.com/deepmind/open_spiel/pull/661 should solve the issue of predator prey game and https://github.com/deepmind/open_spiel/pull/662 has been created to solve the issue for crowd modelling 2d game.

PR #649, opened 2021-06-28
JAX DeepCFR implementation

StochasticEntropy:
JAX implementation of DeepCFR, reusing most of the TF2 implementation code.

The code achieves NashConv of 0.3 after 100 iterations with 1500 traversals, at leduc which seems ok. It seems to converge slower when tested against the TF2 version in the beginning, but I'll run a few more tests.


lanctot:
This is an awesome addition, thanks @StochasticEntropy ! 

Two first things:

1. Can you resolve the current conflicts with master?
2. Can you show me a graph on Leduc up to 300 iterations? Given the problems we've had with our Deep CFR implementations I'd like to first make sure we can at least reproduce the well-known benchmarks. You can use Figure 3 of this paper as a guide to how it should look: https://arxiv.org/abs/2103.00187

Thanks!

lanctot:
@StochasticEntropy that NFSP jax tests was re-added back last week. Can you try resolving the conflict via the github page? It will merge the changes from master in with it afterward.

lanctot:
Hmm, this didn't work .. possibly easier to copy your JAX impl and start a new PR from the current master.

Issue #602, opened 2021-05-17
JAX NFSP tests failing (only sometimes)


lanctot:
Happened twice today. 

Here's an example: https://github.com/deepmind/open_spiel/runs/2603201182?check_suite_focus=true

@perolat can you take a look?

StochasticEntropy:
setting a np.random.seed in the test should probably fix that to not get unlucky reservoir buffer sampling in the test

PR #651, opened 2021-06-29
fixed random seed in nfsp_jax_test (Issue #602)

StochasticEntropy:
added fixed np-seed to reservoir buffer sampling in nfsp_jax_test to avoid occasional unlucky sampling as seen in #602


lanctot:
Ah thanks @StochasticEntropy, @perolat recently removed these stochastics tests instead, we just have not had to to sync to github yet. Appreciate the fix, though! (And thanks for Deep CFR in JAX!)

lanctot:
Thanks again for the fix, but we removed these tests in this commit: https://github.com/deepmind/open_spiel/commit/34560a28c1ba3f46b8adf19771782b6e89bdc432

PR #658, opened 2021-07-05
Use basic tests random sim for crowd modelling test C++

TheoCabannes:
This is done after working on https://github.com/deepmind/open_spiel/pull/648
As the previous PR might found a lot of bugs in python code some code in C++ can already be submitted


lanctot:
Ok thanks, I will ask @perolat to take a look

PR #659, opened 2021-07-05
Add MaxChanceNodesInHistory in python game binding

TheoCabannes:
This is done after working on #648.
This enables using random_sim in python mfgs tests. Because of other issues in python mfgs tests, I am only submitting this PR for python mfgs.

@lanctot let me know if you prefer keep one bigger PR (#648) instead of many small PRs.

PR #660, opened 2021-07-05
Add binding and test for one population mfg using random sim

TheoCabannes:
This PR already contains changed in https://github.com/deepmind/open_spiel/pull/659 and https://github.com/deepmind/open_spiel/pull/658.
It was originally based on https://github.com/deepmind/open_spiel/pull/648 but does not fix bugs for crowd modelling 2d C++ game neither for predator pray python game (only multipopulation mfg so far in open spiel).


lanctot:
Ok, I updated github just now. Can you the conflicts? I assume they will easy enough to do straight via github itself

TheoCabannes:
Conflicts have been resolved @lanctot . Thank you a lot!

PR #662, opened 2021-07-06
Use random simulation for crowd modelling 2d test

TheoCabannes:
This PR is failing due to an issue on crowd modelling 2d implementation. I do not know what the issue is but I am creating it in order to solve it. This PR is the child of the closed PR https://github.com/deepmind/open_spiel/pull/648

### DIFF
--- a/open_spiel/games/mfg/crowd_modelling_2d.h
+++ b/open_spiel/games/mfg/crowd_modelling_2d.h
@@ -167,7 +167,7 @@ class CrowdModelling2dGame : public Game {
   }
   std::vector<int> ObservationTensorShape() const override;
   int MaxChanceOutcomes() const override {
-    return std::max(size_, kNumChanceActions);
+    return std::max(size_ * size_, kNumChanceActions);
   }
   std::unique_ptr<State> DeserializeState(
       const std::string& str) const override;
--- a/open_spiel/games/mfg/crowd_modelling_2d_test.cc
+++ b/open_spiel/games/mfg/crowd_modelling_2d_test.cc
@@ -56,43 +56,8 @@ void TestLoadWithParams2() {
 }
 
 void TestRandomPlay() {
-  std::mt19937 rng(7);
-  // TODO(author15): Should we adapt and use testing::RandomSimTest instead?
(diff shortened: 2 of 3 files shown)


lanctot:
Internally, I am getting the same error as you were getting with #648:

```
ChanceOutcomesTest, game = mfg_crowd_modelling_2d

RandomSimTest, game = mfg_crowd_modelling_2d, num_sims = 3
Information state vector size: 0
Observation vector size: 41
Utility range: -inf inf
Starting new game..
Initial state:
State:
initial
player -1
Spiel Fatal Error: third_party/open_spiel/tests/basic_tests.cc:104 action < expected_length
action = 10, expected_length = 10
```

TheoCabannes:
Quoting @RaphaelMarinier comment on the previous PR #648:

> Regarding crowd_modelling_2d_test, I have a high suspicion that this is caused by a bug when computing the observation tensor. Those bugs were fixed in the 1D game [here](https://github.com/deepmind/open_spiel/commit/052d88ca99063834cf8fe70bec3a1bdaa2d5ed45), but are still present in the 2D version. I just fixed them internally.

lanctot:
Hmm, it is failing here: https://github.com/deepmind/open_spiel/blob/c81400755106328ad25a522a9d9bc79f4e3501a7/open_spiel/tests/basic_tests.cc#L118

That means there is an action in the legal actions that is larger than MaxChanceOutcomes (if it's a chance node) or NumDistinctActions (if it's a decision node)

TheoCabannes:
I wonder if MaxChanceOutcomes should be size*size:
https://github.com/deepmind/open_spiel/blob/c81400755106328ad25a522a9d9bc79f4e3501a7/open_spiel/games/mfg/crowd_modelling_2d.h#L170

TheoCabannes:
@lanctot using size * size for MaxChanceOutcomes solve the issue

lanctot:
> @lanctot using size * size for MaxChanceOutcomes solve the issue

Great! Let's confirm with @perolat that it is correct, and then I will continue with the import.

lanctot:
Merged in https://github.com/deepmind/open_spiel/commit/769b159da3edded0558a8585f9ef8542577609bb

PR #663, opened 2021-07-06
JAX DeepCFR implementation

StochasticEntropy:
JAX implementation of DeepCFR, reusing most of the TF2 implementation code.

### DIFF
--- a/open_spiel/python/CMakeLists.txt
+++ b/open_spiel/python/CMakeLists.txt
@@ -194,7 +194,14 @@ if (OPEN_SPIEL_ENABLE_JAX)
   # Only current JAX test is the bridge supervised learning example below.
   set (PYTHON_TESTS ${PYTHON_TESTS}
     jax/dqn_jax_test.py
+<<<<<<< HEAD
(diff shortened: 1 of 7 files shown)


lanctot:
Great! What do the exploitability values look like for Leduc?

StochasticEntropy:
NashConv for Leduc over 500 iterations

deep_cfr_jax_nashconv

lanctot:
That looks good! Thanks.

StochasticEntropy:
The current open spiel optax version doesn't have the l2 loss defined yet. If that can be updated I can use the optax.l2loss, currently I just copied  their implementation so that it will be consistent later.

lanctot:
I woukd rather do the cleaner thing now. Can you try replacing the version you need here? https://github.com/deepmind/open_spiel/blob/c81400755106328ad25a522a9d9bc79f4e3501a7/open_spiel/scripts/python_extra_deps.sh#L27

StochasticEntropy:
done, optax.l2loss works now.

PR #666, opened 2021-07-07
Changes to C++ DQN

lanctot:
@Asugawara I started this PR in order to start looking at the C++ DQN implementation and the return graphs on catch, and to discuss some of the changes.

I'm not an expert on Libtorch so some of these will be be a shot in the dark. Going to tag @michalsustr for some help, and @jhtschultz because I will cite some graphs from his experiments.

I also plan to add a self-play example with a multiplayer game, probably Tic-Tac-Toe.


lanctot:
@Asugawara I already made some changes to how the illegal actions are chosen in `DQN::EpsilonGreedy`, can you take a look?

michalsustr:
I'm tagging @mwalton in this devillish PR :) since he wanted to run some DQN experiments as well

Asugawara:
HI! Dr. @lanctot
I appreciate for opening this PR!
I took a look at the change.  Is the purpose to not raise type error before one realizes it?
I think it seems better than before!

Will other changes be added in the future?

lanctot:
Yes, absolutely. I just started the PR early.

I want to find the source of the main bug so that it converges on catch as we expect from the known graphs. I want to mention it in the release and so want to be somewhat confident that it is fully correct.

This one was mainly because `torch::empty` seems to fill the tensor with arbitrary values (not necessarily zeroes) and the I believe with the multiply by 1s you were assuming that there had to be at least one positive q-value, so it does not work for all negative q values, because then an illegal move would have the largest q value of zero.

But it does not fix the catch convergence so there are definitely more problems.

Asugawara:
>I want to find the source of the main bug so that it converges on catch as we expect from the known graphs. I also want to make sure it works for the multiplayer case. I want to mention C++ DQN in the release and so I would like us to be somewhat confident that it is fully correct.

I'll get serious to research the bugs and fix them as soon as I can.

>This one was mainly because, at least from the docs, torch::empty seems to fill the tensor with arbitrary values (not necessarily zeroes)

I completely understand the reason for the changes.
Thank you for confirming the docs. 

> it does not work if all the q values are negative, because then an illegal move would have the largest q value of zero.

Indeed, I had assumed that situation. It's my mistake!

lanctot:
No rush. Earliest we can release is mid August. I started the PR mostly to make it easier to collaborate on fixing it and discussing. If you want you can checkout this branch (maybe there is a way that you could also push commits to it, if I give you access?)

lanctot:
As promised, just posting the results of John's runs of our Python DQN on C++ Catch:

image

So as we can see in most cases it gets to >= 0.5 by 100k episodes.

Asugawara:
>If you want you can checkout this branch (maybe there is a way that you could also push commits to it, if I give you access?)

When I would find a problem, I'd be happy to give me access to this branch.

One thing I noticed is that q_value in Learn method suddenly becomes Nan. 
That's probably a bad thing for the catch convergence. I'll investigate more deeply!

lanctot:
Ok thanks. I have not looked at `DQN::Learn` yet. I think there are still a few things in `Step` that we need to change. I will send a commit soon.

Also I think this line should be `prev_state`, right? https://github.com/deepmind/open_spiel/blob/ba5c4fbf605c8add8ebb2da745511805319e1b19/open_spiel/algorithms/dqn_torch/dqn.cc#L124

Asugawara:
Thank you!

>Also I think this line should be prev_state, right?

I don't think so ...
TF DQN uses `time_step`, not `prev_step`
https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/dqn.py#L292

lanctot:
Oh yes, you are right.. my mistake. They are needed for the max over legal actions at the next state.

Asugawara:
Hi! Dr. @lanctot ! 
There are good news and bad news! :)

The good news is, I found the cause of to not converge.
①This boolean is supposed to be `true` (in .h file). But now is `false`.
https://github.com/deepmind/open_spiel/blob/6dd7f4ae221441bcb63bbe28a3a6b91062eba534/open_spiel/algorithms/dqn_torch/simple_nets.cc#L32

ref:

- TF
  - https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/simple_nets.py#L33
 - PyTorch
   - https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/pytorch/dqn.py#L98

②`action` doesn't update in example.
`agent->Step(*state, true);` => `action = agent->Step(*state, true);`
https://github.com/deepmind/open_spiel/blob/6dd7f4ae221441bcb63bbe28a3a6b91062eba534/open_spiel/examples/dqn_torch_example.cc#L44

The bad news is, it is to converge much faster than jhtschultz's experiment if you fix ① and ②.
Is it something wrong??
スクリーンショット 2021-07-10 22 51 46

lanctot:
Great work!

For the fast convergence, might be explained by hyperparams or size of the nets? When I imported yoyr PR I changed the values to something I found in our Python tests (or example) which was designed for catch, which had quite a highe learning rate IIRC.

Maybe we can try running the python ones again to see. @jhtschultz do you remember the hyperparams you used?

jhtschultz:
Great to see it's converging! My experiments always used the default DQN params.

Asugawara:
>When I imported your PR I changed the values to something I found in our Python tests (or example) which was designed for catch, which had quite a high learning rate IIRC.

Yes!
So. The params I used are in this branch(lanctot_cppdqn3).
https://github.com/deepmind/open_spiel/blob/lanctot_cppdqn3/open_spiel/examples/dqn_torch_example.cc#L61

lanctot:
Ok that makes sense then, I used these: https://github.com/deepmind/open_spiel/blob/ba5c4fbf605c8add8ebb2da745511805319e1b19/open_spiel/python/examples/single_agent_catch.py#L76

So eventually we should probably make the defaults consistent with the python/pytorch dqn and try those. Thanks, great progress.

lanctot:
Hi @Asugawara, I have applied the changes you described in this commit: https://github.com/deepmind/open_spiel/pull/666/commits/54ffb443d47aebdf7945e8e7b5b8af295da687d4

But, it seems like it's not enough:
```
examples/dqn_torch_example --max_iterations=1000000 --eval_every=1000
0 -1
1000 -0.58
2000 -0.54
3000 -0.58
4000 -0.66
5000 -0.6
6000 -0.54
7000 -0.7
8000 -0.54
9000 -0.66
10000 -0.68
11000 -0.56
12000 -0.58
13000 -0.52
14000 -0.38
15000 -0.58
16000 -0.52
17000 -0.5
18000 -0.64
19000 -0.68
20000 -0.46
21000 -0.6
22000 -0.62
```

Was there another thing that was missing, maybe? Can you open a different PR with your version of the files? That way we can see all the changes to your local files.

Asugawara:
Oh! Sorry!
Solving to not converge needed the part I thought would be unaffected...

Please check it.
https://github.com/deepmind/open_spiel/pull/670/files

lanctot:
Yep, with those changes to `dqn.cc`, it works: 

```
0 -1
1000 0.64
2000 1
3000 1
4000 1
```

Nice!

Asugawara:
Yes!!!!!!
Thanks!!!!!

lanctot:
> Yes!!!!!!
> Thanks!!!!!

No problem... but especially thank _you_! This was the last thing I wanted to clear up before feeling comfortable with the release. It will be quite a cool thing to include in the release notes. I don't know many C++ DQN implementations! 

I will add a multiplayer self-play example (e.g. Tic-Tac-Toe) this week and will then import this PR once that's working. Thanks again!!

Asugawara:
I agree!
If you would enhance C++ implementations,  OpenSpiel will be the only one!! 
I'm looking forward to the next release! :)

>I will add a multiplayer self-play example (e.g. Tic-Tac-Toe) this week and will then import this PR once that's working. 

If you find anything strange, feel free to contact me.

lanctot:
Ok all, it's done. Debugging This PR I found a few minor bugs across the library. Perfectly in time to fix for the release. ;-) 

BTW @michalsustr, can you take a look at https://github.com/deepmind/open_spiel/pull/666/commits/21056d7988f92932407c2462dcbfb6ce359916d4 ? I added observation strings and tensors to matrix games. While fixing the existing games, I noticed that the classes in `coordinated_mp` did not inherit from MatrixGame + MatrixState (or NormalFormGame / NormalFormState).. was there a reason for that?

@Asugawara, can you take a look and let me know what you think? I think it's quite ready for importing now.

Asugawara:
That’s amazing!!
Passing seed as an argument is what I wanted when I experimented!!

PR #670, opened 2021-07-11
For checking the changes to solve to not converge

Asugawara:
for https://github.com/deepmind/open_spiel/pull/666


lanctot:
Thanks @Asugawara, got it working in #666, so going to close this and continue in that PR.

PR #677, opened 2021-07-23
Enable simultaneous game and python benchmarks in python examples

TheoCabannes:


open_spiel/python/examples/benchmark_games.py
```diff
@@ -57,8 +59,15 @@ def _rollout_until_timeout(game_name, time_limit, give_up_after):
...
         num_giveups += 1
         break
-      action = random.choice(state.legal_actions(state.current_player()))
-      state.apply_action(action)
+      if not state.is_simultaneous_node():
+        action = random.choice(state.legal_actions(state.current_player()))
+        state.apply_action(action)
+      else:
+        actions = [
+          random.choice(state.legal_actions(i))
```
RaphaelMarinier, line 67: Should we handle the case where there is no legal action? (Given https://github.com/deepmind/open_spiel/blob/e3a597b1d5e34a3130f36fa833702a1ae720a83f/open_spiel/spiel.h#L568)

open_spiel/python/examples/benchmark_games.py
```diff
@@ -57,8 +59,15 @@ def _rollout_until_timeout(game_name, time_limit, give_up_after):
       if len(state.history()) > give_up_after:
         num_giveups += 1
         break
-      action = random.choice(state.legal_actions(state.current_player()))
-      state.apply_action(action)
+      if not state.is_simultaneous_node():
```
RaphaelMarinier, line 62: Nit: It's slightly easier to read when conditions are in positive form.
I would do:
if state.is_simulateneuous_node():
  ...
else:
 ...

Otherwise, the else branch is a double negation, which is harder to read than what I propose.

open_spiel/python/examples/benchmark_games.py
```diff
@@ -22,6 +22,8 @@
 from absl import logging
 import pandas as pd
 
+from open_spiel.python import games  # pylint: disable=unused-import
+from open_spiel.python.mfg import games as mfg_games  # pylint: disable=unused-import
```
RaphaelMarinier, line 26: You will need to handle mean field nodes for MFGs to work in the benchmark, but this can wait another pull request.

RaphaelMarinier (changes requested):

open_spiel/python/examples/benchmark_games.py
perolat, line 26: Does this run on any MFG? There seem to be no code to deal with the distribution.

TheoCabannes: Good point. It cannot run on MFG yet.

open_spiel/python/examples/benchmark_games.py
perolat, line 62: A simpler way to handle this is to load the game as turn based if it is simultaneous. Note that most algorithm (DQN, A2C) deal with simultaneous move games this way.

TheoCabannes: I know. Should I add a flag (`convert_simultaneous_game_to_turn_based`) to do that?

As this module is done for benchmarking, I think it is still useful to benchmark simultaneous games (as the benchmark might help finding where the code can be optimized)

open_spiel/python/examples/benchmark_games.py
perolat: Maybe raise an error or put a TODO if the game is mean field. Or if you are not under time pressure, adding it is a very nice contribution :)

open_spiel/python/examples/benchmark_games.py
perolat: I'd load it as turn based if it is simultaneous. But as if you prefer to handle the simultaneous case, it seems fine by me (@lanctot do you agree?).

lanctot: I'm fine either way, I see the benefit of both.

open_spiel/python/examples/benchmark_games.py
TheoCabannes: Done.

open_spiel/python/examples/benchmark_games.py
TheoCabannes: Done: added a flag.

TheoCabannes: Done.

open_spiel/python/examples/benchmark_games.py
TheoCabannes: Add NotImplementedError

TheoCabannes: See remark from Julien just below.

perolat (approved):
Look good to me.

lanctot (approved):
Looks good, thanks!

Issue #679, opened 2021-07-26
Bug in RBC?


lanctot:
@michalsustr 

We are seeing memory issues with RBC, in the first test where there are two successive pass moves. You can reproduce if you compile in debug mode and run rbc_test under valgrind.

I think it is caused by some checks or functions which should not apply in the case of pass moves.  Also, `illegal_move_attempted_` was being immediately set back to false after being set to true; that does not seem right. 

To the best of my understanding, I changed the `DoApplyAction` to the following:

```
void RbcState::DoApplyAction(Action action) {
  if (phase_ == MovePhase::kSensing) {
    sense_location_[CurrentPlayer()] = action;
    phase_ = MovePhase::kMoving;
    illegal_move_attempted_ = false;
    move_captured_ = false;
  } else {
    chess::Move move = chess::kPassMove;
    illegal_move_attempted_ = false;
    move_captured_ = false;

    // Handle special cases for RBC.
    if (action == chess::kPassAction) {
      move = chess::kPassMove;
    } else if (Board().IsBreachingMove(move)) {
      move = ActionToMove(action, Board());
      SPIEL_DCHECK_FALSE(Board().IsMoveLegal(move));
      Board().BreachingMoveToCaptureMove(&move);
      // Transformed move must be legal.
      SPIEL_DCHECK_TRUE(Board().IsMoveLegal(move));
      // And it must be a capture.
      SPIEL_DCHECK_NE(Board().at(move.from).color, Board().at(move.to).color);
      move_captured_ = true;
    } else if (!Board().IsMoveLegal(move)) {
      illegal_move_attempted_ = true;
      move = chess::kPassMove;
    } else {
      // All other moves
      SPIEL_DCHECK_NE(Board().at(move.from).color, chess::Color::kEmpty);
      move_captured_ = Board().at(move.from).color != Board().at(move.to).color;
    }

    SPIEL_DCHECK_TRUE(move == chess::kPassMove || Board().IsMoveLegal(move));
    moves_history_.push_back(move);
    Board().ApplyMove(move);

    ++repetitions_[current_board_.HashValue()];
    phase_ = MovePhase::kSensing;
  }
  cached_legal_actions_.reset();
}
```

.. but now it is causing a crash on the first step of the random simulations in ChessBoard::IsBreachingMove:

```
chess_board.cc:659 CHECK_TRUE(piece.type == PieceType::kQueen || piece.type == PieceType::kRook || piece.type == PieceType::kBishop)

[...]

PR #680, opened 2021-07-27
Update RBC based on #679

michalsustr:
Possibly a memory leak, wrong illegal_move_attempted_ flag setting, simplify pass action check.

### DIFF
--- a/open_spiel/games/rbc.cc
+++ b/open_spiel/games/rbc.cc
@@ -275,7 +275,9 @@ RbcState::RbcState(std::shared_ptr<const Game> game, int board_size,
                    const std::string& fen)
     : State(game),
       start_board_(*chess::ChessBoard::BoardFromFEN(
-          fen, board_size, /*king_in_check_allowed=*/true)),
+          fen, board_size,
+          /*king_in_check_allowed=*/true,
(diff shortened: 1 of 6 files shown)


open_spiel/games/rbc.cc
```diff
@@ -275,43 +275,46 @@ RbcState::RbcState(std::shared_ptr<const Game> game, int board_size,
...
       SPIEL_DCHECK_NE(Board().at(move.from).color, Board().at(move.to).color);
+      move_captured_ = true;
     } else if (!Board().IsMoveLegal(move)) {
       illegal_move_attempted_ = true;
-      move = chess::kPassMove;
+      move = chess::kPassMove;  // Illegal move was chosen, so pass instead.
+    } else {
+      // All other moves
+      SPIEL_DCHECK_NE(Board().at(move.from).color, chess::Color::kEmpty);
+      move_captured_ = Board().at(move.from).color != Board().at(move.to).color;
```
lanctot, line 314: The problem was here. On a pass move, the move.from and move.to here were both -1, so the index into the array was failing. I'm surprised valgrind did not find that, maybe you need to have a flag for checking array bounds. Anyway I added a special case for the pass move and now it's fine.

Issue #681, opened 2021-07-28
UCIBot basic test is not working


lanctot:
@vofak 

Upon import of https://github.com/deepmind/open_spiel/pull/496, I added a basic test that implements a simple UCI chess program that takes random moves. Unfortunately it seems to hang and not work. So I have marked the bot as experimental and opening this issue to discuss it.

lanctot:
You can see the test I added in this merge commit: https://github.com/deepmind/open_spiel/commit/a432cf7441acc08479b868ffc2cde208464d403e

@michalsustr did you also work on this? Or do you have contact with @vofak.. ? I have not heard from them in some time. It would be good to get a basic test in place to ensure the UCIBot works.

lanctot:
Fixed by https://github.com/deepmind/open_spiel/commit/ba9220ecd0225ba558f79853b0c3a877bcd7ce10

PR #683, opened 2021-07-29
Fix issue: UCIBot basic test is not working #681

michalsustr:
Fix for issue #681 -- now the test passes. I ran the test successfully with a `stockfish` binary as well. 

The test wasn't passing because the random bot was not parsing absl flags, so the supplied bot file was not found. There was no error reporting about this, so I added one. Also, parsing of the UCI commands was not correct and I fixed it.

I also looked at the UCI bot code and made some other (cosmetic) changes. The code could be improved, perhaps in another PR. If I understand it correctly, reading from the pipe is done in a busy loop. There are no checks for the bot timeouts.


lanctot:
Amazing! I got it working internally as well. Thanks!

Issue #685, opened 2021-08-02
Saving trained tabular qlearner


EmmaUnimelb:
Is there a method implemented in openSpiel for saving a trained tabular qlearner and restoring it for later use? The existence of a lambda seems to stop it from pickling at the moment

EmmaUnimelb:
Yes, I believe that's the one.

The error is:
AttributeError: Can't pickle local object 'QLearner.__init__.<locals>.<lambda>'

EmmaUnimelb:
I tried your suggestion and it worked - thanks

lanctot:
Great! Would you like to submit a fix?

EmmaUnimelb:
Sure.
I have sent a pull request. Github is having a little trouble showing me the CLA for the right email though

lanctot:
> Sure.
> I have sent a pull request. Github is having a little trouble showing me the CLA for the right email though

Ah yes I think you probably have to logout of your current account or access github while signed into your account associated with your github account.

lanctot:
Solved by https://github.com/deepmind/open_spiel/pull/690, thanks!

PR #687, opened 2021-08-02
remove lambda from tabular_qlearner issue #685

EmmaUnimelb:


EmmaUnimelb:
Having trouble signing the CLA due to my workplace's SSO. Is there another way of verifying myself as a contributor?

lanctot:
> Having trouble signing the CLA due to my workplace's SSO. Is there another way of verifying myself as a contributor?

Ah, the only ways are listed in that google-cla comment above (does the Corporate signers section apply to you?) I've heard of issues if you're logged into an account that is not your github email, so might work if you log out of your one account and login to your github-associated one (assuming you're signing as an individual signer). 

I suggested the PR because I wanted you to get public credit for the find & fix. If it's too much trouble, I can fix it on my end and will mention you and your original issue in the commit. Let me know what you prefer.

EmmaUnimelb:
That's very kind of you. I think what I'll do is to close this and submit it again from my non-work github account which hopefully doesn't have the same issues. If that doesn't work then we can go with plan B
Thanks!

PR #693, opened 2021-08-10
HIGC tournament implementation

michalsustr:
This PR adds the HIGC tournament implementation as discussed with @lanctot


lanctot:
Hey, don't use std::filesystem. It has compatibility issues on older compilers which we need to build the wheel. There should be something from utils/file you can use instead.

michalsustr:
>  There should be something from utils/file you can use instead.

Thanks for the suggestion! I updated the code. I still need to check if it runs properly in Travis, there might be an issue with relative paths when I try to get the correct python executable via `venv` that has pyspiel in it.

michalsustr:
Hmm not sure why `build_and_test / build (ubuntu-18.04, 3.6, 0, OFF, OFF, OFF)` is failing, while the other ones are working. The error message complains about non-existence of a file that is included in the PR (and this runs fine in the other checks).

```

Bot#0: /home/runner/work/open_spiel/open_spiel/open_spiel/higc/bots/test_bot_buffer_overflow.sh  <-- should exist
Bot#0: subprocess: execvpfailed at /home/runner/work/open_spiel/open_spiel/open_spiel/../open_spiel/higc/subprocess.h:96: No such file or directory <--- but claims it doesn't :/
```

lanctot:
Maybe a fluke? I've seen sporadic trouble with relative paths and files before. I've restarted the tests. If it happens again, it's probably legitimate.

lanctot:
If you didn't do this, probably the best thing is to pass the paths to these scripts via the command-line. Because then in the Makefile you can add specific arguments and use the CMake environment vars to point to the specific file. See how I did this for the UCI bot test here: https://github.com/deepmind/open_spiel/blob/f63bd349470f65cc521b14f324bfcbd59c109e32/open_spiel/bots/uci/CMakeLists.txt#L7

michalsustr:
I suspect the problem is in not passing PYTHONPATH correctly... I will have a look tomorrow as well.

michalsustr:
@lanctot Yay! It works now. I also exposed bindings to call the tournament and process the results from Python.

open_spiel/higc/base64.cc
```diff
@@ -0,0 +1,132 @@
+/*
```
elkhrt, line 1: There's some process around us importing third party code which I'd rather avoid.
Would you be able to use absl::Base64Escape instead?

elkhrt (changes requested):

open_spiel/higc/base64.cc
michalsustr: Yes, I updated the code to use absl implementation.

elkhrt (approved):

lanctot:
Alright I have now imported this. I had to make a few changes due to C++ style / conventions. The tests still run and work, but most notably I had to remove exceptions and add a `Referee::StartedSuccessfully` function, which will return false if the subprocesses fail to open. So now there's an extra step of `SPIEL_CHECK_TRUE(ref.StartedSuccessfully())` in a number of places.. when you see the code just verify that I got them all.

michalsustr:
I will take a look to verify. Thank you Marc!

PR #699, opened 2021-09-08
Add policy value distribution computation

TheoCabannes:
Currently the policy value modules only computes the expected return of a player in the mean field game. However, one can also want to get the distribution of the returns of a player and not only the expected return. This should enable doing chance programming on the distribution of returns (for example minimize expected return under the constraints that at least 90% of the returns should be below a given value).

### DIFF
--- a/open_spiel/python/mfg/algorithms/policy_value_distribution.py
+++ b/open_spiel/python/mfg/algorithms/policy_value_distribution.py
@@ -1,0 +1,153 @@
+# Copyright 2019 DeepMind Technologies Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sgirgin (commented):
From an abstract point of view, the code looks good to me (a generalization of the policy value algorithm), but my main concern would be that now we are using floating point numbers as keys of a dictionary. This is often quite fragile due to numerical accuracy and should be avoided. You may easily get duplicate entries for seemingly the same reward.

If you need this algorithm, my suggestion would be to add a post-processing step that merges the entries that are close to each other (i.e. < epsilon). We can also discuss offline.

TheoCabannes:
Thank you Sertan for the remark. I agree that float should not be keys of dictionary. I like the idea of using bins for the dictionary keys. I will update the PR to use predefined bins of length epsilon.

PR #700, opened 2021-09-08
Fix nit in noisy_policy

TheoCabannes:
If player_id is 0 then not player_id is True, however we only want not player_id to be True in the case of player_id = None. Therefore, I have changed this to test player_id is None.

I have updated the tests. I guess the test in the test is due to stochasticity of the output of the test (b/141737795).

I wonder if noisy_policy has been ever used, it seems that even this initialization is not correct: https://github.com/deepmind/open_spiel/blob/38941dee3beb52ffdb134b66f420a758634d9a20/open_spiel/python/algorithms/best_response.py#L318

### DIFF
--- a/open_spiel/python/algorithms/noisy_policy.py
+++ b/open_spiel/python/algorithms/noisy_policy.py
@@ -113,11 +113,11 @@ class NoisyPolicy(openspiel_policy.Policy):
 
     # If self._player_id is None, or if self.player_id == current_player, add
     # noise.
-    if (not self.player_id) or (state.current_player() == self.player_id):
+    if (self.player_id is None) or (state.current_player() == self.player_id):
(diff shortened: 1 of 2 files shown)


TheoCabannes:
Btw, the name of the noisy policy test does not seem correct.

lanctot:
> I do not understand why I got some error for build wheels on macOS-10.15.
> Can you help me understand this error?

I did not see any error logs, so I could not tell what the problem was. I restarted the tests. Maybe it was just a fluke?

lanctot:
I am not sure who wrote this code originally nor where it is used. When I find out, i will tag them or contact them.

PaulFMMuller:
Thanks Theo, looks good & good find :)

PR #701, opened 2021-09-08
Implement expected game score for simultaneous game

TheoCabannes:
In order to test the function I have slightly modified iterated prisoners dilemma to have a max length (to avoid infinite recursion).

### DIFF
--- a/open_spiel/python/algorithms/expected_game_score.py
+++ b/open_spiel/python/algorithms/expected_game_score.py
@@ -18,12 +18,18 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import itertools
+from typing import List, Union
+
 import numpy as np
+
+from open_spiel.python import policy
+
 
 PROBABILITY_THRESHOLD = 0
 
 
-def policy_value(state, policies):
+def policy_value(state, policies: Union[List[policy.Policy], policy.Policy]):
   """Returns the expected values for the state for players following `policies`.
 
   Computes the expected value of the`state` for each player, assuming player `i`
@@ -31,14 +37,13 @@ def policy_value(state, policies):
 
   Args:
     state: A `pyspiel.State`.
-    policies: A `list` of `policy.Policy` objects, one per player.
+    policies: A `list` of `policy.Policy` objects, one per player for sequential
+      games, one policy for simulatenous games.
 
   Returns:
     A `numpy.array` containing the expected value for each player.
   """
-  assert not state.is_simultaneous_node()
-
-  num_players = len(policies)
+  num_players = state.get_game().num_players()
 
   if state.is_terminal():
     values = np.array(state.returns())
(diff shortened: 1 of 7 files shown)


open_spiel/python/algorithms/expected_game_score.py
```diff
@@ -47,10 +52,22 @@ def policy_value(state, policies):
...
-        "Policy Value is not implemented for simultaneous games")
+    values = np.zeros(shape=num_players)
+    action_prob_dicts = []
+    for player in range(num_players):
+      action_prob_dicts.append(
+        policies.action_probabilities(state, player))
+    actions = itertools.product(*action_prob_dicts)
+    probabilities = itertools.product(
+      *[action_prob.values() for action_prob in action_prob_dicts])
+    for action_tuple, probability_tuple in zip(actions, probabilities):
```
elkhrt, line 64: It would be clearer to add a helper function that produced a list of [(actions), prob] or something like it for a policy in a simultaneous move game. That would make this code simpler, and also the helper function could be independently tested.

open_spiel/python/algorithms/expected_game_score_test.py
```diff
@@ -35,6 +36,16 @@ def test_expected_game_score_uniform_random_kuhn_poker(self):
...
 
+  def test_expected_game_score_uniform_random_iterated_prisoner_dilemma(self):
+    game = pyspiel.load_game(
+      "python_iterated_prisoners_dilemma(max_game_length=6)")
+    uniform_policy = policy.UniformRandomPolicy(game)
+    uniform_policy_values = expected_game_score.policy_value(
+        game.new_initial_state(), uniform_policy)
+    print(uniform_policy_values)
+    self.assertTrue(
+      np.allclose(uniform_policy_values, [19.43373108, 19.43373108]))
```
elkhrt, line 47: Can you explain these values?
I would have expected 4 per round * expected number of rounds, i.e. around 17.6
(perhaps the game is lasting longer than the specific maximum number of rounds?)

TheoCabannes: This is correct, the value should be around 17.6385498. I have changed the game length implementation (current time step at initialization is now 1).

open_spiel/python/algorithms/expected_game_score.py
TheoCabannes: Should I implement the function in C++ and add a python wrapper?
I agree that this would be easier.

TheoCabannes (commented):
Working on implementing functions for simultaneous games on another PR.
This would simplify the code a lot.

open_spiel/python/algorithms/expected_game_score.py
TheoCabannes: @elkhrt ?

elkhrt: Python is fine unless you have C++ use cases in mind.

elkhrt: If you don't need it from C++, then pure Python is fine too - whichever is easier.

TheoCabannes: Done in `policy.py`. Thanks!

PR #703, opened 2021-09-10
Enable computing best response in simultaneous games

TheoCabannes:
Test can be added, when the change in iterative prisoner dilemma will be merged (https://github.com/deepmind/open_spiel/pull/701). I can also try to test with simultaneous game coded in C++

### DIFF
--- a/open_spiel/python/algorithms/best_response.py
+++ b/open_spiel/python/algorithms/best_response.py
@@ -115,11 +115,11 @@ class BestResponsePolicy(openspiel_policy.Policy):
   def decision_nodes(self, parent_state):
     """Yields a (state, cf_prob) pair for each descendant decision node."""
     if not parent_state.is_terminal():
-      if parent_state.current_player() == self._player_id:
+      if (parent_state.current_player() == self._player_id
+          or parent_state.is_simultaneous_node()):
         yield (parent_state, 1.0)
       for action, p_action in self.transitions(parent_state):
         if parent_state.is_simultaneous_node():
-          yield (parent_state, 1.0)
           child = parent_state.clone()
           child.apply_actions(action)
         else:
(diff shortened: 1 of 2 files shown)


open_spiel/python/algorithms/best_response.py
```diff
@@ -124,26 +133,63 @@ def transitions(self, state):
       # Counterfactual reach probabilities exclude the best-responder's actions,
       # hence return probability 1.0 for every action.
       return [(action, 1.0) for action in state.legal_actions()]
-    elif state.is_chance_node():
```
elkhrt, line 127: Could you retain the if /elif / elif / else structure?

open_spiel/python/algorithms/best_response.py
```diff
@@ -124,26 +133,63 @@ def transitions(self, state):
...
+    if state.is_chance_node():
       return state.chance_outcomes()
-    else:
-      return list(self._policy.action_probabilities(state).items())
+    if state.is_simultaneous_node():
+      action_prob_dicts = []
```
elkhrt, line 139: This block looks rather complex to be in-line. Coud it be extracted into one or more helper functions (which could be independently tested)?

open_spiel/python/algorithms/best_response.py
```diff
@@ -124,26 +133,63 @@ def transitions(self, state):
...
+        in list(zip(actions, probabilities))]
+    return list(self._policy.action_probabilities(state).items())
 
   @_memoize_method(key_fn=lambda state: state.history_str())
   def value(self, state):
     """Returns the value of the specified state to the best-responder."""
     if state.is_terminal():
       return state.player_return(self._player_id)
-    elif state.current_player() == self._player_id:
+    if state.current_player() == self._player_id:
```
elkhrt, line 161: Please retain the elif / else

open_spiel/python/algorithms/best_response.py
```diff
@@ -152,8 +198,12 @@ def best_response_action(self, infostate):
     infoset = self.infosets[infostate]
     # Get actions from the first (state, cf_prob) pair in the infoset list.
     # Return the best action by counterfactual-reach-weighted state-value.
+    def legal_actions_wrapper(state):
```
elkhrt, line 201: There's no need for this indirection. We just need a list of legal actions to pass to `max`.
I think state.legal_actions(self.player_id) would work in all cases, but if not we can construct the list here.

open_spiel/python/algorithms/best_response.py
```diff
@@ -124,26 +133,63 @@ def transitions(self, state):
...
+        # change action of _player_id
+        actions[self._player_id] = self.best_response_action(
+          state.information_state_string(self._player_id))
+        child.apply_actions(actions)
+        return self.value(child)
+      actions, probabilities = zip(*self.transitions(state))
+      return sum(p * q_value_sim(state, a)
+                 for a, p in zip(actions, probabilities/sum(probabilities))
                  if p > self._cut_threshold)
+    return sum(p * self.q_value(state, a)
```
elkhrt, line 177: Please keep the else: clause

open_spiel/python/algorithms/best_response.py
```diff
@@ -124,26 +133,63 @@ def transitions(self, state):
...
-    elif state.current_player() == self._player_id:
+    if state.current_player() == self._player_id:
       action = self.best_response_action(
-          state.information_state_string(self._player_id))
+        state.information_state_string(self._player_id))
       return self.q_value(state, action)
-    else:
-      return sum(p * self.q_value(state, a) for a, p in self.transitions(state)
+    if state.is_simultaneous_node():
+      def q_value_sim(state, actions):
```
elkhrt, line 166: Why can't q_value_sim be a member method like q_value?
Even better, why can't `q_value` handle both cases?

open_spiel/python/algorithms/best_response.py
```diff
@@ -115,7 +118,13 @@ def decision_nodes(self, parent_state):
       if parent_state.current_player() == self._player_id:
         yield (parent_state, 1.0)
       for action, p_action in self.transitions(parent_state):
-        for state, p_state in self.decision_nodes(parent_state.child(action)):
+        if parent_state.is_simultaneous_node():
+          yield (parent_state, 1.0)
```
elkhrt, line 122: This seems to imply yielding the parent state multiple times - is that intended?

open_spiel/python/algorithms/best_response.py
TheoCabannes: Done.

open_spiel/python/algorithms/best_response.py
TheoCabannes: Change to have `q_value` handle both cases. However, for simultaneous node we need to know the actions of all players and not only the player _player_id. In q_value, the action is the action of the player _player_id.

open_spiel/python/algorithms/best_response.py
TheoCabannes: I seems to me that in the case of sequential games, best_response_action can be called on decision nodes of other players. However, in the case of simultaneous games, we would like to do the max over actions of the player _player_id.

TheoCabannes (commented):
Thanks I will add tests using the C++ sim game before asking for a new review.

TheoCabannes:
Test has been added.

open_spiel/python/algorithms/best_response.py
TheoCabannes: Sorry, I understand this remark now. It has been changed.

elkhrt:
I think you might want to revisit this one #701 has been imported, so I'm going to leave it for now.

TheoCabannes:
Moving this PR to internal review

PR #705, opened 2021-09-16
Commit first version for armac, but not sure about its correctness.

dvstter:
I just implement the first version for armac. In fact, I am still not sure about its correctness. I hope anybody could help me review these codes. There are two major differences between the original implementation of the paper and this: 
1. I use the simplest Q-learning to replace the Treebackup. Because that is not quite so hard to modify, the foremost important question is that whether this code is right or not. 
2. I was just curious about the 'adaptive policy selection'. So there is not anything about this.

### DIFF
--- a/open_spiel/python/pytorch/armac.py
+++ b/open_spiel/python/pytorch/armac.py
@@ -1,0 +1,335 @@
+import collections
+import torch as T
+import torch.nn as nn
+import torch.optim as O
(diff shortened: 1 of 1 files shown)


dvstter:
@googlebot I signed it!

dvstter:
@googlebot I fixed it.

dvstter:
@googlebot I fixed it.

PR #707, opened 2021-09-16
Fix nit about distribution computation:

TheoCabannes:
Enable having two states evolving in the same state after the mean field node.

I find this issue by coding a MFG where the dynamics depend on the distribution.
My game is a mean field routing game where travel time on link depends on the distribution.

### DIFF
--- a/open_spiel/python/mfg/algorithms/distribution.py
+++ b/open_spiel/python/mfg/algorithms/distribution.py
@@ -109,10 +109,11 @@ def one_forward_step(current_states: List[pyspiel.State],
       ]
       new_state = state.clone()
       new_state.update_distribution(dist)
-      new_states.append(new_state)
-      if state_to_str(new_state) not in new_distribution:
-        new_distribution[state_to_str(new_state)] = 0.0
-      new_distribution[state_to_str(new_state)] += distribution.get(
+      new_state_str = state_to_str(new_state)
+      if new_state_str not in new_distribution:
+        new_states.append(new_state)
+        new_distribution[new_state_str] = 0.0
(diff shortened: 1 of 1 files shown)


open_spiel/python/mfg/algorithms/distribution.py
```diff
@@ -109,8 +109,11 @@ def one_forward_step(current_states: List[pyspiel.State],
       ]
       new_state = state.clone()
       new_state.update_distribution(dist)
-      new_states.append(new_state)
-      new_distribution[state_to_str(new_state)] = distribution.get(
+      new_state_str = state_to_str(new_state)
```
sgirgin, line 112: Should this be better handled by the caller Theo? That is you should ensure that the current states is unique in your game (and do the proper calculations, here you are doing an addition, so there is nothing preventing the value to exceed 1.0, right?).

TheoCabannes: I am not sure if this remark is about this line of code or about the whole PR.
We can add a check that the distribution does not exceed a value of 1.0. However the code is similar than https://github.com/deepmind/open_spiel/blob/5a5ba2a9254fe6f9bf0e762919aa1058606851f0/open_spiel/python/mfg/algorithms/distribution.py#L63

The goal of this PR is to solve the following issue:
the code is wrong in case two states (A and B) becomes a same state (C) after calling distribution_support and update_distribution.

In the current games implementation, as the dynamics does not depend on the mean field, calling update_distribution does not change the states. But in the mean field routing game I want to implement, the dynamics depends on the mean field and the current `distribution.py` module is not correct.

sgirgin (approved):
Discussed offline.

PR #713, opened 2021-09-30
Update TurnBasedSimultaneous game to use Observation API.

michalsustr:
This PR extends #712.


lanctot:
> Hi @TheoCabannes, when I debugged the code locally it was failing on the dynamic route games.
> 
> Can you change
> 
> https://github.com/deepmind/open_spiel/blob/9a27b78845c0d7c5a46c1a270da9af6dde3b3784/open_spiel/python/games/dynamic_routing_utils.py#L32
> to
> 
> ```
> import pyspiel
> NO_POSSIBLE_ACTION = pyspiel.INVALID_ACTION
> ```
> 
> and update other game code, as that is the correct way to encode that a non-participating player is not acting in a simultaneous-move game? I tried to fix this naively, but playthrough generation was failing afterwards with some not-easy-to-understand game-specific problem. Thank you!

Tagging @elkhrt.

Hi @michalsustr, I think we decided to use 0 for this case recently, like this past summer. There was a case that it let us handle which was easier to do with assuming 0 rather than INVALID_ACTION here. 

IIRC, it was two things: the conversion between flat actions and joint actions works seemlessly if you use 0 here rather than INVALID_ACTION. And now I'm forgetting the second reason. Theo, do you remember the PR or issue where we discussed it?

TheoCabannes:
The discussion was here: https://github.com/deepmind/open_spiel/issues/588
One issue was some checks that actions are `>=0`, while `INVALID_ACTION = -1`. 
@michalsustr I think that you will not only have playthrough generation issue by fixing `NO_POSSIBLE_ACTION = pyspiel.INVALID_ACTION` but also some pyspiel CHECK_FALSE issue.

Let me know if I can be helpful

TheoCabannes:
Btw, @michalsustr feel free to share how the code is failing, `NO_POSSIBLE_ACTION` for dynamic routing game should be seen as a valid action from an abstract point of view. But I am aware of some issues when trying to do api integration tests on python games (this is not done currently in the code, but should soon be updated).

Happy to help in any way, even though I am not sure to fully understand the issue here.

michalsustr:
I think [this comment](https://github.com/deepmind/open_spiel/issues/588#issuecomment-842543696) is relevant to the discussion -- i.e. the proper way to deal with this is to set kInvalidAction to non-acting players in sim-move nodes, and the turn-based game from sim game will skip over the non-acting players, filling in `kInvalidAction` in the `action_vector_`. I will take a look and update the PR next week... I use sim-move games quite a lot and the conversion to turn-based games is often a painful experience, so I'd like to "iron it out" as much as possible...

lanctot:
> I think [this comment](https://github.com/deepmind/open_spiel/issues/588#issuecomment-842543696) is relevant to the discussion -- i.e. the proper way to deal with this is to set kInvalidAction to non-acting players in sim-move nodes, and the turn-based game from sim game will skip over the non-acting players, filling in `kInvalidAction` in the `action_vector_`.

As Theo says, this will cause failures in other tests. That's why we recommended he use 0, and why (as I said before) we switched to assuming/requiring 0 for those cases. I think he also added checks to ensure that it's 0 (maybe that is the case that is failing) and also updated the docs, see https://github.com/deepmind/open_spiel/blob/ef9e24d1cef64af6949fa1f1b98d7c10d97bd5b7/open_spiel/spiel.h#L572

It also means the converter back and forth from joint actions to flat actions just works without special cases.

michalsustr:
Makes sense. I see three ways of handling the case when a subset of players doesn't act in sim-move nodes:

1) Use zeros as so far. 
2) Use `kInvalidAction` (i.e. -1)
3) Use a new constant (i.e. -2), maybe something called like `kNotActingAction`

I'd like to argue that using zeros is semantically ambiguous as it can clash with a domain-specific meaning ("Does this zero-action correspond to player not acting, or moving a pawn forward two cells?" kind of a thing).

Similarly, `kInvalidAction`, is, as the names suggests, invalid -- but having non-acting players is perfectly valid logic in some games such as the dynamic routing game. Therefore I would like to use the last option if you are fine with that.

I can update all the code to handle it properly. I already checked that requiring zero and positive actions passes all the tests, so making a new constant is fine.

michalsustr:
BTW seems like Eigen dependency https://gitlab.com/libeigen/eigen.git is down (and [it's known](https://eigen.tuxfamily.org/index.php?title=Main_Page)).

lanctot:
Thanks for the heads-up about Eigen, I've temporarily disabled the dependency in my sync yesterday.

I think I mostly agree. I have now forgotten why I recommended Theo to go with zeroes, but I my guess is that it was easier at the time and it'd involve changing too much code, and he was new to the code base. @TheoCabannes if you remember the argument differently, can you correct me?

Of your options I like the third one best, but I'd strongly prefer that `kNotActingAction` not share a value with the other special values (I think maybe -2 is already taken? if not, happy with it).

TheoCabannes:
> @TheoCabannes was making a similar argument during the summer and I convinced him to use zeros, I have forgotten why other than the convenience reasons I stated above (he was new to the code base and it'd involve significant changes). I think I mostly agree. @TheoCabannes if you remember the argument differently, can you correct me?

Yes that was for convenance as the change is quite significant, it needs to update code, tests and documentations.
And when I hit the issue that @michalsustr is currently fixing, I did not have much experience into OpenSpiel.

michalsustr:
> Of your options I like the third one best, but I'd strongly prefer that kNotActingAction not share a value with the other special values (I think maybe -2 is already taken? if not, happy with it).

It should be available.

> However, I am not sure the change is as easy as you think, so tread carefully :-p

No worries :) I believe there is large enough test coverage to catch all cases. (I have some more sim-move games tests on private fork.)

lanctot:
Just a quick status check on this PR: @michalsustr, are you still working on it? If so, you will let me know when it's ready?

michalsustr:
Sorry, I got a little busy. I have something half-done, but didn't have time to finish it as I run into some obstacles.

But I would like to  :) so if it's ok, can we keep this open?

lanctot:
Yes, no problem! I was just checking in.

lanctot:
Closing for now, please re-open if you want to continue working on this.

PR #714, opened 2021-10-06
Implemented restricted nash response wrapper

milecdav:
Implemented restricted nash response wrapper from [Johanson, Michael,… Michael Bowling, and Martin Zinkevich. "Computing robust counter-strategies." (2007).] It can be used as duplication of game tree with correct observations in which the strategy can be fixed using bandits or it can be used already with fixed strategy and it will convert fixed player nodes to chance nodes based on the strategy.


lanctot:
Update: I've fixed the crashing problem, I will submit a fix and update github.

However now I'm running into a different problem:
```
Unknown game 'rnr_blotto'. Available games are:
```

Then it lists all the games. I will open a new thread for this.

lanctot:
Ok the crashing bug is now fixed: https://github.com/deepmind/open_spiel/commit/eebcdf539a80a26062d73060f17ced5a0a31cc0a.

Issue #715, opened 2021-10-13
higc referee_test fails: No module named 'numpy'


VitamintK:
When attempting to install latest master open_spiel, I ran into the following failed test:
```
Starting players.
Bot#0: /Users/kevin/github/open_spiel_fork/open_spiel/higc/bots/random_bot_py.sh
Bot#1: /Users/kevin/github/open_spiel_fork/open_spiel/higc/bots/test_bot_start.sh
Bot#0: Traceback (most recent call last):
  File "/Users/kevin/github/open_spiel_fork/open_spiel/higc/bots/random_bot.py", line 19, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
Bot#0 did not respond 'ready'
Bot#0 response was: ''
Bot#0 also timed out.
Bot#1 ready ok.
Could not start all players correctly, cannot play the tournament.
Spiel Fatal Error: /Users/kevin/github/open_spiel_fork/open_spiel/higc/referee_test.cc:131 results->num_matches() == 1
results->num_matches() = 0, 1 = 1
```

I followed the steps
> ./install.sh
> pip3 install -r requirements.txt
> ./open_spiel/scripts/build_and_run_tests.sh

I have numpy installed on the `python` I'm using, as well as my system's python, so I'm not sure why this is happening.

I can disable the test to move forward, but I'm filing this issue here in case anyone else experiences this or knows how to fix it.

cc: @michalsustr

lanctot:
Another quick update: this is now causing the Hithub Actions wheels test to fail on MacOS X, so I'm disabling the higc_referee_test until we fix it, see https://github.com/deepmind/open_spiel/pull/719 for details.

I think we should fix this ASAP. @michalsustr, have you been able to reproduce it from a fresh checkout of master?

@VitamintK can you let me know what your environment is? It only started showing up for me after upgrading my Ubuntu to 21.10 and only on the wheels test yesterday (e.g. they were working fine last week).

VitamintK:
I'm using Mac OS with a conda (not venv) environment.

`import numpy` works both with the conda environment's `python` and `/usr/bin/python`.

VitamintK:
ah, but the tests were running with `open_spiel/venv/bin/python`, which I wasn't using, and which doesn't have `numpy`.  (Found the python directory by putting `print(os.path.dirname(sys.executable), file=sys.stderr)` in `random_bot.py`.)

[...]

PR #723, opened 2021-10-18
Referee update: Explicit is better than implicit.

michalsustr:
Follow-up on #715 

- Enable the referee tests.
- Make sure that bots can be called with shell commands, thus passing proper paths directly.
- Delete auxiliary shell test files: now the bots are called directly.
- Add tests that check if pyspiel/open_spiel.python can be imported in the Python interpreter.
- Add flag to pass build directory to referee tests.
- Drop the use of StartedSuccessfully -- the use-case was for testing availibility of the binaries, and this is checked by shell automatically.

### DIFF
--- a/open_spiel/python/CMakeLists.txt
+++ b/open_spiel/python/CMakeLists.txt
@@ -199,7 +199,6 @@ set(PYTHON_TESTS ${PYTHON_TESTS}
   tests/bot_test.py
   tests/games_bridge_test.py
   tests/games_sim_test.py
-  tests/higc_referee_test.py
   tests/policy_test.py
   tests/pyspiel_test.py
   tests/rl_environment_test.py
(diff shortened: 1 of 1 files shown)


michalsustr:
> It seems to have worked. Nicely done!

I thought about it a little more and I will still make some changes. I think the test flakiness stems from the fact that there are time limits for subprocess communication and if they are exceeded the test will fail. Now the time limits are _very generous_ for what is being tested (things like 1 sec to pick a random value from an array), so I never had an issue on my PC. But I suspect Github Actions uses something like spot instances, where they run the tests only if they have nothing else to run with a higher priority, potentially causing unpredictable delays and hence flakiness.

lanctot:
> This would require adding a special env flag "I-am-running-on-CI", which I am not sure is a good idea to introduce 🤔 The other option with long timeouts can still randomly flake.

This sounds good to me.. I was going to suggest it as well, I didn't because I thought it'd be a larger nontrivial change

> I was thinking -- internally it wasn't causing any issues? If not, then the cpp/py tests could still optionally run in blaze. However, in preemptive test environments they can be flaky, as the _point_ of the testing is to check _precisely_ if the bots correctly respond within the timeouts.

It's not running internally, because of the way it uses piping/signals and paths, it would require a lot of work to support both the way you have it now externally and using the google equivalents internally. So right now it's only being tested externally (and when I run the CMake tests before each sync).

PR #738, opened 2021-11-04
Return all reachable states in a game if it contains loops

mlschmidt366:
The function to return all (potentially chance) states in a game got stuck in a loop if the same state can be repeatedly encountered throughout a game. This commit fixes the issue by only expanding nodes that were not visited before, or which were not saved in the case of chance nodes.

I also adjusted the value iteration example, which relies on the above method, to take command line arguments. I tested the value iteration example with the game of "Pig" and a winscore of 20 (to save memory). It previously ran forever, now it works as expected.


lanctot:
Thanks @mlschmidt366 ! Sorry for the delay, I've been a bit busy... but will take a look soon.

Can I ask you to pull changes from master so that we can run the CI tests? (Specifically, you just need the changes in [this commit](https://github.com/deepmind/open_spiel/commit/2b86fdf83dca3a587c863aef6c7513489ce99d91), so you can also modify it in your branch if you prefer.)

mlschmidt366:
Hello,
I simplified the suggested change a lot, now it's really just an additional 'return' statement. Also, I realized that the value_iteration_example is used as unit test for the algorithm too, so I reincluded the sanity checks based on tic tac toe.
Best, Marvin

PR #740, opened 2021-11-10
Expectiminimax algorithm in C++

mlschmidt366:
Hello,

I translated the [existing Python implementation](https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/minimax.py) of Expectiminimax to C++. In line with the Python implementation, but in contrast to the alpha-beta search, I have utilized cloning of the states instead of do->undo actions, because some stochastic games do not yet implement an UndoAction().

I also added a test case for the algorithm based on the game of 'Pig' and added an example to showcase its usage using a simple heuristic value estimate. I verified (on a separate branch) that the implementation is also compatible with and correct when passing a solution obtained via e.g. value iteration as value function.

Feedback welcome!
Marvin

### DIFF
--- a/open_spiel/algorithms/minimax.cc
+++ b/open_spiel/algorithms/minimax.cc
@@ -37,9 +37,9 @@ namespace {
 //     alpha, the MAX player will avoid it).
 //   beta: the best value that the MIN currently can guarantee (if the value is
 //     >= than beta, the MIN player will avoid it).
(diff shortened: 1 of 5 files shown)


lanctot:
Nice, thanks!

Also thanks for updating the other PRs. I am off for a few days but will take a deeper look some time next week.

PR #741, opened 2021-11-12
Adding deep best response computation with a BR value function parameter for Fictitious Play

ajain-23:
This PR adds functionality for computing the best response value with RL methods (e.g. DQN, as opposed to standard exact BR) in the MFG Fictitious Play algorithm. It is a first pass towards Deep Fictitious Play for mean-field games. Incorporates changes from incorporate the changes from 5b149b2.


open_spiel/python/mfg/algorithms/fictitious_play.py
```diff
@@ -102,4 +127,4 @@ def iteration(self):
         self._game, list(range(self._game.num_players())),
         [self._policy, greedy_pi], [distrib, distrib_greedy],
         [1.0 * self._fp_step / (self._fp_step + 1), 1.0 /
-         (self._fp_step + 1)]).to_tabular(states=self._states)
+        (self._fp_step + 1)]).to_tabular(states=self._states)
```
TheoCabannes, line 130: Add a line break at the end of the file.

open_spiel/python/mfg/algorithms/fictitious_play_test.py
```diff
@@ -78,6 +107,33 @@ def test_fp_cpp_game(self):
...
+
+    dfp_policy = dfp.get_policy()
+    nash_conv_dfp = nash_conv.NashConv(game, dfp_policy)
+
+    self.assertAlmostEqual(nash_conv_dfp.nash_conv(), 1.0558451955622807)
+
 
 if __name__ == "__main__":
-  absltest.main()
+  absltest.main()
```
TheoCabannes, line 139: Add a line break at the end of the file.

TheoCabannes (commented):
LGTM

lanctot:
Hi @ajain-23 there are now some conflicting files. Can you pull & merge from master?

lanctot:
Hi @ajain-23,

The wheels tests are failing now, and that's because the fictitious_play_test now depends on JAX (via hiaku) since you've added an import to the JAX DQN. 

To fix this, you simply have to move the test from the general python tests section into the JAX tests section of the CMakeLists, here: https://github.com/deepmind/open_spiel/blob/c3d50a3fb3de7846bcba632619649b9f5d2f284c/open_spiel/python/CMakeLists.txt#L216

ajain-23:
> Hi @ajain-23,
> 
> The wheels tests are failing now, and that's because the fictitious_play_test now depends on JAX (via hiaku) since you've added an import to the JAX DQN.
> 
> To fix this, you simply have to move the test from the general python tests section into the JAX tests section of the CMakeLists, here:
> 
> https://github.com/deepmind/open_spiel/blob/c3d50a3fb3de7846bcba632619649b9f5d2f284c/open_spiel/python/CMakeLists.txt#L216

Done!

lanctot:
Hi @ajain-23,

Could you try modifying the wheels.yml like in this PR? https://github.com/deepmind/open_spiel/pull/748

It seems to be causing the current test failures.

ajain-23:
> Hi @ajain-23,
> 
> Could you try modifying the wheels.yml like in this PR? #748
> 
> It seems to be causing the current test failures.

Done.

lanctot:
Ok now it seems like a legit failure, but the same one as before (required change to python/CMakeLists).

Maybe I missed it but I don't see the change I suggested here in anh of your commits: https://github.com/deepmind/open_spiel/pull/741#issuecomment-980484930

ajain-23:
> Ok now it seems like a legit failure, but the same one as before (required change to python/CMakeLists).
> 
> Maybe I missed it but I don't see the change I suggested here in any of your commits: [#741 (comment)](https://github.com/deepmind/open_spiel/pull/741#issuecomment-980484930)

image
Is this not that change?

lanctot:
Ah I think you moved python/algorithms/fictitious_play_test.py rather than the mfg one :)

ajain-23:
> Ah I think you moved python/algorithms/fictitious_play_test.py rather than the mfg one with the same name :)

Whoops! My apologies, should be fixed.

PR #747, opened 2021-11-25
Piglet variant for the game of Pig

mlschmidt366:
Hi,

The main contribution of this PR is adding the Piglet variant for the game of Pig, roughly extending on its definition given in [1]. The difference to Pig is that whenever the turn total is increased, it is increased by a fixed step of 1, instead of the number that the die shows. Therefore, it is ideal to combine this variant with a lower 'winscore' in order to achieve a much smaller state space while not giving up too much of the difficulty of the original Pig game.

Additional improvements are:
- removing the TODO on 'MaxChanceNodesInHistory'
- The computation of the observation tensor utilizes a 2d tensorview to increase readability of the code
- 'NumDistinctActions' corrected to 2 (roll or stop)

[1] Todd W. Neller and Clifton G.M. Presser. [Optimal Play of the Dice Game Pig](http://cs.gettysburg.edu/~tneller/papers/pig.zip), The UMAP Journal 25(1) (2004), pp. 25-47.


lanctot:
Nice!! Thanks. I will take a deeper look on Monday.

open_spiel/games/pig.cc
```diff
@@ -234,10 +222,18 @@ std::vector<std::pair<Action, double>> PigState::ChanceOutcomes() const {
...
   std::vector<std::pair<Action, double>> outcomes;
 
-  // Chance outcomes are labelled 0+, corresponding to rolling 1+x.
-  outcomes.reserve(dice_outcomes_);
-  for (int i = 0; i < dice_outcomes_; i++) {
-    outcomes.push_back(std::make_pair(i, 1.0 / dice_outcomes_));
+  if (piglet_) {
+    // Chance outcomes are labelled 0 or 1, corresponding to rolling 1 or not 1 respectively
+    outcomes.reserve(2);
+    outcomes.push_back(std::make_pair(0, 1.0 / dice_outcomes_));
```
RaphaelMarinier, line 228: Is it expected that dice_outcomes_ is used here, since the original piglet uses a fair coin?
If it is, maybe explain better how dice_outcomes_ in piglet mode? It might be clearer to introduce another parameter instead of changing the meaning of dice_outcomes_ depending on whether we are in piglet mode or not.

mlschmidt366: Good point, this was unclear. It is indeed intended, but I added a paragraph to explain the Piglet variant in the header file. The meaning of dice_outcomes_ does not change, it still indicates the number of sides of the die that's used. Original Piglet uses a 2-sided die (a coin), original Pig uses a 6-sided die.

PR #754, opened 2021-12-19
Add Morpion Solitaire

kwj2104:
Adding a new game [Morpion Solitaire](https://en.wikipedia.org/wiki/Join_Five) (or Join Five).  Morpion Solitaire is typically a paper and pencil game for one or two players, played on a plus-shaped grid of dots. For this specific implementation, this is only for one player, and uses a simplified variation **"4D" (4 dots per line, disjoint variation)**. This game has been subject to a [variety of theoretical studies](http://www.morpionsolitaire.com/), and it would be interesting to see more modern RL techniques applied in the future. 

Not sure how much of the `spiel.h` interface needs to be implemented but happy to iterate on the PR. 

Thanks for this awesome repo!


lanctot:
Nice, thanks!! I am on vscation now until after the holidays, so I will take a look in January.

lanctot:
Meanwhile, I started the tests.

lanctot:
Can you also add it to the games list in docs/games.md?

PR #760, opened 2022-01-04
MCTS: Sample from prior

### LINKED ISSUE #753: Libtorch AlphaZero at Test-Time
Quick, and possibly silly, question regarding libtorch AZ at test-time (as in alpha_zero_torch_game_example).

alpha_zero_torch_game_example is running AZ as an MCTSBot with AZ's value network -- i.e. it inits an MCTSBot with the VPNetEvaluator instead of the standard evaluator. So it seems that (i) AZ is forced to use mcts at test-time (which requires at least max_simulations=2) and (ii) AZ uses the same number of mcts simulations (max_simulations) as its opponent.

Getting around number (ii) is pretty straightforward --> adding another flag separating, say, max_az_simulations from max_opponent_simulations, but I'm not quite sure about (i). So far, I've just been running AZ with max_simulations=2, but I'm not sure if this is equivalent to not using mcts at all. But I can confirm that changing max_simulations at test-time changes AZ's performance (especially from an early checkpoint where AZ is playing sub-optimally).

Any feedback/insight would be much appreciated!

ngrupen:
Fixes #753.

Modifies MCTS to handle max_simulations = 0 and max_simulations = 1 by sampling from prior, rather than erroring out.

Main Changes:
- Add SampleFromPrior() to SearchNode --> samples action using evaluator over legal actions.
- Step() --> Replaces check for children.size() = 0 with check for max_simulations = 0 or max_simulations = 1 and calls SampleFromPrior() in this case. Otherwise, proceeds as normal.
- ApplyTreePolicy() --> checks for current_node->children.size() = 0 and calls SampleFromPrior() in this case. Otherwise, selects actions as normal.

### DIFF
--- a/open_spiel/algorithms/mcts.cc
+++ b/open_spiel/algorithms/mcts.cc
@@ -174,6 +174,13 @@ std::string SearchNode::ToString(const State& state) const {
       children.size());
 }
 
+Action SearchNode::SampleFromPrior(const State& state, std::shared_ptr<Evaluator> evaluator, std::mt19937* rng) const {
+  std::unique_ptr<State> working_state = state.Clone();
+  ActionsAndProbs prior = evaluator->Prior(*working_state);
+  Action chosen_action = SampleAction(prior, *rng).first;
+  return chosen_action;
+}
+
 std::vector<double> dirichlet_noise(int count, double alpha,
                                     std::mt19937* rng) {
   std::vector<double> noise;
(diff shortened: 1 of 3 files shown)


ngrupen:
Re-opening after push to fix global variables.

lanctot:
Great, I've imported it and added a simple test. It'll go through a quick internal review, so should be available on the next update to github. Thanks again!

PR #771, opened 2022-01-23
Amazons implementation v1

selfsim:
Hello @lanctot !
After a while away from RL I am back to work on this project. Here is a new PR. I hope this works.
Let me know!
Cheers,
selfsim


lanctot:
Awesome! Welcome back. Can I ask you to merge the latest changes from master to fix the conflict?

lanctot:
Should be as simple as running

```
git pull origin master  # fetch changes and commit the merge locally
git push
```

lanctot:
(In this case I think the merge is easy, you can even do it via the 'Resolve conflict' button above.

selfsim:
@lanctot done

open_spiel/python/tests/pyspiel_test.py
```diff
@@ -27,7 +27,8 @@
 import pyspiel
 
 # Specify game names in alphabetical order, to make the test easier to read.
-EXPECTED_GAMES = frozenset([
+EXPECTED_GAMES = set([
```
lanctot, line 30: Did you merge from master..? This should be frozenset

open_spiel/games/amazons_test.cc
```diff
@@ -0,0 +1,166 @@
...
+int main(int argc, char** argv) {
+  open_spiel::amazons::BasicSpielTests();
+
+  // These tests check whether certain board configurations indicate the correct number of moves
+  open_spiel::amazons::PlayerOneSimpleWinTest();
+  open_spiel::amazons::PlayerTwoSimpleWinTest();
+  open_spiel::amazons::PlayerOneTrappedByAmazonsTest();
+  open_spiel::amazons::PlayerOneTrappedByBlocksTest();
+
+
```
lanctot, line 165: Remove the blank lines

lanctot (changes requested):
Can you add Amazons to the games list (see docs/games.md)

lanctot:
Hi @selfsim, looks good! I've imported it. It's currently undergoing internal review now, so probably will be merged by the next update.

A few quick C++ tips:

- Avoid `using namespace std`. It's customary to either use specific things (e.g. `using std::min`) or just prepend the namespace to the functions when you use them. The former is preferred when the namespace is long and there are many uses in the file.
- Don't use C-style casts. Prefer C++ style casts: e.g. `static_cast<unsigned char>(value)`
- Don't use sprintf. Use `absl::StrCat`, `absl::StrAppend`, or `absl::StrFormat`.

Thanks again... great game addition to OpenSpiel!

selfsim:
@lanctot thanks for the review! I will be refactoring the code as my C++ improves. 

I have a question:

Within my game implementation a move is broken into submoves; instead of giving a value for a move (select amazon, move amazon, shoot arrow) in one go, the move is built incrementally. One idea I had was to take a hierarchical RL approach towards agent design for amazons, where the subturn policies are learned by individual networks. Do you think this would be possible within the open_spiel framework?

Issue #779, opened 2022-01-28
Quoridor: No support for 3 or 4 players


maxspahn:
I just learned that Quoridor can also be played with 3 or 4 players. According to the documentation this is also supported in open_spiel.

However, the number of players is fixed in the header.

https://github.com/deepmind/open_spiel/blob/f522d174a1e2e8fbaf2007294985869d6e520669/open_spiel/games/quoridor.h#L35

Either the documentation or the implementation should be updated accordingly.

Looking forward to hearing from you.

lanctot:
From the looks of it, our implementation only supports the two-player version. 

Which documentation are you referring to.. the games list? Can I ask you to submit a PR fix to clarify the docs?

It would be quite fun to add support for 3-4 players! I encourage you to add this and contribute it. ;-) 

Tagging the original author: @tewalds

lanctot:
Merged in https://github.com/deepmind/open_spiel/pull/780

PR #780, opened 2022-01-28
Corrects docs for quoridor

maxspahn:
This fixes only the documentation in the games.md file.
I will work on the feature to integrate 3 and 4 players support.
Addresses #779 .


lanctot:
Thanks @maxspahn !

PR #782, opened 2022-02-02
Ft quoridor four player

maxspahn:
Adds option to play quoridor with up to four players.
Adds row and column numbering in the ToString method.
Updates the playthrough accordingly.

Questions: 
- I was not able to generate a new playthrough with up to four players, because the
 `open_spiel/scripts/generate_new_playthrough.sh` failed due to the Non-ASCII character in `generate_playthrough.py`. How can I do this?
- How can I verify if the observation tensor is correct?
- Should I squash all commits to one commit?


lanctot:
Great, thanks for the PR!

>     * I was not able to generate a new playthrough with up to four players, because the
>       `open_spiel/scripts/generate_new_playthrough.sh` failed due to the Non-ASCII character in `generate_playthrough.py`. How can I do this?

Odd. My guess is that you are accessing a bad part of memory that you should not be (maybe a buffer overflow). The games_sim_test failed in python.. seems like maybe there are some states that do not have any legal moves, which should not happen. My guess is these two are related: probably something is trying to sample an action from an empty list and then reading from unallocated memory when indexing into it, hence the non-ASCII characters.

My advice is to run thousands / millions of the game simulation tests to catch the cause of empty legal move list.

>     * How can I verify if the observation tensor is correct?

We usually manually inspect the playthrough and also a few random games in text. There's no better way, unfortunately. Once these basic tests pass, I will ask the original author of this game to review the code, and that will help. 

>     * Should I squash all commits to one commit?

Nope, no need. Our internal tool does that when importing the PR.

maxspahn:
> My guess is these two are related: probably something is trying to sample an action from an empty list and then reading from unallocated memory when indexing into it, hence the non-ASCII characters.

Indeed, there might be a very unlikely situation in which no legal moves are available, see https://boardgames.stackexchange.com/questions/31963/in-a-four-person-game-of-quoridor-can-a-player-be-forced-to-pass-a-turn . As there don't seem to be any official rules for this situation, the option is either a forced 'pass (no action)' or a direct end of the game, like a draw in chess. What do you think is better?

lanctot:
Interesting! I don't know the game but based on a very quick skim of the rules and the conversation in the thread you linked, I'd say just adding a "pass" action (that is only legal when there are no other legal moves) makes the most sense.

@tewalds what do you think?

lanctot:
I spoke to @tewalds (original implementor of the game) and another Quoridor player, we agreed adding a pass move (only legal in the case where there are not legal moves) is a good option.

maxspahn:
Alright, then I change it accordingly.

open_spiel/games/quoridor.h
```diff
@@ -28,11 +28,14 @@
 //   "board_size"        int     Size of the board (default = 9)
 //   "wall_count"        int     How many walls per side (default = size^2/8)
 //   "ansi_color_output" bool    Whether to color the output for a terminal.
+//   "players"           int     Number of players (default = 2)
 
 namespace open_spiel {
 namespace quoridor {
 
 inline constexpr int kNumPlayers = 2;
```
lanctot, line 36: Can you rename this to kDefaultNumPlayers?

lanctot (changes requested):
In addition to the 4-player playthrough, can you add some basic tests to quoridor_test.cc for the 3-player and 4-player games?

open_spiel/games/quoridor.h
```diff
@@ -28,20 +28,25 @@
...
-inline constexpr int kNumPlayers = 2;
+inline constexpr int kDefaultNumPlayers = 2;
+inline constexpr int kMinNumPlayers = 2;
+inline constexpr int kMaxNumPlayers = 4;
 inline constexpr int kDefaultBoardSize = 9;
 inline constexpr int kMinBoardSize = 3;
 inline constexpr int kMaxBoardSize = 25;
 inline constexpr int kMaxGameLengthFactor = 4;
-inline constexpr int kCellStates = 1 + kNumPlayers;
+inline constexpr int kCellStates = 1 + kDefaultNumPlayers;
```
maxspahn, line 43: I am not sure what this value should be. 
I think it must be adapted according to `num_players_`.
Maybe, the initial author knows better, @tewalds.

lanctot: Yes this should no longer be a constant. You should remove it and instead make it a function inside the game or state object and call that function whenever that value is need. (Kind of surprised the 4-player case would work at all without this change)

lanctot:
Ignore the failing wheel test, that's our fault :)

open_spiel/games/quoridor_test.cc
```diff
@@ -34,15 +34,28 @@ void BasicQuoridorTests() {
         *LoadGame(absl::StrCat("quoridor(board_size=", i, ")")), 5);
   }
 
+  for (int i = 2; i <= 2; i++) {
```
lanctot, line 37: A loop from i=2 to i<=2? Did you mean i <= 4 here?

lanctot (changes requested):

open_spiel/games/quoridor_test.cc
lanctot: I've imported this and fixed this to <= 4. No need to update. It's undergoing internal review now, I'll add the "imported" label when it's done.

maxspahn: Ah yes, it was meant to test all possible numbers of players. Thank you! ;)

PR #786, opened 2022-02-08
Update MFG cross ref and fix nits in description

TheoCabannes:
- Fix nits in routing games description
- Update MFG cross reference


lanctot:
Thanks! The problem with the failed test is our fault, I'll fix that.

PR #788, opened 2022-02-14
Upgrade TF version to 2.8.0 (postponed to April/May)

lanctot:
In python_extra_deps.sh


lanctot:
Looks like numpy 1.20 (required by TF 2.8.0) is not avilable for Python 3.6, which is the Python version on Ubuntu 18.04.

Ubuntu 18.04 is still supported until April 2023, but Python 3.6 reached end of life on Dec 23rd, 2021.

Still, we're in no rush for this upgrade.. so I anticipate we'll remove support for Python 3.6 in the next few months, and then can push this through afterward.

lanctot:
Replacing this PR with https://github.com/deepmind/open_spiel/pull/807

Issue #789, opened 2022-02-19
NFSP with CNN


BedirT:
Do we have NFSP implementation using a CNN rather than a linear network? I can't seem to locate it, I would appreciate it if someone could direct me (if it exists).

lanctot:
Hi Bedir,

We do not. Most of the imperfect information games have single dimensional info set tensors. 

Extending NFSP to support CNNs (with e.g. an example on Dark Hex) would make a nice contribution!

BedirT:
Yeah I will work on that :ok_hand:

BedirT:
@lanctot 

I have a couple of questions (All on the DarkHex version, but it is probably the same for some other phantom games):
- The game implements the info-states as one-hot encoding which I don't think would capture the spatial connection I aim to have with CNN implementation. If that sounds fine I will change the game itself. Let me know if there is a better way.
- Is it fine if I used tf2 & default layering system with Functional API, or should I write custom layers?

lanctot:
>     * The game implements the info-states as one-hot encoding which I don't think would capture the spatial connection I aim to have with CNN implementation. If that sounds fine I will change the game itself. Let me know if there is a better way.

If you're planning to train it only on the current "view" (losing all historical information) then I suggest you use the ImperfectRecallDarkHex and override the information state tensor to be just the spatialized view. 

To see how to organize the tensor for use by CNNs, see some of the perfect information games which lay them out in planes (e.g. Tic-Tac-Toe, Breakthrough) which can then be reshaped on the python size. The AlphaZero implementations use CNNs if you want to see how those observations get used in CNN training. 

>     * Is it fine if I used tf2 & default layering system with Functional API, or should I write custom layers?

For extending NFSP? I don't have a preference as long as the basic NFSP we have now still works as before (i.e. make CNNs optional and disabled by default, or copy the file and make it a new NFSP specifically for CNNs).

BedirT:
I am pretty much done, except that I cannot figure out some c++ related stuff :) 

When I implement a non-one-hot version of the `information_state` and `observation_state`, I can run the game, but I get weird shapes from `InformationStateTensorShape` and `ObservationTensorShape`... I couldn't find a proper way to handle the issue.

Here is an example from the playthrough;
```bash
NumDistinctActions() = 12
PolicyTensorShape() = [12]
MaxChanceOutcomes() = 0
[...]

PR #791, opened 2022-02-23
NFSP CNN + Updated Imperfect Recall Dark Hex

BedirT:
NFSP part is not fully complete, mainly uploading not to have multiple PRs for the same purpose.

IR-DarkHex updates work fine on the tests. When trying to generate new playthroughs, I get a weird shape on `ObservationTensorShape`,  It works well when I remove the function and inherit the old one (even though the numbers are wrong in that case), so I think the issue is on

https://github.com/deepmind/open_spiel/blob/7e6473cc6e64b1131b79553e4823ff3910f58c95/open_spiel/games/dark_hex.cc#L341-L343

I am not using `num_cells_` since it acts weird when I make it public or protected instead of private. Addressing #789


open_spiel/games/dark_hex.cc
```diff
@@ -322,6 +322,42 @@ std::vector<int> DarkHexGame::ObservationTensorShape() const {
...
+
+  const auto& player_view = (player == 0 ? black_view_ : white_view_);
+  for (int cell = 0; cell < num_cells(); ++cell) {
+    values[cell] = static_cast<int>(player_view[cell]) - kMinValueCellState;
+  }
+}
+
+std::vector<int> ImperfectRecallDarkHexGame::ObservationTensorShape() const {
+  return {num_cells()};
+}
```
lanctot, line 360: These should be multi-dimensional. 

See Breakthrough and Tic-Tac-Toe. Use the TensorView helper to make planar observations as done in those games.

open_spiel/games/dark_hex.cc
```diff
@@ -323,5 +325,63 @@ std::vector<int> DarkHexGame::ObservationTensorShape() const {
...
+      SPIEL_CHECK_TRUE(plane >= 0 && plane < 3);
+      view[{plane, r, c}] = 1.0;
+    }
+  }
+}
+
+void ImperfectRecallDarkHexState::ObservationTensor(Player player,
+                                     absl::Span<float> values) const {
+  // Same as the InformationStateTensor.
+  InformationStateTensor(player, values);
```
BedirT, line 379: Not sure if this is legal for OS, but it is unnecessary to repeat the code I think.

BedirT:
Yeah, I am using the updated ones for my experiments right now, but couldn't find time to clean up. Will do it in couple of weeks hopefully.

Yeah subclass will be better I think, I will do that.

lanctot:
Closing for now, feel free to re-open a new PR.

PR #801, opened 2022-03-11
Implemented partial observable mode for laser tag

Jjschwartz:
Hey Open Spiel team,

This pull request relates to adding a partially observable mode to the Laser Tag game based on the descriptions in the related papers. 

Some things to note about the implementation:
- by default the laser tag game still uses the fully observable mode and is unchanged, 
- partially observable mode can be activated by setting the game parameter `fully_obs=false` and the observation dimensions can be changed using the `obs_front`, `obs_back`, `obs_side` parameters (by default they are set to the same as for the PSRO paper: 17, 2, 10),
- the returned tensor uses the tensor format currently implemented (with 4 binary planes), as opposed to returning an RGB tensor. 

I should be easy to extend the proposed implementation to also support RGB observation tensors. Let me know if you think it would be useful to add.  

Thanks for the awesome repo!


lanctot:
Thanks! We will try to take a look this week.

open_spiel/games/laser_tag.cc
```diff
@@ -493,6 +527,63 @@ void LaserTagState::ObservationTensor(int player,
   }
 }
 
+void LaserTagState::PartialObservationTensor(int player,
+					     absl::Span<float> values) const{
```
jhtschultz, line 531: Add space after `const`, see comment above.

open_spiel/games/laser_tag.cc
```diff
@@ -529,7 +620,8 @@ double LaserTagGame::MaxUtility() const {
 }
 
 std::vector<int> LaserTagGame::ObservationTensorShape() const {
-  return {kCellStates, grid_.num_rows, grid_.num_cols};
+  if (fully_obs_) return {kCellStates, grid_.num_rows, grid_.num_cols};
+  return {kCellStates, obs_front_ + obs_back_ + 1, obs_side_ * 2 + 1};
```
jhtschultz, line 624: This should be an if-else statement.

open_spiel/games/laser_tag.cc
```diff
@@ -115,8 +130,18 @@ constexpr std::array<std::array<int, 10>, 4> col_offsets = {
      {0, 0, -1, 1, 0, 0, 0, -1, -1, 0}}};
 }  // namespace
 
-LaserTagState::LaserTagState(std::shared_ptr<const Game> game, const Grid& grid)
-    : SimMoveState(game), grid_(grid) {}
+LaserTagState::LaserTagState(std::shared_ptr<const Game> game, const Grid &grid)
```
jhtschultz, line 133: Please adhere to convention of putting the `&` after the type.

open_spiel/games/laser_tag.cc
```diff
@@ -481,8 +506,17 @@ void LaserTagState::ObservationTensor(int player,
...
   SPIEL_CHECK_LT(player, num_players_);
 
+  if (fully_obs_) {
+    FullObservationTensor(values);
+  } else {
+    PartialObservationTensor(player, values);
+  }
+}
+
+void LaserTagState::FullObservationTensor(absl::Span<float> values) const{
```
jhtschultz, line 516: Add space between `const` and `{`. In general it's a good practice to run any PRs through cpplint (https://google.github.io/styleguide/cppguide.html#cpplint).

open_spiel/games/laser_tag.cc
```diff
@@ -481,8 +506,17 @@ void LaserTagState::ObservationTensor(int player,
...
+
   TensorView<3> view(values, {kCellStates, grid_.num_rows, grid_.num_cols},
-                     true);
+		     true);
```
jhtschultz, line 519: Looks like there's extra white space here. Linter should remove it.

open_spiel/games/laser_tag.cc
```diff
@@ -567,11 +659,15 @@ Grid ParseGrid(const std::string& grid_string) {
 }
 }  // namespace
 
-LaserTagGame::LaserTagGame(const GameParameters& params)
+LaserTagGame::LaserTagGame(const GameParameters &params)
```
jhtschultz, line 662: Please adhere to convention of putting the `&` after the type.

open_spiel/games/laser_tag_test.cc
```diff
@@ -132,6 +132,14 @@ void BasicLaserTagTestsBigGrid() {
       10);
 }
 
+void BasicPartiallyObservableLaserTagTests() {
```
jhtschultz, line 135: Please add a test of the observation tensor and observation string. Making the game partially observable isn't really changing the internal game logic, so these tests aren't testing much. If there are bugs, they're likely to show up in the tensor & string.

You don't need to go overboard with the tests, but I would verify that we're getting what we expect in a couple situations, and I'd vary the default params `obs_forward`, `obs_back`, and `obs_side` in the same situations and confirm that works as well.

jhtschultz (changes requested):
This is a nice addition! Left some mostly minor comments.

I don't see any handling of the observation string, only the tensor. I believe the string should be updated as well to reflect the partial observability.

open_spiel/games/laser_tag.cc
Jjschwartz: Thanks for letting me known I'll make sure to do that for future PRs.

open_spiel/games/laser_tag_test.cc
Jjschwartz: Hey, thanks for the feedback. I realized while reviewing your suggestions that I had forgotten to update the ObservationString methods as well. I have rectified this and added some tests of the observation tensor and observation string as suggested.

Let me know if there is anything I missed or needs changing.

jhtschultz (changes requested):
Tests look great! Very thorough, thanks for adding them. Expected behavior is now much clearer to anyone looking to use this option, and we can have a lot more confidence there aren't any bugs in the partial obs mode.

Left one more nitty suggestion which should only take a min to add and then the PR should be good to go.

open_spiel/games/laser_tag_test.cc
```diff
@@ -140,6 +144,311 @@ void BasicPartiallyObservableLaserTagTests() {
...
+        tensor[2 * obs_grid_size + i - num_newlines] = 1.0;
+        break;
+      case '*':
+        tensor[3 * obs_grid_size + i - num_newlines] = 1.0;
+        break;
+      case '\n':
+        num_newlines += 1;
+        break;
+      default:
+        // Reached 'O' in "Orientations"
```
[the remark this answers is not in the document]
Jjschwartz: No problem, should be all fixed now.

jhtschultz (approved):
Great contribution, thanks!

Jjschwartz:
My pleasure. Thanks for the awesome library.

PR #805, opened 2022-03-19
Implemented Boltzmann Q-learning python

JoppeDJ:
I have implemented a variation of the tabular Q learner in python that uses a Boltzmann probability interpretation of the Q-values for action selection instead of e-greedy (the already existing implementation). According to equation (2) page 2 here: [paper1](https://arxiv.org/pdf/1109.1528.pdf)

This action selection implementation corresponds with the selection mechanism that results in the 'boltzmannq' replicator dynamics that are implemented in python/egt/dynamics and thus seemed like a nice add-on to me for the algorithms.

I'm glad to hear about any changes that I should make to my code to make it more open_spiel friendly.

Other notes/questions:
- The already implemented Q learner had a reference to an example where it is used, should a similar example be provided for my implementation?
- When it comes to testing I have verified (by hand) that the learning trajectories of my implementation indeed correspond with the already implemented boltzmannq replicator dynamics for different matrix games and values of the temperature parameter. Do I need to provide other programmatic ways of testing the implementation?

### DIFF
--- a/open_spiel/python/algorithms/tabular_qlearner.py
+++ b/open_spiel/python/algorithms/tabular_qlearner.py
@@ -53,8 +53,9 @@ class QLearner(rl_agent.AbstractAgent):
         self._q_values = collections.defaultdict(valuedict)
         self._prev_info_state = None
         self._last_loss_value = None
+        self._current_epsilon = None
 
-    def _epsilon_greedy(self, info_state, legal_actions, epsilon):
+    def _epsilon_greedy(self, info_state, legal_actions):
         """Returns a valid epsilon-greedy action and valid action probs.
 
     If the agent has not been to `info_state`, a valid random action is chosen.
@@ -72,10 +73,17 @@ class QLearner(rl_agent.AbstractAgent):
         greedy_actions = [
             a for a in legal_actions if self._q_values[info_state][a] == greedy_q
         ]
-        probs[legal_actions] = epsilon / len(legal_actions)
-        probs[greedy_actions] += (1 - epsilon) / len(greedy_actions)
+        probs[legal_actions] = self._current_epsilon / len(legal_actions)
+        probs[greedy_actions] += (1 - self._current_epsilon) / len(greedy_actions)
         action = np.random.choice(range(self._num_actions), p=probs)
         return action, probs
+
+    def _get_action_probs(self, info_state, legal_actions):
+        """Returns a selected action and the probabilities of legal actions.
+
+        To be overwritten by subclasses that implement other action selection methods.
+        """
+        return self._epsilon_greedy(info_state, legal_actions)
 
     def step(self, time_step, is_evaluation=False):
         """Returns the action to be taken and updates the Q-values if needed.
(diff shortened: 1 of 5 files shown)


lanctot:
Thanks, I will take a look soon!

lanctot:
Hi @JoppeDJ, thanks! 

This is a larger change than I expected. I am wondering, would it be possible to have a function that exposed the policy in the base class, which defaults to epsilon-greedy and then have a very minimal subclass that simply overrides this function and chooses based on softmax? @sgirgin did something like this recently for the Boltzmann DQN in JAX. https://github.com/deepmind/open_spiel/commit/40992707a8b14e61de0336a7f3d36512668736a6

(It might also be worth putting it into its own file, too.)

And yes, please add at least one basic test for this new agent. You can simply copy the one(s) from the Q-learning and runn it on the Boltzmann Q-learner.

JoppeDJ:
@lanctot That is definitely possible, I'll get to work on that.

JoppeDJ:
@lanctot it seems that I can't find any tests present for the tabular qlearner, is this normal?

lanctot:
It's possible we never wrote a test for that, but we should add one. Can you do that? I suggest a very basic one for both the standard Q-learning and the Boltzmann Q-learning, you can copy this simple 1-step game in the DQN test, here: https://github.com/deepmind/open_spiel/blob/062416fdd173424440acfa8938867df8b9e22735/open_spiel/python/algorithms/dqn_test.py#L42

open_spiel/python/algorithms/tabular_qlearner.py
```diff
@@ -87,46 +95,46 @@ def step(self, time_step, is_evaluation=False):
...
+        legal_actions = time_step.observations["legal_actions"][self._player_id]
+
+        # Prevent undefined errors if this agent never plays until terminal step
+        action, probs = None, None
+
+        # Act step: don't act at terminal states.
+        if not time_step.last():
+            self._current_epsilon = 0.0 if is_evaluation else self._epsilon
+            action, probs = self._get_action_probs(
+                info_state, legal_actions)
```
JoppeDJ, line 111: For some reason my pycharm changed the indentation for this whole chunk of code, but I have only made changes in these 3 lines compared to the original function. I'm adding this pointer to make it easier for anyone who is reviewing this piece of code to find this change.

lanctot:
Cool. To include the python test you need to add it to python/CMakeLists.txt

JoppeDJ:
@lanctot in what way can I run these tests locally? I haven't figured this out yet. I will be switching from tensorflow testing to absl testing for the Q learners since tensorflow isn't actually needed. I will push the changed tests one more time but if they fail again I'll make sure to have the tests pass locally before pushing

lanctot:
Yes please use absl tests by default. To test locally you can run the `build_and_run_tests.sh` scripts, but if you already have `pyspiel` built and the PYTHONPATH properly set, then you just need to run the individual test file, e.g. `python boltzmann_tabular_qlearner_test.py`.

lanctot:
I see the test is failing, I think 100 is too low and also we should be collecting evaluation values. I ran into a similar problem in my recent PR. 

I suggest the patch I made: 
```
   total_reward = 0

    for _ in range(100):
    total_eval_reward = 0
    for _ in range(1000):
      time_step = env.reset()
      while not time_step.last():
        agent_output = agent.step(time_step)
        time_step = env.step([agent_output.action])
        total_reward += time_step.rewards[0]
      agent.step(time_step)
    self.assertGreaterEqual(total_reward, 75)
    for _ in range(1000):
      time_step = env.reset()
      while not time_step.last():
        agent_output = agent.step(time_step, is_evaluation=True)
        time_step = env.step([agent_output.action])
        total_eval_reward += time_step.rewards[0]
    self.assertGreaterEqual(total_eval_reward, 250)
```

lanctot (changes requested):
For both tests, can you `import numpy as np` at the top of the file, define a fixed `SEED = ...` and then right before calling `absltest.main()` seed the RNG via `np.random.seed(SEED)`. This ensures that the tests are not stochastic.

lanctot (approved):

PR #807, opened 2022-03-21
Fixes necessary for Ubuntu 22.04 / Python 3.10

lanctot:
- Add support for Python 3.10
- Drop support for Python 3.6 and Ubuntu 18.04
- Upgrade versions to TF/Keras to 2.8.0, also tensorflow-probability to 0.16.0 and tensorflow-datasets to 4.5.2
- Update JAX extra deps to newer versions requiring python >= 3.7 (chex, optax, jax, rlax, dm-haiku)
- Upgrade PyTorch version to 1.11.0
- Upgrade numpy >= 1.21.5, scipy >= 1.7.3 and several other packages to support Python 3.10
- Misc fixes caused by backward incompatibility


lanctot:
All tests now passing, except ubuntu-22.04 stsyed queued for 8 hrs. Will try again once it is released.

lanctot:
Ok, everything working again but Ubuntu 22.04 not yet available in CI. So will push these changes through without the Ubuntu 22.04 test and add it when it's available.

PR #832, opened 2022-05-06
adding flag to export `compile_commands.json`

sabify:
Hello,

This will be helpful for language servers and IDEs to work properly.

https://clangd.llvm.org/installation.html#project-setup


lanctot:
Thanks!

PR #835, opened 2022-05-09
add UtilitySum method to repeated games

VitamintK:
I noticed that python's `exploitability.exploitability()` was erroring because `utility_sum()` wasn't defined for repeated games.

In this PR, I add the `UtilitySum` method to repeated games. Also, I noticed that the `python/tests/game_transforms_test.py` file didn't seem like it was actually being run anywhere, so I added it to `PYTHON_TESTS` in `CMakeLists.txt`.

### DIFF
--- a/open_spiel/game_transforms/repeated_game_test.cc
+++ b/open_spiel/game_transforms/repeated_game_test.cc
@@ -124,6 +124,8 @@ void RepeatedPrisonersDilemaTest() {
   SPIEL_CHECK_EQ(repeated_game->GetType().min_num_players, 2);
   SPIEL_CHECK_EQ(repeated_game->GetType().utility,
                  GameType::Utility::kGeneralSum);
+  // repeated_game->UtilitySum() should raise an error. This is checked in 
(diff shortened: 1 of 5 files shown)


open_spiel/game_transforms/repeated_game_test.cc
```diff
@@ -124,6 +124,8 @@ void RepeatedPrisonersDilemaTest() {
   SPIEL_CHECK_EQ(repeated_game->GetType().min_num_players, 2);
   SPIEL_CHECK_EQ(repeated_game->GetType().utility,
                  GameType::Utility::kGeneralSum);
+  // repeated_game->UtilitySum() should raise an error. This is checked in 
+  // game_transforms_test.py as it's simpler to catch the error from Python.
```
VitamintK, line 128: I based this off of https://github.com/deepmind/open_spiel/blob/2acbd485af75fc92a4815fefd610c37292537127/open_spiel/tests/basic_tests.cc#L224

PR #841, opened 2022-05-21
Add new game: Kalah

Jazeem:
Mancala is one of the oldest games that's still played with small stones and rows of holes in a board. The objective is to capture the most stones (https://en.wikipedia.org/wiki/Mancala)

Even though the rules are simple, players can develop complex strategies in Mancala. It seems like a good addition for openspiel.


lanctot:
I agree, this will make a great addition to OpenSpiel. Thanks!

open_spiel/games/mancala_test.cc
```diff
@@ -0,0 +1,34 @@
...
+
+namespace open_spiel {
+namespace mancala {
+namespace {
+
+namespace testing = open_spiel::testing;
+
+void BasicMancalaTests() {
+  
+}
```
lanctot, line 26: Please add some basic tests. At the bare minimum, you should do the simulation tests (see e.g. breakthrough_test.cc). 

It's always better to have also have a few Mancala-specific tests too (see e.g. backgammon_test.cc), especially f there are any tricky rules situations that might occur.

open_spiel/games/mancala.cc
```diff
@@ -0,0 +1,237 @@
...
+  int num_beans = board_[move];
+  board_[move] = 0;
+  int current_pit = move;
+  for(int i = 0; i < num_beans; ++i){
+    current_pit = GetNextPit(current_player_, current_pit);
+    board_[current_pit]++;
+  }
+
+  //capturing logic
+  if(board_[current_pit] == 1 && IsPlayerPit(current_player_, current_pit) && board_[GetOppositePit(current_pit)] > 0){
```
lanctot, line 95: Space after the bracket here: e.g.

if (...) {

lanctot: (Similarly elsewhere throughout the file)

open_spiel/games/mancala.cc
```diff
@@ -0,0 +1,237 @@
...
+      if(board_[board_.size() - 1 - i] > 0){
+        moves.push_back(board_.size() - 1 - i);
+      }
+    }
+  }
+  return moves;
+}
+
+std::string MancalaState::ActionToString(Player player,
+                                           Action action_id) const {
```
lanctot, line 126: Remove two spaces

open_spiel/games/mancala.cc
```diff
@@ -0,0 +1,237 @@
...
+  std::string str;
+  std::string separator = "-";
+  absl::StrAppend(&str, separator);
+  for (int i = 0; i < kNumPits; ++i) {
+    absl::StrAppend(&str, board_[board_.size() - 1 - i]);
+    absl::StrAppend(&str, separator);
+  }
+  absl::StrAppend(&str, "\n");
+
+
```
lanctot, line 150: Remove extra blank line, similarly below

open_spiel/games/mancala.cc
```diff
@@ -0,0 +1,237 @@
...
+  SPIEL_CHECK_EQ(values.size(), kTotalPits);
+  auto value_it = values.begin();
+  for (int count : board_) {
+    *value_it++ = count;
+  }
+  SPIEL_CHECK_EQ(value_it, values.end());
+}
+
+void MancalaState::UndoAction(Player player, Action move) {
+  //Yet to implement Undo
```
lanctot, line 223: Either remove the comment, or implement the function. Many of the games do not have Undo, so you can just remove the function if it's not correct.

If you leave it, please add one of the simulation tests that tests the undo: https://github.com/deepmind/open_spiel/blob/63e8a99b70a4ea2eb934740d8fb1d2add2dd258b/open_spiel/tests/basic_tests.h#L57

lanctot (changes requested):
I made a few minor comments on style, and you need to add some tests before we can import it, but otherwise looks good!

(This is nicely in time for the release we're planning this or next week..!)

lanctot (commented):
Hi @Jazeem, 

Thanks for the changes. I just remembered we already have Oware (see [here](https://github.com/deepmind/open_spiel/blob/master/open_spiel/games/oware.cc)) and Mancala is a family of games. What variant or specific version does this one implement? Is it different from Oware?

Jazeem:
Hi @lanctot,

I was unaware of this and have done a bit of reading regarding the same. This specific variant seems to be called Kalah (https://en.wikipedia.org/wiki/Kalah).

Also, since there's already a Mancala game in OpenSpiel, it might be better to use the same terminology in the code regarding game elements, eg: Houses, Seeds.

lanctot:
Ok so would you say Kalah and Oware are actually different in terms of rules?

(If so, maybe best to rename your game to kalah?)

If they are indeed equivalent then we should probably not have two identical versions of the same game.

Jazeem:
Yes. Kalah and Oware are different in terms of boards used and rules involved. But have a few common elements too.

It would be better to rename this to Kalah. But Mancala is the more popular name, so this confusion that Mancala is actually a family of games would still remain I guess.

Jazeem:
@lanctot Renamed game to Kalah, and renamed game elements house, seed, store to match with Oware

lanctot:
> @lanctot Renamed game to Kalah, and renamed game elements house, seed, store to match with Oware

Oh no, I've already imported it and I made more changes from internal review. I think it's ok to keep the name as "mancala" and I just added a note that the rule set is Kalah (and added clarifications in both headers to point to Oware and Mancala). 

If you want, you can submit a follow-up PR when I sync with github and it gets merged on Monday for the renaming. I think it is also fine to leave it as it is, though, too.

Thanks again for this game!

Jazeem:
Hi @lanctot,

Let's leave it is as is then.
I saw your call for new games. I can contribute to one of them also if needed. Let me know if there's any that's not been taken up yet.

lanctot:
> I saw your call for new games. I can contribute to one of them also if needed. Let me know if there's any that's not been taken up yet.

Cool! I'll probably code crib myself (but slowly over time). From that list: if you're looking for easier ones, I think Nim and Nine Men's Morris are not being worked on. For the more difficult, Checkers would be really good to have.

But also feel free to use that thread to suggest more, yourself, too!

Jazeem:
Hi @lanctot,

Okay, I'll work on checkers. Since you already have Clobber, I'll start by extending that.

lanctot:
Ok it has been merged into master.

However, I just noticed-- I forgot to ask you to add it to the [games list](https://github.com/deepmind/open_spiel/blob/master/docs/games.md). Can I ask you to submit another PR for that?

PR #845, opened 2022-05-25
fixes an issue in C++ DQN related to choosing illegal actions

sabify:
Fixes an issue related to the poorly implementation of the algorithm that causes choosing illegal actions.

### DIFF
--- a/open_spiel/algorithms/dqn_torch/dqn.cc
+++ b/open_spiel/algorithms/dqn_torch/dqn.cc
@@ -146,7 +146,7 @@ Action DQN::EpsilonGreedy(std::vector<float> info_state,
   Action action;
   if (legal_actions.empty()) {
     // In some simultaneous games, some players can have no legal actions.
-    return kInvalidAction;
+    return 0;
   } else if (legal_actions.size() == 1) {
     return legal_actions[0];
   }
--- a/open_spiel/algorithms/dqn_torch/dqn.cc
+++ b/open_spiel/algorithms/dqn_torch/dqn.cc
@@ -31,7 +31,8 @@ namespace open_spiel {
 namespace algorithms {
 namespace torch_dqn {
 
-constexpr const int kIllegalActionLogitsPenalty = -1e9;
+constexpr const float kIllegalActionLogitsPenalty =
+    std::numeric_limits<float>::lowest();
 
 Action RandomAgent::Step(const State& state, bool is_evaluation) {
   if (state.IsTerminal()) {
@@ -216,7 +217,6 @@ void DQN::Learn() {
         torch::from_blob(t.legal_actions_mask.data(),
                          {1, t.legal_actions_mask.size()},
                          torch::TensorOptions().dtype(torch::kInt32))
-            .to(torch::kInt64)
             .clone());
     actions.push_back(t.action);
     rewards.push_back(t.reward);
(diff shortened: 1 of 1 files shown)


open_spiel/algorithms/dqn_torch/dqn.cc
```diff
@@ -141,37 +142,37 @@ void DQN::AddTransition(const State& prev_state,
 }
 
 Action DQN::EpsilonGreedy(std::vector<float> info_state,
-                          std::vector<Action> legal_actions,
-                          double epsilon) {
+                          std::vector<Action> legal_actions, double epsilon) {
   Action action;
   if (legal_actions.empty()) {
     // In some simultaneous games, some players can have no legal actions.
-    return 0;
```
lanctot, line 149: This should be zero actually (see other comment)

sabify: Done

sabify:
Done.

lanctot:
@christianjans @michalsustr can you take a quick look at this PR?

I'm not very familiar with the PyTorch C++ API.

@sabify can you explain in words what the problem was. I don't quite understand.

sabify:
Short answer: Integer/float overflows in addition can make illegal actions, legal. The masking was not implemented correctly.

Long answer: I'm going to explain it with an example. 

We have 5 total actions (`num_actions_`) and 2 legal actions. The current algorithm first try to allocate a tensor of 5 elements that has 3 elements with illegal action value (-1e9) and 2 elements for legal actions with zero values. Then, it is going to sum this tensor with Q values before selecting the highest value. The issue is that if we have a very large negative value and sum it with illegal action value, it can cause negative integer/float overflow or a big Q value in illegal action's indexes can be big enough that the addition be bigger than the legal action's values and selecting the highest value is no longer in the legal actions.

This PR correctly masks the legal values with torch built-in functions.

lanctot:
That makes sense... wow, very subtle... nice catch!!

@Asugawara @jhtschultz

Asugawara:
Thank you for fixing the bug!

>it can cause negative integer/float overflow or a big Q value in illegal action's indexes can be big enough that the addition be bigger than the legal action's values and selecting the highest value is no longer in the legal actions.

Am I correct in understanding that the target value is much smaller than 'kIllegalActionLogitsPenalty', so the possibility is very small?

p.s. Is that the same??
https://github.com/deepmind/open_spiel/blob/b3dae9dcd34fa75171baac09831e60d17c95ce1d/open_spiel/algorithms/dqn_torch/dqn.cc#L233-L238

lanctot:
@sabify I have already imported this before your last commit. So when I do a sync on Monday, it will be merged without your most recent commit. You can submit a new PR to make those changes.

sabify:
> p.s. Is that the same??

Yes, and I just updated the file with new implementation.

> Am I correct in understanding that the target value is much smaller than 'kIllegalActionLogitsPenalty', so the possibility is very small?

It maybe depends on the game and its dqn configuration. In my testing for a 4-player game with 53 actions, 2822 tensor size and 128x128 hidden layers configuration, the issue arise at 100000th game and the game would never proceed anymore. Unfortunately, the games in c++ dqn example won't catch this issue.

sabify:
> @sabify I have already imported this before your last commit. So when I do a sync on Monday, it will be merged without your most recent commit. You can submit a new PR to make those newest changes. (After it is merged on Monday.)

I will create new PR and they don't have any conflict, you can import both...

lanctot:
That works, thanks.

PR #870, opened 2022-06-27
Fix wrapping of an absl::optional in wheels tests

lanctot:


lanctot:
@jhtschultz just flagging. See https://github.com/deepmind/open_spiel/pull/868#issuecomment-1167990725.

The `absl::optional` types cause problems with pybind, specifically the wheel tests. Here is a slightly awkward fix for one but another came up. Let's chat about a better solution when we are both back.

lanctot:
Closing this PR in favor of https://github.com/deepmind/open_spiel/pull/873

PR #871, opened 2022-06-28
Add implementation of MMD with dilated entropy

ryan-dorazio:
Adding implementation of magnetic mirror descent (MMD) with dilated entropy as a solver for reduced normal-form quantal response equilibria. 

The implementation comes with two files, one that implements sequence form utility functions and is standalone. And the other implements the MMD algorithm with the help of the first utility file.


lanctot:
Nice, thanks!

There is currently a known problem with the wheels test (which we will fix soon) so it will probably fail, but if the other tests pass then it should be fine to import soon.

Also there might be some delay for importing due to holiday schedules.

ryan-dorazio:
> Could you add an example of solving a matrix game to mmd_nash_example.py?
>
Do you mean change the example in the file to a matrix game? Right now there is only support for sequential games so it would look like:
`game = pyspiel.load_game_as_turn_based("matrix_game")`

lanctot:
No I meant add an example of how to use it to solve matrix games in the same file, yes it's ok to do via load_as_turn_based just like in the test.. just so people see the usage example for matrix games in addition to extensive games.

lanctot:
(I guess you can't easily pull out the strategy in the matrix game case?)

ryan-dorazio:
@lanctot I added a warning if using MMD to compute an approximate Nash equilibrium and a matrix game example, also showing how to extract the policies two different ways (sequence form and behavioural form are the same in this case).

lanctot (changes requested):
Hi @ryan-dorazio, can you add MMD to the algorithms page (docs/algorithms.md)

ryan-dorazio:
> Hi @ryan-dorazio, can you add MMD to the algorithms page (docs/algorithms.md)

Done! I put it as "implemented but lightly tested," not sure what qualifies as thoroughly tested.

lanctot (approved):

lanctot:
@PaulFMMuller will take a quick look and then we should be good to import. Thanks again!

lanctot:
Actually, I had a bit of time so I imported it. I will do any back-and-forth with @PaulFMMuller internally.

Thanks for this implementation and contributing to OpenSpiel!

PR #881, opened 2022-07-10
Add pytorch NeuRD

Asugawara:
Hi! Dr.@lanctot !
I haven't seen you for a long time.
I was really struck by [the article](
https://arxiv.org/pdf/2206.15378.pdf)! So, I tried to implement NueRD in PyTorch:)

This figure is a comparison of the Exploitability averaged over 10 times of NeuRD.
Need to pay attention to PyTorch's significant figures,
Exploitability is not much different between PyTorch and TensorFlow.
However, PyTorch is about 40 times faster than TensorFlow.(only CPU)

```
TensorFlow: 472.83000090387134 sec
PyTorch: 11.534181308746337 sec

```

tf_NeuRD_vs_pt_NeuRD

```
$ sw_vers
ProductName:    macOS
ProductVersion: 11.4
BuildVersion:   20F71

$ python3 -V
Python 3.9.9

$ pip3 freeze | grep open-spiel
open-spiel==1.1.1
```

### DIFF
--- a/open_spiel/python/CMakeLists.txt
+++ b/open_spiel/python/CMakeLists.txt
@@ -254,6 +254,7 @@ if (OPEN_SPIEL_ENABLE_PYTORCH)
       pytorch/eva_pytorch_test.py
       pytorch/losses/rl_losses_pytorch_test.py
       pytorch/policy_gradient_pytorch_test.py
+      pytorch/neurd_pytorch_test.py
   )
 endif()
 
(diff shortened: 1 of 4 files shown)


lanctot:
Wow! This is amazing, thank you so much as always.

If it does not cost too much time, could you produce the same graph for Leduc just to check it with the paper and the TF impl?

Asugawara:
Hi!
I experimented with Leduc poker 5 times. The results are shown in Figure1.

figure1_leduc_5time_tf_something_wrong

Excluding the worst case.(Figure 2)
figure2_leduc_remove_worst_experiment

Figure2 shows that PyTorch performance is a little worse than TensorFlow...

Issue #883, opened 2022-07-16
Installation error (install.sh)


pklehre:
I get the following error message while running ./install.sh

```
++ pushd open_spiel/pybind11_abseil
~/private/dev/open_spiel/open_spiel/pybind11_abseil ~/private/dev/open_spiel
++ git checkout 73992b5
error: pathspec '73992b5' did not match any file(s) known to git
```

This can be fixed by adding the following at line 117 in 'install.sh' 

```
git fetch
```

lanctot:
Hmm.. odd, it works fine on the CI and local tests as is 

Can you tell me what platform you are using?

Also is this from a fresh checkout of master or rerunning from a recent pull of an outdated clone?

pklehre:
I am using Debian. Here is a minimal Dockerfile which illustrates the problem:

```
FROM debian

RUN apt-get update; apt-get install --no-install-recommends -y git ca-certificates
RUN git clone https://github.com/deepmind/open_spiel.git
RUN cd open_spiel ; \
    ./install.sh
```

pklehre:
I suspect the issue arises from the parameters `--single-branch --depth 1` passed to `git clone` in the function `cached_clone()`.

lanctot:
Hmm.. I wonder if it's due to a default setting of git from a fresh Debian install within Docker.

Can you submit a PR to add that line, it will trigger the CI tests; to confirm that it won't interfere with the non-Docker workflow?

pklehre:
In fact, adding `git fetch` did not fix the issue. I had to remove '--single-branch --depth 1'. I can prepare a PR for that line.

lanctot:
Actually, that should not be `--depth 1`, it should be `--depth 15`. Can you try with that?

lanctot:
(I meant to copy the one from the Hanabi Learning Environment slightly below that)

pklehre:
Yes, `--depth 15` works on my machine. However, isn't this problem likely to return when further commits are made to the pybind11_abseil repository?

lanctot:
Oh yes you're right (I misunderstood what the --depth did at first). Thanks!

pklehre:
I have created a new pull request which removes the --depth etc. for both pybind11_abseil and hanabi. I have also signed the CLA.

lanctot:
Fixed by https://github.com/deepmind/open_spiel/commit/2651e77362d43af5aabb4e4b78f85d77bf5cf9e1

PR #884, opened 2022-07-17
git error workaround

pklehre:
As discussed in https://github.com/deepmind/open_spiel/issues/883, I get a git error when running the install script. This seems to be related to the git clone options `--single-branch --depth 1`. The install script works fine on my machine when removing these.

### DIFF
--- a/open_spiel/scripts/install.sh
+++ b/open_spiel/scripts/install.sh
@@ -112,7 +112,7 @@ fi
 
 DIR="open_spiel/pybind11_abseil"
 if [[ ! -d ${DIR} ]]; then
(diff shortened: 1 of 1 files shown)


lanctot:
The CLA had an issue, can you look into it?

Also, can you remove the same flags from the Hanabi clone slightly below (I think it's the only other one that we fix by specific commit number).

lanctot:
Closing in favor of https://github.com/deepmind/open_spiel/pulls/885

lanctot:
Closing in favor of https://github.com/deepmind/open_spiel/pull/885

PR #885, opened 2022-07-17
Fix git2

pklehre:
Workaround for install.sh errors during git clone.


lanctot:
Thanks!

lanctot:
This is a fix for #883

PR #889, opened 2022-07-23
Adapt PPO algorithm from CleanRL to OpenSpiel. Adapt Gym Atari environment to OpenSpiel

newmanne:
@lanctot @ssokota: @gregdeon and I have adapted @vwxyzjn [CleanRL's](https://github.com/vwxyzjn/cleanrl) PPO implementation to work with OpenSpiel. In order to test the implementation, we have also wrapped Gym's Atari games into an OpenSpiel game. 

# Notes

- We included a SyncVectorEnv class modelled after [Gym's class](https://github.com/openai/gym/blob/master/gym/vector/sync_vector_env.py) but using OpenSpiel RL Environments. Using multiple environments is a big deal for performance. 
- This only works for single player games right now. We plan to get a multiplayer version running, but there are some challenges. Namely, when using a SyncVectorEnv, the "turn order" of players in each of the environments can quickly get out of sync, and that really is not trivial to resolve. The games @gregdeon and I are studying are simultaneous move games with no player elimination - this class of games can never go out of sync (the same agent will always be acting in every environment). This is all we were thinking to support, but perhaps we are overlooking a simpler way to support arbitrary turn orders (in combination with vector environments).
- Atari will only work if gym and the ROMS and stable baselines are all installed. 

# Evaluation
- Catch will converge to a score of 1.0 every episode fairly quickly (71_680 steps)
- Performance of three seeds of Breakout on TensorBoard (heavily smoothed) after 10_000_000 iterations (~8 wall time hours on our setup)
atari You can compare to [this example](https://wandb.ai/openrlbenchmark/openrlbenchmark/reports/Atari-CleanRL-PPO-vs-PPO-MultiGPU-vs-PPO-envpool--VmlldzoxOTQ4NzI3), we don't match exactly but seem to be in the right ballpark.

# Example commands:
`python ppo_example.py --game-name catch`  // runs catch
`python ppo_example.py`  // runs breakout


lanctot:
Awesome!

First thing: something went wrong with the CLA. Can you look into the message above and follow the steps, and let me know when you have and I will rerun it.

lanctot:
Would be nice to have a non-Atari test that solves a really basic single-player game. See the pytorch DQN test for how to construct one using the gambit EFG format.

lanctot:
> Awesome!
> 
> First thing: something went wrong with the CLA. Can you look into the message above and follow the steps, and let me know when you have and I will rerun it.

It seems that one of the commits is marked as an AWS user that has not signed the CLA: 
image

Unfortunately I can't import the PR until this is resolved. 

I think probably the easiest thing is to do a fresh branch of master and copy over all the files (individually-- not with the .git subdirectories) and open a fresh new PR with those files. That way, it should only show up as a single commit from the main author.

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,345 @@
...
+            else:
+                returns = torch.zeros_like(self.rewards).to(self.device)
+                for t in reversed(range(self.steps_per_batch)):
+                    next_return = next_value if t == self.steps_per_batch - 1 else returns[t + 1]
+                    nextnonterminal = 1.0 - self.dones[t]
+                    returns[t] = self.rewards[t] + self.gamma * nextnonterminal * next_return
+                advantages = returns - self.values
+
+        # flatten the batch
+        b_legal_actions = self.legal_actions_mask.reshape((-1, self.num_actions))
```
vwxyzjn, line 264: This should be `b_legal_actions_mask`

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,345 @@
...
+        return self.critic(x)
+
+    def get_action_and_value(self, x, legal_actions_mask=None, action=None):
+        if legal_actions_mask is None:
+            # All valid
+            legal_actions_mask = torch.ones((len(x), self.num_actions)).bool()
+
+        # Fill with invalids
+        INVALID_ACTION_PENALTY = -1e6
+        logits = torch.full((len(x), self.num_actions), INVALID_ACTION_PENALTY).to(self.device)
```
vwxyzjn, line 48: It's undesirable to create this `logits` full of `-1e6` every time during action sampling. It's better to save this as part of the module's buffer and reuse. See

https://github.com/vwxyzjn/gym-microrts/blob/e4e11405a36044eab49d4cd2c2ed084b019bb999/experiments/ppo_gridnet.py#L199
https://github.com/vwxyzjn/gym-microrts/blob/e4e11405a36044eab49d4cd2c2ed084b019bb999/experiments/ppo_gridnet.py#L211

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,345 @@
...
+    def get_value(self, x):
+        return self.critic(x)
+
+    def get_action_and_value(self, x, legal_actions_mask=None, action=None):
+        if legal_actions_mask is None:
+            # All valid
+            legal_actions_mask = torch.ones((len(x), self.num_actions)).bool()
+
+        # Fill with invalids
+        INVALID_ACTION_PENALTY = -1e6
```
vwxyzjn, line 47: This should be moved to the top of the file as a constant

vwxyzjn (commented):
Hey @newmanne, this is great work. I left a couple of comments for minor issues 🙂

I saw `open_spiel` also has a folder for jax learning algorithms. Just a relevant note - we are experimenting with PPO + JAX integration in CleanRL, which shows promising speed improvement w/ [Envpool](https://envpool.readthedocs.io/en/latest/). 

Don't worry about integrating it now since the prototype is in its early stage (messy code), but it might be worth integrating it in the future once we finalize it w/ more tests and documentation.

image

lanctot (changes requested):
Can you add a simple test, i.e. the one described here: https://github.com/deepmind/open_spiel/pull/889#issuecomment-1193052529

open_spiel/python/pytorch/ppo.py
newmanne: Ah, cool! Thanks for the pointer

lanctot:
Awesome, thanks. Can you add PPO to docs/algorithms.md and Atari to docs/games.md?

lanctot (approved):

lanctot:
Thanks @vwxyzjn  for taking a look!

Apologies @newmanne for the delay, I was on vacation for a large part of July. I'm back now and will ask one of the team to take a quick look as well.

open_spiel/python/examples/ppo_example.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # Algorithm specific arguments
+    parser.add_argument("--num-envs", type=int, default=8,
+        help="the number of parallel game environments")
+    parser.add_argument("--num-steps", type=int, default=128,
+        help="the number of steps to run in each environment per policy rollout")
+    parser.add_argument("--anneal-lr", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="Toggle learning rate annealing for policy and value networks")
+    parser.add_argument("--gae", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="Use GAE for advantage computation")
```
lanctot, line 56: Can you use absl flags to be consistent with the rest of the OpenSpiel code?

lanctot (changes requested):
Hi guys, I noticed you're using four space indent, but we use 2 space indentation in OpenSpiel, can you change that to use the 2 spaces?

Also noted a few other things in separate comments

vwxyzjn:
Hi also a quick comment: would it be possible to add more benchmark? I am quite interested to see the performance of the agent in the environments that OpenSpiel provides. E.g,. see our `ppo_atari.py` [docs](https://docs.cleanrl.dev/rl-algorithms/ppo/#experiment-results_1).

image

We certainly don't need to do it in this PR, but having more benchmark would be quite nice. 

Ideally, it would be great if you can contribute the tracked experiments to the [Open RL Benchmark](https://wandb.ai/openrlbenchmark/openrlbenchmark/reportlist), which makes everything about the experiments super transparent. It is leveraging wandb (a proprietary service), however, so no worries if this is difficult.

lanctot:
Hmmm, test failed in Colored Trails -- that's probably a bug on our end, I'll look into it today.

lanctot:
> Hmmm, test failed in Colored Trails (due to observing a utility greater than the maximum) -- that's probably a bug on our end, I'll look into it today.

Confirmed bug on our side, fix is lines 186-187 in `colored_trails.h` here: https://github.com/deepmind/open_spiel/pull/900/files.

jhtschultz (commented):
Thanks for the submission! Cool to see someone integrate Atari with OpenSpiel. Left some minor comments. Didn't dig into the PPO implementation as that appears to have been forked from a stable version.

Please run the code through yapf (https://github.com/google/yapf/) or pylint to conform with the Google styleguide (https://google.github.io/styleguide/pyguide.html).

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,344 @@
...
+
+  def get_action_and_value(self, x, legal_actions_mask=None, action=None):
+    return self.network.get_action_and_value(x, legal_actions_mask, action)
+
+  def step(self, time_step, is_evaluation=False):
+    if is_evaluation:
+      singular_env = False
+      if not isinstance(time_step, list):
+        time_step = [time_step]
+        singular_env = True
```
PaulFMMuller, line 192: I would suggest fixing input and output types to always be lists and doing these transformations outside of the algorithm. Changing input/output types may lead to confusing behavior.

@lanctot do you have a preference ?

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,344 @@
...
+    super().__init__()
+
+    self.input_shape = input_shape
+    self.num_actions = num_actions
+    self.num_players = num_players
+    self.player_id = player_id
+    self.device = device
+
+    # Training settings
+    self.num_envs = num_envs
```
PaulFMMuller, line 138: Am I correct in assuming this is the number of consecutive environment steps in each batch element  ? 
I would suggest a more explicit name - num_env_steps_in_batch_elems. Or anything more elegant. ;)

open_spiel/python/pytorch/ppo.py
PaulFMMuller, line 192: I would suggest setting input and output types to always be lists and doing these transformations outside of the algorithm. Changing input/output types may lead to confusing behavior.

@lanctot do you have a preference ?

open_spiel/python/pytorch/ppo.py
```diff
@@ -0,0 +1,344 @@
...
+    self.total_steps_done += self.num_envs
+    self.cur_batch_idx += 1
+
+      
+  def learn(self, time_step):
+    next_obs = torch.Tensor(np.array([ts.observations['info_state'][self.player_id] for ts in time_step])).to(self.device)
+
+    # Annealing the rate if instructed to do so.
+    if self.num_annealing_updates is not None:
+      frac = 1.0 - (self.updates_done) / self.num_annealing_updates
```
PaulFMMuller, line 238: Make sure LR doesn't go to 0 / negative ?

open_spiel/python/pytorch/ppo.py
gregdeon: No - this is the number of environments that are run concurrently in each step. It's a tweak to PPO that we haven't seen elsewhere in OpenSpiel, but tends to have a significant impact on performance. We'll add a note in the PPO docstring.

open_spiel/python/pytorch/ppo.py
gregdeon: The intention here is for PPO to work with a "vector environment", which allows multiple environments to run in parallel, while still working with a single environment using the same API (both input and output) as other RL algorithms in OpenSpiel.

lanctot:
Thanks guys. I've started the tests, but they will probably fail. There was a change on GitHub's configuration that broke our tests today. I finally got it fix and updated master a few minutes ago. You'll probably need to pull changes from master in order for the tests to pass (which would be a good thing anyway since this PR's branch has some large changes and was started a while back).

open_spiel/python/pytorch/ppo.py
vwxyzjn: It wouldn't go to 0 / negative when `num_annealing_updates=num_updates`:

```
learning_rate = 2.5e-4
total_timesteps = 500000
num_envs = 4
num_steps = 128
num_updates = total_timesteps // (num_envs * num_steps)
num_annealing_updates = num_updates
print(f"num_updates={num_updates}")
for update in range(num_updates):
    # Annealing the rate if instructed to do so.
    frac = 1.0 - (update) / num_annealing_updates
    lrnow = frac * learning_rate
print(f"frac={frac}")
frac = 1.0 - (update+1) / num_annealing_updates
print(f"do an extra annealing step is not necessary because frac would become {frac}")
```
```
num_updates=976
frac=0.0010245901639344135
do an extra annealing step is not necessary because frac would become 0.0
```

However, if `num_annealing_updates` is set to `num_updates - 2`, then 0 / negative happens. Maybe instead of a `ValueError` we can check the `num_annealing_updates` is greater than `num_updates - 1` @newmanne?

```
num_updates=976
frac=-0.0010266940451746365
do an extra annealing step is not necessary because frac would become -0.002053388090349051
```

lanctot:
Hi guys, can you update the conversations on github PR thread to either reply or mark the as resolved?

open_spiel/python/games/atari.py
```diff
@@ -0,0 +1,162 @@
...
+    """Applies the specified action to the state."""
+    observation, reward, done, info = self.env.step(action)
+    self.last_info = info
+    self.last_reward = reward
+    self.tracked_rewards += reward
+    if done:
+      self._is_terminal = True
+    self.observation = observation # Store this for later
+
+  def information_state_tensor(self, player_id):
```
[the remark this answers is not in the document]
newmanne: Changes have been made accordingly

open_spiel/python/games/atari.py
```diff
@@ -0,0 +1,162 @@
+import gym
+import numpy as np
+import pyspiel
+
```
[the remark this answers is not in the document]
newmanne: Added

open_spiel/python/pytorch/ppo.py
newmanne: The input has been switched to lists only

open_spiel/python/games/atari.py
```diff
@@ -0,0 +1,162 @@
...
+    self.capture_video = params.get('capture_video', False)
+    self.run_name = params.get('run_name', 'default')
+    self.use_episodic_life_env = params.get('use_episodic_life_env', True)
+
+    env = gym.make(self.gym_id) 
+    env = gym.wrappers.RecordEpisodeStatistics(env)
+    if self.capture_video and self.idx == 0:
+      env = gym.wrappers.RecordVideo(env, f"videos/{self.run_name}")
+    
+    # Wrappers are a bit specialized right nwo to Breakout - different games may want different wrappers.
```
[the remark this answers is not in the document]
newmanne: We've updated the comment to provide a bit more detail. We don't have advice on how to set wrappers for each individual atari game, we just want to note that this is the "standard" set we borrowed from CleanRL and YMMV for other games (e.g., adjusting frameskips in space invaders)

open_spiel/python/pytorch/ppo.py
newmanne: We've moved the learning rate annealing outside of the agent. By default there is no annealing, and it can be put into the training loop if the user wants (as seen in `ppo_example.py` with the `anneal_lr` flag.

newmanne:
@lanctot @gregdeon I think everything should be resolved now

lanctot (approved):

Issue #891, opened 2022-07-24
Implement the sample-based NeuRD


godmoves:
Hi, I have just read the [DeepNash](https://arxiv.org/abs/2206.15378) paper and I am quite interested in the NeuRD algorithm. But it seems the sample-based version is not implemented in the OpenSpiel repository, so I am trying to do it myself.

I find there are some discussions about it in #148 and #316, so I just follow the instructions in #316 and add a **PyTorch** version. I choose to use the **regular policy entropy** instead of the q-value entropy used in https://arxiv.org/abs/2002.08456 because I still have some questions about it and decide to leave it for a future task.

To compare the results between different algorithms, I run each algorithm 5 times using the OpenSpiel PyTorch code and reproduce the results in [Figure 3](https://arxiv.org/abs/1908.09453) (except NFSP). The hyperparameters are from #155. The final performances appear to be basically the same as in the paper, except for the RPG, which is **worse**.

Due to the different entropy types used, I need to do more hyperparameter searches before I finally confirm the performance of my implementation of NeuRD. I will update the results once I have done that.

![image](https://user-images.githubusercontent.com/25997471/180639084-c026e79b-2c70-43e7-8c26-057db0648209.png)

godmoves:
To verify the RPG result of the PyTorch code, I also ran an extra experiment on the Tensorflow code. But the result is still worse than the value reported in the paper.

It is mentioned that the learning rate is decayed to 0 in the [RPG paper](https://arxiv.org/abs/1810.09026). Is this also used in the OpenSpiel experiments?

> These updates were done using separate SGD optimizers with their respective learning rates of fixed 0.001 for policy evaluation, and annealed from a starting learning rate to 0 over 20M steps for policy improvement.

There is a temperature for softmax, too.

> In addition, the agent’s policy is controlled by a(n inverse) temperature added as part of the softmax operator. The temperature is annealed from 1 to 0 over 1M steps to ensure adequate state space coverage.

By the way, the hyperparameter I used, and all learning rate is fixed during the training:
```
loss_str:rpg
batch_size:64
critic_lr:0.01
pi_lr:0.01
entropy_cost:0.01
num_critic_before_pi:32
```
![image](https://user-images.githubusercontent.com/25997471/180682456-c091de4b-637d-4f2a-bd98-9da4959d954d.png)

lanctot:
Hi @godmoves, this is awesome! It's been repeatedly asked for (as you found out), so it would be great to finally include in the library.

[...]

PR #892, opened 2022-07-26
Add sample-based NeuRD loss

godmoves:
This pull request adds the sample-based NeuRD loss with regular policy entropy. More experiment results are in #891.

### DIFF
--- a/open_spiel/python/pytorch/losses/rl_losses.py
+++ b/open_spiel/python/pytorch/losses/rl_losses.py
@@ -42,6 +42,15 @@ def _assert_rank_and_shape_compatibility(tensors, rank):
                        (tensor.shape, tmp_shape))
 
 
+def thresholded(logits, regrets, threshold=2.0):
+  """Zeros out `regrets` where `logits` are too negative or too large."""
+  can_decrease = logits.gt(-threshold).float()
+  can_increase = logits.lt(threshold).float()
+  regrets_negative = regrets.clamp(max=0.0)
+  regrets_positive = regrets.clamp(min=0.0)
+  return can_decrease * regrets_negative + can_increase * regrets_positive
+
+
 def compute_baseline(policy, action_values):
   # V = pi * Q, backprop through pi but not Q.
   return torch.sum(torch.mul(policy, action_values.detach()), dim=1)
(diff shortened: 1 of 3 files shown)


lanctot:
Great! Can you add a test to this file: https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/pytorch/policy_gradient_pytorch_test.py

Maybe you could run it on Kuhn poker (for a low number of iterations, so the test doesn't take too long) and then check the exploitability is lower than some threshold?

godmoves:
I added the results of NeuRD on Kuhn poker in #891.

lanctot:
@godmoves the tests failed.. any idea why? (You can see the logs by clicking on them below)

godmoves:
@lanctot It seems that we step all players in the Hanabi test, and this causes some `None` values to the agent's action history. An error occurs when we try to convert the action history to tensor.

This is not triggered before because the game ends in just 7 steps. 7 is small than the batch size so no training will happen. After the NeuRD loss test is added, the random seed changes and the game lasts more than 16 steps, and this causes the `None` value error mentioned above.

https://github.com/deepmind/open_spiel/blob/ae9371a1c92a6bfb2312e8315882bfe7d3e69a1a/open_spiel/python/pytorch/policy_gradient_pytorch_test.py#L102-L106

lanctot:
I see, ok how about just having a separate test for NeuRD in that file, which runs it on Kuhn for 100 episodes and checks the exploitability is less than 1?

lanctot:
The main purpose is to have a test in place that runs it, so that if someone else makes a change and it breaks NeuRD it will get caught within this test.

godmoves:
@lanctot  I added a test of NeuRD on Kuhn poker. But I'm not sure if it's appropriate to use the code from the example file in my test.

The upper bound value of 0.7 is determined by running the test with 500 different seeds, and it seems 0.7 may be a good value to check potential changes that break the NeuRD code.

o2AWci5tyB

lanctot:
Should be ok as long as the seed is fixed in the test, it will be fully deterministic.

lanctot:
Thanks, I agree 0.7 is a good choice given your graph.

PR #926, opened 2022-09-02
Remove outdated comments describing chess action encoding

VitamintK:
Looks like https://github.com/deepmind/open_spiel/commit/479b1941c4bc9b6711e233266275e9748b41b4d8 changed the way that chess actions were encoded so that these comments are outdated. Also delete an enum that doesn't seem to be used anywhere.


lanctot:
Thanks! (There is currently a known issue with the tests fixed by https://github.com/deepmind/open_spiel/pull/925)

PR #935, opened 2022-09-26
add relu to torch mlp model

mattrek:
Update alpha_zero_torch C++ mlp model construction to match the python version.


mattrek:
Will resubmit after cleaning up my commit email to satisfy the CLA

PR #776, opened 2022-01-27
Add support for non-deterministic games to AlphaZero

egeozsoy:
Addresses the issue #774.

PR #936, opened 2022-09-27
add relu to torch mlp model

mattrek:
Update alpha_zero_torch C++ mlp model construction to match the python version.


mattrek:
Looks like a new version of jax was released yesterday (jax-0.3.18, Sep 26) and... I think what's happening is that it is getting pulled into the env during the build install, clobbering the expected version (jax-0.3.14)

mattrek:
clu-0.0.6 requires flax (https://github.com/google/CommonLoopUtils/blob/v0.0.6/setup.py#L49)

current version of flax = 0.6.0
flax-0.6.0 requires jax>=0.3.16 (https://github.com/google/flax/blob/v0.6.0/setup.py#L29)
flax-0.5.3 requires jax>=0.3.2 (https://github.com/google/flax/blob/v0.5.3/setup.py#L29)

lanctot:
Ah yes this seems to happen every once in a while.. I'll take a look in the next few days. Thanks!

mattrek:
Ok, thanks - I was able to get checks to pass on a private branch by pinning to flax-0.5.3.  I submitted a separate pr in case that would be a good enough fix for now.

lanctot:
Excellent, whenever this happens I do exactly that by trial and error .. so this saves me the work this time. Much appreciated!

lanctot:
@christianjans can you take a quick look at this PR?

christianjans:
Thanks for the change, I will take a look at it later today.

open_spiel/algorithms/alpha_zero_torch/model.cc
```diff
@@ -371,15 +384,10 @@ std::vector<torch::Tensor> ModelImpl::forward_(torch::Tensor x,
...
-        x = layers_[i]->as<torch::nn::Linear>()->forward(x);
-      } else if (i >= num_torso_blocks_ + 1) {
-        output = layers_[i]->as<MLPOutputBlockImpl>()->forward(x, mask);
-      } else {
-        x = layers_[i]->as<torch::nn::Linear>()->forward(x);
-      }
+    for (int i = 0; i < num_torso_blocks_; i++) {
+        x = layers_[i]->as<MLPTorsoBlock>()->forward(x);
     }
+    output = layers_[num_torso_blocks_]->as<MLPOutputBlockImpl>()->forward(x, mask);
```
christianjans, line 390: Note that if the above comment is accepted, then we would need to change the implementation of this `forward_` method:

```suggestion
    x = layers_[0]->as<MLPTorsoBlock>()->forward(x);
    for (int i = 1; i < num_torso_blocks_ + 1; i++) {
        x = layers_[i]->as<MLPTorsoBlock>()->forward(x);
    }
    output = layers_[num_torso_blocks_]->as<MLPOutputBlockImpl>()->forward(x, mask);
```

christianjans (commented):
Awesome addition, thanks! Just a comment on the `nn_depth` (`num_torso_blocks_`) hyper-parameter.

open_spiel/algorithms/alpha_zero_torch/model.cc
mattrek: Thanks for the comments.  Oops - I did it this way because I thought I was matching the python version, but looking again (https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/model.py#L225) I see that the python version is cascading nn_depth layers after an input layer, so... yes, your suggestion does seem more consistent.  Will try to commit a patch soon

mattrek: Oh wait, I guess I'm confused:  In the python code, 'observations' is a tf,placeholder, which I think means it's not a layer, just a specification of the input size. Similar for the python 'conv2d' net (not yet implemented for C++ torch)  And then 'resnet' does have a separate input block.   Anyway, I'm indifferent about this - will still plan to commit a patch (for C++), as I do think your suggestion seems natural.  But then the python might need a separate update to be consistent?

lanctot:
Geez, now rust dependencies seem to be broken on Ubuntu 22.04. You've gotten quite unlucky with the tests....! I will take a look this week.

mattrek:
Yeah, no worry - I may get a chance to look too

lanctot:
I believe this fixes it: https://github.com/deepmind/open_spiel/pull/939

I will import and update github soon, and then you can pull from master... but you can also just make that one line change locally and tests should not fail for that reason anymore.

open_spiel/algorithms/alpha_zero_torch/model.cc
christianjans: So sorry for the late reply! I had little reception this past weekend. Thanks for your comment, I certainly see your point about consistency with the Python version.

However, for consistent meaning of the hyper-parameter across LibTorch models, and because of how the "legacy" non-residual model was implemented in #776 (i.e with the number of torso layers not including the input and output layers), let's go with the suggestions mentioned here (I see you have already implemented them in faca14c, thanks so much!).

mattrek: Thanks!  The requested changes have been submitted, so marking this resolved

open_spiel/algorithms/alpha_zero_torch/model.cc
```diff
@@ -298,9 +311,9 @@ ModelImpl::ModelImpl(const ModelConfig& config, const std::string& device)
     register_module("layers", layers_);
 
   } else if (config.nn_model == "mlp") {
-    layers_->push_back(torch::nn::Linear(input_size, config.nn_width));
     for (int i = 0; i < num_torso_blocks_; i++) {
-      layers_->push_back(torch::nn::Linear(config.nn_width, config.nn_width));
+      layers_->push_back(
+          MLPTorsoBlock((i == 0 ? input_size : config.nn_width), config.nn_width));
     }
```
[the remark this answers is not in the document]
mattrek: Done, including renaming `MLPTorsoBlock` to `MLPBlock`

mattrek:
Yeah, should be all set now.

Issue #942, opened 2022-10-06
Shared library in Windows?


roblim1:
Referencing #933, I am trying to build open_spiel as a shared library in Windows in MS Visual Studio 2019, but get the following error:

`Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK2019	unresolved external symbol "void __cdecl open_spiel::LogUsage(void)" (?LogUsage@open_spiel@@YAXXZ) referenced in function "class std::shared_ptr<class open_spiel::Game const > __cdecl open_spiel::LoadGame(class std::map<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,class open_spiel::GameParameter,struct std::less<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >,class std::allocator<struct std::pair<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const ,class open_spiel::GameParameter> > >)" (?LoadGame@open_spiel@@YA?AV?$shared_ptr@$$CBVGame@open_spiel@@@std@@V?$map@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@VGameParameter@open_spiel@@U?$less@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@2@V?$allocator@U?$pair@$$CBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@VGameParameter@open_spiel@@@std@@@2@@3@@Z)	C:\Users\user\apps\spiel_win\open_spiel\open_spiel\out\build\x64-Debug\open_spiel	C:\Users\user\apps\spiel_win\open_spiel\open_spiel\out\build\x64-Debug\spiel.cc.obj	1	
`

The build produces an open_spiel.dll file, but when I add that file as a shared library in my project, I get this error:

`Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK1107	invalid or corrupt file: cannot read at 0x370	some_project	C:\some_project\dependencies\open_spiel\lib\open_spiel.dll	1`

I can build open_spiel without the shared library option without any problems.  

Any suggestions on how to get the shared library to build correctly?

roblim1:
Looks like I am able to build a shared library in Windows properly, per Issue #825.

lanctot:
@roblim1 ok but did you have to move utils/usage_logging.h and .cc to the core?

roblim1:
[...]

Issue #933, opened 2022-09-18
Can I use open_spiel as a shared library on windows?


87a:
[library.md](https://github.com/deepmind/open_spiel/blob/master/docs/library.md) only provides this on linux and macos, so are there any methods to do this on windows? or it only supports linux and macos?

lanctot:
We've never tested it, but I would be surprised if it didn't work. 

You just have to have the `BUILD_SHARED_LIB` environment variable set to `ON`, or you can manually set it on line 335 of CMakeLists.txt here: https://github.com/deepmind/open_spiel/blob/7f9a27cbca89f412802fb0d5283de8acaa19c4e2/open_spiel/CMakeLists.txt#L335

```
set (BUILD_SHARED_LIB ON)
```

On Windows, the extension would be `.dll`, but everything else should be the same. Let us know if it works!

87a:
Thanks, by doing it I can see `open_spiel.lib` and `open_spiel.dll` in the `out/build/x64-Debug` folder, and can you provide an example of how to use it in cmakelist.txt?

lanctot:
Once you have OpenSpiel as a shared library, you probably can set it up via its own (separate) project in Visual Studio. You'd need to setup the project to link against that library and set the include paths.  That is probably the easiest way to do it. See [here](https://stackoverflow.com/questions/7845886/linking-dll-in-visual-studio) and [here](https://stackoverflow.com/questions/2676417/how-do-include-paths-work-in-visual-studio).

You can do it as a separate CMake project too, if you want (see this tutorial for basic usage: https://cmake.org/cmake/help/latest/guide/tutorial/index.html), but the main CMakeLists.txt it would look something like this:

```
cmake_minimum_required(VERSION 3.10)

# set the project name
project(87a_project)

# This has to point to the root of OpenSpiel so you can using #include in the same way as the OpenSpiel headers
include_directories(open_spiel)

# This tells the compiler where to find the library
link_directories(out/build/x64-Debug)

# This adds the link command when building
link_libraries(open_spiel)

# add the executable
add_executable(87a_main 87a_main.cc)
```

Hope this helps!

PR #943, opened 2022-10-07
Moved logging utils to core, which crashes otherwise for shared libra…

roblim1:
…ry tests in Windows.


lanctot:
Hi @roblim1, can you click on the link above to fix the failed CLA invocation?

lanctot:
Just a note for reference purposes, this fixes is a problem that has arisen a few times:

- https://github.com/deepmind/open_spiel/issues/942
- https://github.com/deepmind/open_spiel/issues/933

PR #949, opened 2022-10-14
Jax Implementation of ARMAC

wbrenton:
Google colab containing preliminary implementation of ARMAC in Jax


lanctot:
Nice, thanks!! 

However since it's in a colab form it's a bit hard to look through / comment over github. 

Can you just copy it to a python file that you submit separately? It does not have to run or be properly tested at this stage if we're just going through it to go over the code. No rush on this of course because it might take us some time before looking through it.

wbrenton:
Here it is in .py form. I've updated it to use SARSE for training the critic, I look to here from you on next steps. Thanks!

Issue #952, opened 2022-10-19
Loading sample EFG files throws an exception


votroto:
Trying to load sample games in the Gambit EFG format from Gambit's documentation results in a thrown exception.

Here are their sample games
https://gambitproject.readthedocs.io/en/v16.0.2/samples.html

Here is the affected code
```
import pyspiel


efg = """
EFG 2 R "Bagwell Commitment and Unobservability Example" { "Player 1" "Player 2" }
"Stackelberg game with imperfectly observed commitment, from Bagwell (1993)"

p "" 1 1 "" { "S" "C" } 0
c "" 1 "" { "s" 99/100 "c" 1/100 } 0
p "" 2 1 "" { "S" "C" } 0
t "" 1 "SS" { 5, 2 }
t "" 2 "SC" { 3, 1 }
p "" 2 2 "" { "S" "C" } 0
t "" 1 "SS" { 5, 2 }
t "" 2 "SC" { 3, 1 }
c "" 2 "" { "s" 1/100 "c" 99/100 } 0
p "" 2 1 "" { "S" "C" } 0
t "" 3 "CS" { 6, 3 }
t "" 4 "CC" { 4, 4 }
p "" 2 2 "" { "S" "C" } 0
t "" 3 "CS" { 6, 3 }
t "" 4 "CC" { 4, 4 }
"""

pyspiel.load_efg_game(efg)
```

Which throws
```
$ python main.py 
OpenSpiel exception: /project/open_spiel/games/efg_game.cc:663 CHECK_TRUE(ParseDoubleValue(NextToken(), &utility))
 while parsing line #7:
p "" 2 1 "" { "S" "C" } 0

Traceback (most recent call last):
  File "/home/votroto/spiel_bug_report/main.py", line 25, in <module>
    pyspiel.load_efg_game(efg)
pyspiel.SpielError: /project/open_spiel/games/efg_game.cc:663 CHECK_TRUE(ParseDoubleValue(NextToken(), &utility))
 while parsing line #7:
p "" 2 1 "" { "S" "C" } 0
```

votroto:
Sure, seems doable. I'll have a look tomorrow.

Agreed, the format is not well documented. I'll use their new-ish pygambit library as a reference.

PR #954, opened 2022-10-19
allow commas in efg payoffs

votroto:
Patch to enable commas in payoff lists of games in the EFG format. This change aims to make the parser closer in behavior to the one in the `pygambit` library.

The real gambit parser accepts at most one comma after every number in the payoff list, including trailing commas.

This PR was created in response to [issue 952](https://github.com/deepmind/open_spiel/issues/952#issuecomment-1283295683).


lanctot:
Awesome, thanks!!

lanctot:
Hi @votroto, just quickly checking in-- is this PR ready to import now?

votroto:
Hi, it should be ready as it is. The failing sample games with commas now work. The crash on whitespace-files was a simple fix. 

I do not currently have time to continue with the extra features and fixes I wanted (uniqueness in labels, payoffs in non-terminal nodes, implicit action lists), but I can open a new PR when I do.

PR #956, opened 2022-10-24
Fix random bug in alpha_zero_torch (pun intended)

mattrek:
Without a seed, `std::mt19937` constructs with a default constant, and so all actors were using the same random sequence, generating duplicate trajectories.

Also tweaked the child selection policy to PUCT for the AZ bot in `alpha_zero_torch_game_example.cc`.


lanctot:
Thanks! (For the contribution *and* the pun! :))

PR #977, opened 2022-12-12
Fix exception causes in visualization.py

cool-RR:
@lanctot  This is the continuation of #973 and extends this fix to `visualization.py`. These were the only two instances of this problem in the project.


lanctot:
Awesome, thanks again!

Issue #992, opened 2023-01-12
Comments contradict the code logic - LegalActions vs. ChanceOutcomes


Henry-E:
Just checking if I'm correct in my interpretation of this
https://github.com/deepmind/open_spiel/blob/e75bdf114de32c2211edf36443703a6e8846a3cb/open_spiel/spiel.h#L263-L270
This comment above the `LegalActions()` function seems incorrect, since the `LegalActions(Player)` function checks for whether it's a chance node or not.
https://github.com/deepmind/open_spiel/blob/e75bdf114de32c2211edf36443703a6e8846a3cb/open_spiel/spiel.h#L255-L257
And instead it collects the actions from `ChanceOutcomes()` as would be sensible and much less effort for people implementing games (don't need to re-include the logic of ChanceOutcomes in LegalActions).
https://github.com/deepmind/open_spiel/blob/e75bdf114de32c2211edf36443703a6e8846a3cb/open_spiel/spiel.h#L622-L628

lanctot:
Sorry, you will have to clarify because I don't understand. 

What is your precise question? Are to you saying that the `All games should implement this function` is misleading, since in games with chance events, you just end up implementing `ChanceOutcomes` for the chance events?

Can you elaborate and be more specific about what you're claiming is incorrect in the comment?

lanctot:
> And instead it collects the actions from `ChanceOutcomes()` as would be sensible and much less effort for people implementing games (don't need to re-include the logic of ChanceOutcomes in LegalActions).

Exactly, this is what most (almost all) of the games with chance events do.

Maybe you can suggest a rewording of the comment to "correct" it (or at least help me understand)?

lanctot:
Maybe this will help: not all calls to get the legal actions will be done via `LegalActions(Player player)`. It is often the case that `LegalActions()` is called directly.

lanctot:
Here is an example: https://github.com/deepmind/open_spiel/blob/e75bdf114de32c2211edf36443703a6e8846a3cb/open_spiel/games/pig.cc#L207. 

Note the first few lines: the function checks if it's a chance node, if so, returns `LegalChanceOutcomes` which calls `ChanceOutcomes`.

Henry-E:
[...]

PR #993, opened 2023-01-12
Update liar's dice to use `LegalChanceOutcomes()`

Henry-E:
Updates liar's poker to use the best practice approach for returning legal chance outcomes from `LegalActions()`. Closes:https://github.com/deepmind/open_spiel/issues/992


lanctot:
Thanks!!

PR #997, opened 2023-01-18
add game: squadro

giogix2:
Hi all.

I've added the implementation of a game called Squadro. It's a two-player, perfect information game. I believe it has some similarities with Quoridor, which is already present in open_spiel. Just like Quoridor, it's distributed by Gigamic. Though, I'm not sure whether it's protected by copyright. 

It doesn't have a Wikipedia page, but you can find it here in boardgamegeek https://boardgamegeek.com/boardgame/245222/squadro.

Hopefully, this is helpful and can be added to the collection of games.

Best,
Gio


lanctot:
Hi @giogix2, thanks! 

First thing, can you sign the CLA? (See the "Details" link above and follow the instructions.)

giogix2:
Hi @lanctot. Signed. 
There was also an issue with the contributor's name. Also fixed that.

PR #1006, opened 2023-01-27
remove unused ILLEGAL_ACTION_LOGITS_PENALTY constants

VitamintK:
I noticed these `ILLEGAL_ACTION_LOGITS_PENALTY` constants aren't used anywhere at the moment, so here's a pull request to delete them.

(There are `ILLEGAL_ACTION_LOGITS_PENALTY` constants that *are* used, defined in https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/dqn.py and https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/jax/dqn.py and https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/pytorch/dqn.py )


lanctot:
Thanks!

PR #1018, opened 2023-02-21
adding domino

morLev:


lanctot:
Thanks!

#1017 has been merged into master, so the problems with the tests are now fixed. Can you pull changes from master and push the merge commit?

morLev:
> Thanks!
> 
> #1017 has been merged into master, so the problems with the tests are now fixed. Can you pull changes from master and push the merge commit?

Done

lanctot:
> Done

Hmm, oddly .. the new commits are showing as part of your PR now. Ok, we'll try this and see what happens, and hopefully they get marked as reverted during the import.

I started the tests now. If they pass, then someone will do a review soon. Thanks again!

lanctot:
Hi @morLev, the tests failed because there is no playthrough for your new game.

Can you take a look at step 9 from these instructions: https://github.com/deepmind/open_spiel/blob/master/docs/developer_guide.md#adding-a-game to generate a new playthrough, then add it to your PR?

morLev:
> Hi @morLev, the tests failed because there is no playthrough for your new game.
> 
> Can you take a look at step 9 from these instructions: https://github.com/deepmind/open_spiel/blob/master/docs/developer_guide.md#adding-a-game to generate a new playthrough, then add it to your PR?

done

lanctot:
The domino_test is now failing when run on Github Actions. You can see the output by clicking on the failure link above. Does it work for you locally?

morLev:
> The domino_test is now failing when run on Github Actions. You can see the output by clicking on the failure link above. Does it work for you locally?

Hi,

Sorry, it was not working locally.
solve it by modifying returns() (to return [0, 0] on nonterminal states).

lanctot:
Great, thanks. Tests are passing now. Someone will do a review soon. Once any changes are made from review, we can import it.

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def __init__(self, iig_obs_type, params):
+        """Initializes an empty observation tensor."""
+        if params:
+            raise ValueError(f"Observation parameters not supported; passed {params}")
+
+        # Determine which observation pieces we want to include.
+        pieces = [("player", 2, (2,))]
+
+        if iig_obs_type.private_info == pyspiel.PrivateInfoType.SINGLE_PLAYER:
```
lanctot, line 260: Can you add some comments to explain the numbers here?

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
...
+
+_NUM_PLAYERS = 2
+
+# The first player to play is the one holding the highest rank tile.
+# The rank of tiles is the following:
+#   1. Highest double.
+#   2. If none of the players hold a double, then highest weight.
+#   3. If the highest weighted tile of both players has the same weight
+#      then the highest single edge of the highest weighted tile.
+
```
lanctot, line 16: Can you also add the game to the games list? In docs/games.md.

lanctot (changes requested):
A few quick comments now.

open_spiel/python/games/domino.py
jhtschultz, line 16: General comment: I think we should rename this game block_dominoes. Dominoes is a family of games and we may want to add more in the future.

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
...
+    provides_observation_tensor=True,
+    provides_factored_observation_string=True)
+_GAME_INFO = pyspiel.GameInfo(
+    num_distinct_actions=8,
+    max_chance_outcomes=len(_DECK),
+    min_utility=-69,
+    max_utility=69,
+    # first player hand: (6,6) (6,5) (5,5) (6,4) (4,5) (6,3) (4,4) , second player hand is empty. can be reduced.
+    num_players=_NUM_PLAYERS,
+    max_game_length=28,  # deal: 14 chance nodes + play: 14 player nodes
```
jhtschultz, line 55: These should be top level constants like _NUM_PLAYERS.

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
...
+        elif len(self.player_legal_actions) == 0:
+            return pyspiel.PlayerId.CHANCE
+        else:
+            return self._next_player
+
+    def _legal_actions(self, player):
+        """Returns a list of legal actions, sorted in ascending order."""
+        assert player >= 0
+        assert player == self._next_player
+        return list(range(0, len(self.player_legal_actions)))
```
jhtschultz, line 137: I don't see how this can be correct. It always returns a sequential list of actions starting from 0. 

If this is indeed a bug, it highlights the need for more thorough tests because somehow the random sims didn't throw any errors.

morLev: self.player_legal_actions contains the optional actions for the this state.
the meaning of actions 0 is depend on the actual state.
hope it make sense.

lanctot: Yeah pretty much all games have state-independent interpretations of action ID numbers and that should be the case in all games, for the reasons John says. 

I'm not sure which game it is, can you send me a link? Is it this game? [Block](https://www.dominorules.com/block)

If so, then I counted 27 tiles. These can be given ID's based on their index in that frozen set. Then you can do it two ways. Option 1: the same player moves twice. The first move is picking the tile from hand, sec

morLev: @lanctot I'm not sure which game it is, can you send me a link? Is it this game? [Block](https://www.dominorules.com/block)
yes this is the game
the only difference is how the player that plays first is chosen (you will find many ways of doing so in different resources).

how about doing something similar to option 1 but instead of indexing the tile we playing and the tile we playing on, indexing the tile we playing and the pip we playing on? 

why is good?
in your idea, we need to know what tile is connected to our open edge to find the index of possible action.

for example, if our board looks like this:
(6,6)(6,5)(5,4)(4,3)

To compute the move index it is not enough to understand that 6 and 3 are the open edges, you also need to know to what tiles these open edges are connected. I like this solution because it's very natural to just track the open edges after each move.

lanctot: By "pip", do you mean it would be 6 or 3 in your example?

What happens if there are two possible places that you can connect the same value pip.. that can happen, right? How do you resolve it? (That is the main reason I suggested the tile id for the connection because they are always unique)

jhtschultz: Was thinking the same thing. If in the process of resolving the ambiguity you end up tracking something else (like the tiles they're connected to), then you need to also somehow encode this in the action space, the meaning of which is going to be less transparent than the unique tile id.

The interpretability of the action space is usually more important than condensing it to as few actions as possible.

morLev: @lanctot 
By "pip", do you mean it would be 6 or 3 in your example?
yes.

What happens if there are two possible places that you can connect the same value pip.. that can happen, right? How do you resolve it? (That is the main reason I suggested the tile id for the connection because they are always unique)

here are some examples:

board: (2,3) (3,4) 
tiles in my hand: (2,4) (1,1) (6,6)
possible actions:  
1. (2,4), 2 
board after the move: (4,2) (2,3), (3,4) 
2. (2,4), 4
board after the move:  (2,3) (3,4), (4,2) 

board: (3,3) 
tiles in my hand: (3,4) (1,1) (6,6)
possible actions:  
1. (3, 3), 3  // the end you playing on the tile (right or left) as no meaning strategy-wise so we have single possible action.


That is the main reason I suggested the tile id for the connection because they are always unique
I think both ways can work.
consider the following situation

board after turn 1:  (0,0) 
board after turn 2:  (0,0) (0,1) 
board after turn 3:  (0,0) (0,1) (1,2)
board after turn 4:  (0,0) (0,1) (1,2) (2,3)
board after turn 5:  (6, 0) (0,0) (0,1) (1,2) (2,3)

for me, it feels more natural to think about the last action as: "put (6,0) on 0" instead of "put (0,6) on (0, 0)".

on the other hand, consider spinner variation: the first double of the board has 4 edges. 
so if use my representation. 
in my representation, playing on a spinner will be represented in the same way as playing on a regular tile.

After this, I prefer the 2 tiles representation since I think it will be easier to transfer to other variations.
I will also need to give some Id for the "empty board" (in the first turn)

lanctot: I am afraid that I am not following.

Suppose you have a board like this: 

(3,2) (2,5) (5,5) (5,3)

And you want to play "(3,4) on the 3".

How do you know which 3.. the left one, or right one?

morLev: I treat it as the same move. strategy-wise these moves are interchangeable.

lanctot: Hmm.. yeah, I see. Ok I'm not super familiar with the game -- are you sure you can't get into a situation where these two moves would not be strategically equivalent?

morLev: in domino block, I am sure it's strategically equivalent.

the are some (all-5) domino variants that it is not.
but also in all-5 in most cases, these moves will be equivalent.

about all-5: 

in all-5, players score points whenever they play a tile to the board so that the Board Count would then equal any multiple of five. The Board Count (BC) is the sum total of all available open ends on the board. 
(http://www.chessandpoker.com/dominoes-strategy.html).

image

in example 3 from the illustration, you can see that the edges of the tile (6,6) are excluded from the board count. the reason for this is that spinners' edges are "burned" when they have two arms connected to them.

now consider the following board:

(1,1) (1,2) (2,3) (3,1)
and your only playable tile is (1,4).
If you will play it on the spinner side (left) the BC will be 5, and you will get points.
If you will play it on the non-spinner side (right) the BC will be 6, and you will not get points.

lanctot: Ok, works for me. (I'm not too concerned about covering all the variants, because if it's extended, we can change the actions later.) @jhtschultz wdyt?

jhtschultz: Yep works for me too. If you don't have to disambiguate between edges then using that for the action space seems fine and we'll revisit if we implement more variants.

lanctot:
Thanks for the updates @morLev ! Can you respond to the threads now in the PR and mark any as resolved where appropriate?

open_spiel/python/CMakeLists.txt
```diff
@@ -224,6 +224,7 @@ set(PYTHON_TESTS ${PYTHON_TESTS}
   games/dynamic_routing_utils_test.py
   games/liars_poker_test.py
   games/tic_tac_toe_test.py
+  games/domino_test.py
```
jhtschultz: Please add in alphabetical order

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
+import copy
+import itertools
+
+import numpy as np
+
+import pyspiel
+
+_NUM_PLAYERS = 2
+
+# The first player to play is the one holding the highest rank tile.
```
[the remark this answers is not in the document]
jhtschultz: Please fix

open_spiel/python/games/domino.py
jhtschultz: Please fix

jhtschultz: The more I think about this the more I'm leaning towards just leaving as is and renaming if we need to in the future. @lanctot wdyt?

open_spiel/python/games/domino.py
```diff
@@ -0,0 +1,346 @@
...
+
+class Action:
+    """ represent player possible action """
+
+    def __init__(self, tile_to_put, pip_to_play_on, player, edges):
+        self.tile_to_put = tile_to_put
+        self.open_pip = pip_to_play_on
+        self.player = player
+        self.edges = edges
+        self.new_edges = self.edges_after_action()
```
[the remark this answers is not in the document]
jhtschultz: Please fix

open_spiel/python/games/domino.py
jhtschultz: Please fix

open_spiel/python/games/domino_test.py
```diff
@@ -0,0 +1,31 @@
...
+# Lint as python3
+"""Tests for Python Kuhn Poker."""
+
+from absl.testing import absltest
+from open_spiel.python.algorithms.get_all_states import get_all_states
+import pyspiel
+
+class DominoTest(absltest.TestCase):
+
+  def test_game_from_cc(self):
```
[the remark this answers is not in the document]
jhtschultz: Please fix

open_spiel/python/games/domino.py
lanctot: I am good with that. When I think of Dominoes this is the default version that comes to mind

jhtschultz:
Awesome looks good! Thanks @morLev! Great to have a dominoes game in Open Spiel :)

PR #1025, opened 2023-03-02
Changes planned for release of Ubuntu 23.04 (Lunar Lobster) / Python 3.11 support

lanctot:
So far, this changes: 
- changes in dependencies
- Add a new test for Python 3.11 to main tests
- Add Python 3.11 wheel target
- Add an install python 3.11 in install.sh (for 22.04)
- Remove Python 3.7 support (corresponding test and wheel target)
- Temporarily remove cvxopt from extra deps and disable related tests. Re-enable them later when binary wheels are available for Python 3.11. See https://github.com/cvxopt/cvxopt/issues/228 for discussion.

### DIFF
--- a/open_spiel/python/algorithms/dqn_test.py
+++ b/open_spiel/python/algorithms/dqn_test.py
@@ -14,6 +14,7 @@
 
 """Tests for open_spiel.python.algorithms.dqn."""
 
+from absl.testing import absltest
 import tensorflow.compat.v1 as tf
 
 from open_spiel.python import rl_environment
--- a/open_spiel/python/algorithms/policy_gradient_test.py
+++ b/open_spiel/python/algorithms/policy_gradient_test.py
@@ -16,6 +16,7 @@
 
 import itertools
 
+from absl.testing import absltest
 from absl.testing import parameterized
 import tensorflow.compat.v1 as tf
 


lanctot:
@StochasticEntropy, it appears that there is a problem caused by the Jax version of DeepCFR depending on tensorflow_datasets. See here: https://github.com/deepmind/open_spiel/actions/runs/4714792415/jobs/8361347425.

I noticed the other implementations of DeepCFR don't rely on tfds, even your own TF2 implementation. Can this dependency be easily removed to fix the problem?

lanctot:
> @StochasticEntropy, it appears that there is a problem caused by the Jax version of DeepCFR depending on tensorflow_datasets. See here: https://github.com/deepmind/open_spiel/actions/runs/4714792415/jobs/8361347425.
> 
> I noticed the other implementations of DeepCFR don't rely on tfds, even your own TF2 implementation. Can this dependency be easily removed to fix the problem?

Resolved. I just needed to upgrade tensorflow_datasets.

StochasticEntropy:
> > @StochasticEntropy, it appears that there is a problem caused by the Jax version of DeepCFR depending on tensorflow_datasets. See here: https://github.com/deepmind/open_spiel/actions/runs/4714792415/jobs/8361347425.
> > I noticed the other implementations of DeepCFR don't rely on tfds, even your own TF2 implementation. Can this dependency be easily removed to fix the problem?
> 
> Resolved. I just needed to upgrade tensorflow_datasets.

Great, sorry I missed the question previously. tensorflow_datasets is used in jax to convert the tensorflow tensors from data loading to numpy for working with jax, which is not needed in the tf2 implementation where we work directly with the tensorflow tensors.

PR #1035, opened 2023-03-13
UniversalPoker/ACPC fixes: fullgame+abstracted MaxGameLength, fullgame LegalActions

VitamintK:
A pull request with 3 small fixes for UniversalPoker/ACPC

1. MaxGameLength for fullgame (i.e. unabstracted game):
`MaxGameLength` previously was written only for abstracted games where the only bet size is pot-size, so it assumed that each successive bet had to double in size. This adds a fix to MaxGameLength for fullgame and FCHPA (where a half-pot bet is legal) where we assume only that each bet has to be greater than the size of the biggest blind. The size of InformationStateTensor is based on MaxGameLength, so this fixes InformationStateTensors.
2. MaxGameLength for abstracted game:
Change `length += NumPlayers()` to `length += NumPlayers() - 1` for a tighter estimate for abstracted games.
3. LegalActions for fullgame
In unabstracted games, when a hand goes to showdown, LegalActions still returned a set of raises, despite the state being terminal. This is a fix for that.


lanctot:
Thanks @VitamintK !

The tests failed because the playthrough dors not match. Can you regenerate the playthrough and add it to the PR?

VitamintK:
Done. Also added a test to [universal_poker_test.cc](https://github.com/deepmind/open_spiel/pull/1035/files#diff-8c2ac33c5241a858de3fe1534cfe8cf3bd9ba7e9de52be3fe036181b05f3ad73) and verified that it fails without the changes in this pull request.

open_spiel/games/universal_poker.cc
```diff
@@ -1091,10 +1094,16 @@ int UniversalPokerGame::MaxGameLength() const {
...
-    length += NumPlayers();  // Each player has to react
+  if ((betting_abstraction_==BettingAbstraction::kFULLGAME) || (betting_abstraction_==BettingAbstraction::kFCHPA)){
+    // with fullgame, the longest game comes from each player can bet/raise the big blind every action.
+    // with FCHPA, the longest game is when each player bets/raise half-pot every action.
+    // however, for now we'll just use the fullgame value for FCHPA too, although it is a big overestimate.
+    length += (maxStack+maxBlind-1)/maxBlind;
+  } else {
+    while (maxStack > maxBlind) {
+      maxStack /= 2.0;             // You have always to bet the pot size
+      length += NumPlayers() - 1;  // 1 player bets, and n-2 players call
```
jhtschultz, line 1105: Why is it n-2 players call and not n-1?

jhtschultz:
Looks good, just left a comment with one question. And thanks for adding the test!

One other thing, please run it through a linter (see recently added bullet 8 from https://github.com/deepmind/open_spiel/blob/master/docs/developer_guide.md#adding-a-game for links to linters) since there's a few style guide things I think the linter will pick up on. Thanks

open_spiel/games/universal_poker.cc
VitamintK: Good question. Here's my thought process. Also let me know if you see anything wrong with the reasoning.

The longest game (with a single betting round, for simplicity) consists of:

n-1 players checking,  
1 player betting, n-2 players calling,  
1 player raising, n-2 players calling,  
etc...,  
1 player raising, n-1 players calling

so each min-bet/min-raise is succeeded by n-2 calls (except for the last bet, which is succeeded by n-1). In addition, there are n-1 checks in the beginning. These n-1 checks + the final 1 call add up to n, which is accounted for a few lines above in the code (`// Check Actions`).

VitamintK: Actually it's probably convenient for future readers of the code if I just put this comment as a code comment. What do ya think?

VitamintK:
I ran downloaded cpplint.py and ran `python cpplint.py open_spiel/games/universal_poker.cc` and there was no output. Does that mean it passed the linter?

jhtschultz:
Hmm I'd be surprised if there were no output. Did you just download the file or pip install cpplint? That could be it; my understanding is that you're supposed to run it as a command line tool.

VitamintK:
Ah, I figured out the problem. The github.com/google/styleguide cpplint is mostly unmaintained and requires python 2, and fails silently when ran with python 3 😑: https://github.com/google/styleguide/issues/132

There's a [fork](https://github.com/cpplint/cpplint) (which is what you get from `pip install cpplint` that works with python3, so I'll run that.

As per https://github.com/google/styleguide/pull/528#issuecomment-592315430 it's not clear whether the google/styleguide cpplint will be maintained going forward, so you might want to change the link in bullet 8 to the `pip install cpplint` fork (maybe [this link](https://pypi.org/project/cpplint/)?), to prevent future contributors from getting the silent failure.

open_spiel/games/universal_poker.cc
VitamintK: > First, the calculation on line 1101 only accounts for the number of raises, not all of the n-2 calls in between. So I believe that has to be updated.

ah yes, thanks. 😅 good catch! 

> I think this is more intuitive than dividing the max stack size in half.

Yup, definitely. The logic in your snippet all looks good, thanks!

> Funny how even these simple things can be tricky!

On the bright side, this has finally made me learn how to calculate a pot-sized bet.

jhtschultz:
Ahh thanks for catching that. Yeah good call I just sent out a CL to update the link in the dev guide.

jhtschultz (approved):
LGTM

PR #1038, opened 2023-03-19
Eligibility traces for tabular q learning and sarsa

giogix2:


giogix2:
Hi all,

I've added the implementation for eligibility traces in the tabular_q_learning and tabular_sarsa algorithm. I was doing this for some personal project and I thought it could be useful for other people as well.

I've tested these implementations through several experiments, observing the cumulative reward plot. I can confirm it works, but this was mostly for the single player games (e.g., catch). It seems to work for other games as well (e.g., tic tac toe), but it's a bit more difficult to verify. 

Let me know if this is something you'd like to have in the code, and if any change/addition is required.

Best,
Gio

lanctot:
Cool, thanks!

lanctot (changes requested):
Hi @giogix2,

Can you add two basic tests that use eligibility traces? Maybe something as simple as playing against uniform random in a simple game ?

lanctot (changes requested):
Also can you check the style conforms to the Google style using cpplint? 
https://pypi.org/project/cpplint/

giogix2:
Sure, I'll add a few tests.
I usually run the cpplint before the last commit, but I might have missed it last time.

lanctot:
> I usually run the cpplint before the last commit, but I might have missed it last time.

You didn't necessarily ;-) It's just become so common that I ask people to do it by default.. :)

bartdevylder:
Thanks for adding these tests and examples!

if you can fix the failing build_and_test reported above:
tabular_q_learning_example.cc:141:36: error: expected ';' after expression

I think we're ready to import it

giogix2:
Yep. The failure is just because of a missing semicolon. I'm also doing some minor refactoring, and push the final version.

giogix2:
I've added a few tests for single and multi-player games (Catch and Tic Tac Toe). While doing this, I've also added a few unit tests for the existing SARSA and Q-learning algorithms.

Since these algorithms can be slow at times, I've added examples in the existing tabular_sarsa_example_example.cc and tabular_q_learning_example.cc files, which also show the effectiveness of the algorithms, without slowing down the testing pipeline too much.

open_spiel/algorithms/tabular_sarsa.cc
```diff
@@ -163,7 +163,30 @@ void TabularSarsaSolver::RunIteration() {
...
+          player != next_state->CurrentPlayer() ? -lambda_ : lambda_;
+      eligibility_traces_[{key, curr_action}] += 1;
+      std::string state;
+      Action action;
+      double prev_q_val_tmp;
+
+      for (auto q_cell : values_) {
+        state = q_cell.first.first;
+        action = q_cell.first.second;
+        prev_q_val_tmp = q_cell.second;
```
bartdevylder, line 182: Hi Giovanni, thanks for the changes.
Quick question regarding the prev_q_val_tmp variable as it is unused. 
Could this variable just be removed or should it actually have been used in the update below?

giogix2: Hi.
I missed that. It was there for debugging purposes, and I forgot to delete it.
I removed the variable.

lanctot:
This has now been merged into the master branch in this commit: https://github.com/deepmind/open_spiel/commit/ca8affc740dcfef1b9032d49d95a6e3c54c03662. Thanks!

Issue #934, opened 2022-09-21
Possible contribution: LOLA 


axelbr:
Hi, 
I am planning to contribute a JAX implementation of [Learning with Opponent-Learning Awareness" (Foerster et al.)](https://arxiv.org/abs/1709.04326) if no one currently works on it and if it is still appreciated (I noticed it was mentioned in your call for contributions list).

I just wanted to check before starting to implement it :slightly_smiling_face:

lanctot:
Nice!! Absolutely welcome, would love to see it. 

Nobody is currently working on it to the best of our knowledge.

lanctot:
@axelbr this is very cool!!

Yes, please do make a PR and we can have more discussion there. In the post, please also include these graphs as well. Thanks.. this will make a wonderful addition to OpenSpiel!

lanctot:
Closing this thread as the PR is mow submitted. Thanks!

Issue #843, opened 2022-05-23
Call for New Games


lanctot:
Hi all, 

I'm starting a thread to collect requests for new games and plans for upcoming games. We will keep this updated as new games get released, and as requests come in (please feel free to comment on this thread to add a request). 

Games we would like to add to OpenSpiel:

- Ant foraging on a grid
- Capture-the-Flag on a gridworld (or other team adversarial games)
- [Chinese Checkers](https://en.wikipedia.org/wiki/Chinese_checkers)
- [Chinese Chess](https://en.wikipedia.org/wiki/Xiangqi) (see https://github.com/deepmind/open_spiel/issues/802)
- [Chinese Dark Chess](https://en.wikipedia.org/wiki/Banqi)
- [Mahjong](https://en.wikipedia.org/wiki/Mahjong)
- Pursuit-evasion game played on a graph or grid
- [Shogi](https://en.wikipedia.org/wiki/Shogi)

Implementing one of these games is a great way to get involved with contributing to OpenSpiel. If you would like to volunteer to implement one, please contact us and let us know as we might already be working on them. 

Games that have been added since posting this thread:
- [2048](https://en.wikipedia.org/wiki/2048_(video_game)) (Thanks @Jazeem. PR now merged: https://github.com/deepmind/open_spiel/pull/894)
- Checkers (thanks @Jazeem! PR now merged: https://github.com/deepmind/open_spiel/pull/861)
- Euchre (thanks @jhtschultz! Added in https://github.com/deepmind/open_spiel/commit/a3dbb1b7483b3b02077a92bb3a84389907afb79b)
- Mancala (Kalah) (thanks @Jazeem! see https://github.com/deepmind/open_spiel/pull/841)
- [Nim](https://en.wikipedia.org/wiki/Nim) (thanks @acforvs! PR merged: https://github.com/deepmind/open_spiel/pull/868)
- Phantom Go (thanks @Syor! PR now merged: https://github.com/deepmind/open_spiel/pull/867)
- [Dou dizhu](https://en.wikipedia.org/wiki/Dou_dizhu): (Thanks @rezunli96! PR merged: https://github.com/deepmind/open_spiel/pull/969)
- [Mensch-Ärgere-Dich-Nicht](https://en.wikipedia.org/wiki/Mensch_%C3%A4rgere_Dich_nicht) (Thanks @thorsten-j ! PR merged: https://github.com/deepmind/open_spiel/pull/915)
[...]

PR #1046, opened 2023-03-24
Opponent Shaping in JAX

axelbr:
Hi there!
Following up on #934, I finished the experiments for the LOLA-DiCE implementation. I tried to reproduce the results from the [DiCE paper ](https://arxiv.org/abs/1802.05098) (Figure 5). The results look somewhat consistent with the paper and the DiCE implementation seems to be working. I am not sure about the original LOLA implementation, however, the results in the [original paper](https://arxiv.org/abs/1709.04326) also seem to deviate from the results reported in the DiCE paper.

Please have a look at the experiment report: [Report from Weights and Biases](https://api.wandb.ai/links/axelbr/i4zv2l5x).

Looking forward to your comments!


lanctot:
Very nice! Thanks @axelbr , this is a great addition!

It's the end of the week so we'll take a closer look next week.

lanctot:
I noticed one of your commits had removing unfinished coin game? Is the one we have not usable as the one from the LOLA-DICE paper? If not, I wonder if we can generalize it, or add the one you had started to the [call for new games](https://github.com/deepmind/open_spiel/issues/843) list.

lanctot:
Can you add the `opponent_shaping_jax_test.py` to the python/CMakeLists, here: https://github.com/deepmind/open_spiel/blob/f7a43409f35921b7486ef76387e0947520add04f/open_spiel/python/CMakeLists.txt#L254  so that the test is triggered on the Github Actions CI ?

lanctot (changes requested):
Also, can you add Opponent Shaping to the list of algorithms in docs/algorithms.md?

open_spiel/python/jax/opponent_shaping_jax_test.py
```diff
@@ -0,0 +1,144 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for open_spiel.python.jax.lola."""
+import typing
+from typing import Tuple
+
+import distrax
```
lanctot, line 18: Ah, I think we'll also need to add distrax to [this line](https://github.com/deepmind/open_spiel/blob/f7a43409f35921b7486ef76387e0947520add04f/open_spiel/scripts/python_extra_deps.sh#L27) for this to run on CI

lanctot (changes requested):

axelbr:
> It's the end of the week so we'll take a closer look next week.

Ok, looking forward to hear from you! Have a nice weekend. 

> I noticed one of your commits had removing unfinished coin game? Is the one we have not usable as the one from the LOLA-DICE paper? If not, I wonder if we can generalize it, or add the one you had started to the [call for new games](https://github.com/deepmind/open_spiel/issues/843) list.

Yes. I wanted to replicate the same results and therefore based this env on the one I found [here](https://github.com/alexis-jacq/LOLA_DiCE). I think that there were some subtly differences in the implementation of the game compared to the version they mentioned in the paper. Furthermore, I found that having an environment that supports vectorization greatly reduced the runtime of my experiments when compared to the default games. I also added a custom matrix games environment which is by default vectorized by using numpy's shape broadcasting.

open_spiel/python/jax/opponent_shaping_jax_test.py
axelbr: I just made the requested changes and pushed a commit.

lanctot:
Ok, great. Thanks... yeah we need to support vectorized environments. It's been requested a few times and been on the TODO list for some time now. This PR should be imported before the next release (probably early May) so we can announce it in the release notes, but maybe a vectorized enviornment is something we should work on for the release after that. Actually, in addition to the Call for New Games, we should have a thread to track these new things in case people are looking for a way to contribute. I think I'll post one in the coming weeks.

lanctot:
Nope, I mean cpplint :) Which supports Google's Python style. It needs to be Google's specific style, which is the one cpplint uses. :)

axelbr:
Hmm..
According to the [cpplint](https://pypi.org/project/cpplint/) package documentation, this tool is only for checking cpp files. It also complains when I try to use it on python files ("`Ignoring ... .py, not a valid file name (h, c, hh, cc, ...)`"). When I explicitly specify .py extensions, it will lint it but it gives me wrong warnings (e.g.: Use `||` instead of `or`).
Furthermore, Google products like [Tensorflow](https://github.com/tensorflow/tensorflow/blob/940d8afc743598870cb6cc3f403587773ae17052/.github/workflows/pylint-presubmit.yml#L47) and [JAX](https://github.com/google/jax/pull/1442) use pylint apparently.

If I am really wrong, could you please give me a hint on how to use cpplint for python files?

lanctot:
Sorry! You are indeed correct, I misunderstood a previous conversation.

It is indeed pylint with the Google pylintrc obtained here: https://google.github.io/styleguide/pyguide.html

Apologies for the confusion!

lanctot:
Hi @axelbr, there will be a delay as I am now on holiday for two weeks. I will reply again after April 15th.

open_spiel/python/jax/opponent_shaping.py
```diff
@@ -0,0 +1,984 @@
...
+      },
+      critic_params=deepcopy(train_state.critic_params),
+      critic_opt_states=deepcopy(train_state.critic_opt_states),
+    )
+    return train_state, {'loss': loss}
+
+  return update
+
+
+class OpponentShapingAgent(rl_agent.AbstractAgent):
```
lanctot, line 536: This is defining an OpenSpiel agent. Do you know if it runs in the general case, e.g. can two opponent shaping agents play tic-tac-toe in self-play?

I ask only because the examples seem to use the vectorized iterated matrix games, and I am curious if you tested on any sequential games.

axelbr: > Hi @axelbr, there will be a delay as I am now on holiday for two weeks. I will reply again after April 15th.

Ok, no problem. Have a nice holiday.



> This is defining an OpenSpiel agent. Do you know if it runs in the general case, e.g. can two opponent shaping agents play tic-tac-toe in self-play?

Yes. In fact, I use the regular repeated matrix games in the unit tests.

axelbr: I did not try it on other games than repeated matrix games though.

lanctot: It's ok. Just please document in the header of the opponent_shaping.py (and as a note inalgorithms.md) that it currently only works for simultaneous move games

axelbr: Ok, I added the info to both files.

lanctot:
Looks like there is a merge conflict now with algorithms.md, can you pull the most recent changes from master and push the merge commit?

lanctot:
> Ok, I linted and reformatted the files according to the suggested style guide.

Hi @axelbr, I still see over ~30 linting errors upon import. Did you use pylint with the Google pylintrc found here? https://google.github.io/styleguide/pyguide.html

axelbr:
Somehow I did not get linter hints from my IDE for the `iterated_matrix_game.py` file. All the issues were due to this file. I fixed it now. 
I also updated the `requirements.txt` file for the example, because JAX has changed the installation procedure.

lanctot:
> Somehow I did not get linter hints from my IDE for the `iterated_matrix_game.py` file. All the issues were due to this file. I fixed it now. I also updated the `requirements.txt` file for the example, because JAX has changed the installation procedure.

I am still getting a lot of linter errors. Are you sure you used the Google pylintrc?

axelbr:
Yes. Here is my workflow:
1. Download [pylintrc](https://google.github.io/styleguide/pylintrc).
2. Running those commands:
`pylint --rcfile pylintrc open_spiel/python/jax/opponent_shaping.py`
`pylint --rcfile pylintrc open_spiel/python/jax/opponent_shaping_jax_test.py`
`pylint --rcfile pylintrc open_spiel/python/environments/iterated_matrix_game.py`
`pylint --rcfile pylintrc open_spiel/python/examples/opponent_shaping/lola_iterated_matrix_games_jax.py`

Each of the checks passes with a score 10/10 without any issues.

lanctot:
Strange, there were still many errors internally. I fixed them all manually.. you can see all the changes in the commit when this is merged on Monday. I am surprised to find that the Google style pylintrc differs from our internal style requirements. I'll look into that.

axelbr:
> Strange, there were still many errors internally. I fixed them all manually.. you can see all the changes in the commit when this is merged on Monday. I am surprised to find that the Google style pylintrc differs from our internal style requirements. I'll look into that.

Ok, I am sorry for the trouble you had with my files. Would it be an option to add the pylintrc file you are using to the repository? 
Anyways, thanks for adapting the files and merging :slightly_smiling_face:

lanctot:
No worries.. I _was_ 98% sure that the Google pylintrc matched the one we use internally... since.. well, it's Google's :) 

I'll look into it and adjust accordingly. Thanks again for the contribution!

PR #1054, opened 2023-04-17
InfostateTree python bind.

maichmueller:
This PR would bring bindings for the infostate tree class to the python module pyspiel. 

The following classes would be bound by this PR:

InfostateTree
InfostateTreeNode
InfostateNodeType (enum)

TreeplexVector< double >
LeafVector< double >
DecisionVector< double>

TreeplexVector< py::object >
LeafVector< py::object >
DecisionVector< py::object >

SequenceId
DecisionId
LeafId

Range< SequenceId >
Range< DecisionId >
Range < LeafId >

kUndefinedDecisionId (constant)
kUndefinedLeafId (constant)
kUndefinedSequenceId (constant)
kDummyRootNodeInfostate (constant)
kFillerInfostate (constant)


maichmueller:
I attempted the binding, but might be missing some details. For one, I am unable to test my bound classes in a local installation in developer mode, since I am unable to import the bindings from pyspiel. I would need help in understanding what part of the process I might have left out.

Once I can verify the bindings exist in the python module I will add to the python test.

I have also been unsure with how to bind the InfostateNode class correctly, since it doesn't have a public constructor and is constructed only by the tree. Since the tree exposes the nodes to the client, I exported the node class as well.  I opted to go for a std unique ptr mock as smart holder which does not manage the lifetime but acts like one (yet remains being copyable).

lanctot:
Going through some very old PRs. Not sure why we never imported this. @maichmueller can you merger changes from master or resubmit?

maichmueller:
Hi @lanctot, I just rebased the branch back onto current master and updated the python test file location to work the new cmake style for tests. 

Note that I also reworked the previous holder semantics for the node owned by the InfostateTree class. Previously, a user would have had to keep a python handle on the tree alive somewhere so that nodes dont silently disappear. Now, the nodes use aliasing shared ptrs to co-own their tree. This should ensure that the tree stays alive as long as any node is still held onto.

lanctot:
Ok #1074 has been merged into master, can you please pull change and submit a merge to this branch?

maichmueller:
@lanctot, rebased and exposed the new args in infostate tree. Also made the optional args (max move limit, store_all_world_states) kw-args only, since a positional swap error already came up in my own tests. Should avoid some users getting bit by this.

PR #1061, opened 2023-04-29
pybind11 for win32

Chersophyte:
UCI bot is not supported in Windows. But the python bindings are still trying to include it.


lanctot:
Thanks!

Issue #999, opened 2023-01-20
allow python subclasses of pyspiel.Policy


VitamintK:
It would be nice if we could subclass pyspiel.Policy, and then pass those policies to pyspiel functions which expect pyspiel.Policys.

So something like this:

```
import pyspiel
class PolicySubclass(pyspiel.Policy):
    def __init__(self, game):
        pyspiel.Policy.__init__(self)
   def get_state_policy_from_state(self, state, player):
       ...
   def get_state_policy_from_infostate_string(self, infostate_string):
      ...
```

Currently, trying to do so throws an error `__init__ must be called when overriding __init__` (https://github.com/pybind/pybind11/pull/2152), presumably because `Policy` has no constructor defined in `policy.h`. It looks like `Game` had the same issue, so a "trampoline class" was defined in `open_spiel/python/pybind11/python_games.h`. Would a similar solution should work for `Policy`?

lanctot:
From a quick read of the thread it seems like the trampoline class was needed because Game is not default constructible. 

I wonder if we can just add an empty constructor in this case. @elkhrt, do you know?

elkhrt:
This would be a great contribution! You'll still need a trampoline class, but the overall setup should be a lot simpler than with `Game`. There's an overview of the necessary setup here: https://pybind11.readthedocs.io/en/stable/advanced/classes.html

lanctot:
This is quite old, seems abandoned. @VitamintK please feel free to re-open if you ever get to this.

PR #1062, opened 2023-05-01
Add trampoline to Policy for python subclasses

maichmueller:
Solves #999


lanctot:
Thanks!

open_spiel/python/tests/policy_test.py
```diff
@@ -50,6 +50,56 @@
 ]
 
 
+class DerivedPolicyTest(absltest.TestCase):
+    def test_derive_from_policy(self):
+        try:
```
elkhrt, line 55: You don't need this try/except/fail block - the exception will cause the test to fail anyway.

open_spiel/python/tests/policy_test.py
```diff
@@ -50,6 +50,56 @@
...
+            # equivalent to:
+            # class DerivedPolicy(pyspiel.Policy):
```
elkhrt, line 57: This would be much clearer; is there any reason not to use it?

open_spiel/python/tests/policy_test.py
```diff
@@ -50,6 +50,56 @@
...
+                    'action_probabilities': lambda this, state: {0: 0.1, 1: 0.9},
+                    'get_state_policy': lambda this, istate: {10: 0.9, 11: 0.1}
+                }
+            )
+        except KeyboardInterrupt:
+            # we don't silently ignore keyboard interrupts
+            raise
+        except Exception as e:
+            # deriving from policy failed. Report the exception.
+            self.fail(f"Exception raised: {e}")
```
elkhrt, line 77: Can delete this

open_spiel/python/tests/policy_test.py
```diff
@@ -50,6 +50,56 @@
...
+        )
+        self.assertEqual(
+            policy.get_state_policy("some infostate"),
+            {10: 0.9, 11: 0.1}
+        )
+        try:
+            policy.serialize()
+        except RuntimeError as e:
+            # we expect this to fail as we didn't implement serialize() in the derived class.
+            pass
```
elkhrt, line 99: `with self.assertRaises(RunTimeError):`

open_spiel/python/pybind11/python_policy.h
```diff
@@ -0,0 +1,49 @@
+
+#ifndef OPEN_SPIEL_PYTHON_POLICY_H
+#define OPEN_SPIEL_PYTHON_POLICY_H
+
+#ifndef SINGLE_ARG
+   #define SINGLE_ARG(...) __VA_ARGS__
+#endif
```
elkhrt, line 7: Can we confine this to .cc files only?

elkhrt (changes requested):
Looks good - thank you!

open_spiel/python/tests/policy_test.py
maichmueller: I wasn't sure whether `class` statements may be parsed by Python at an earlier stage than execution, so I just opted for `type` which I knew to be dynamic. Changed it to the clearer declaration.

open_spiel/python/tests/policy_test.py
maichmueller: I wasn't sure whether class statements may be parsed by Python at an earlier stage than execution, so I just opted for type which I knew to be dynamic. Changed it to the clearer declaration.

open_spiel/python/pybind11/python_policy.h
maichmueller: whoops, this was supposed to be in the .cc anyway

open_spiel/python/tests/policy_test.py
```diff
@@ -52,33 +52,17 @@
 
 class DerivedPolicyTest(absltest.TestCase):
```
elkhrt, line 53: It would also be good to add a test that the derived policy can be used from C++ code.

maichmueller:
One thing to take note of is that I changed
```
      .def("action_probabilities",
           (std::unordered_map<Action, double>(open_spiel::Policy::*)(
               const open_spiel::State&) const) &
               open_spiel::Policy::GetStatePolicyAsMap)
      .def("get_state_policy_as_map",
           (std::unordered_map<Action, double>(open_spiel::Policy::*)(
               const std::string&) const) &
               open_spiel::Policy::GetStatePolicyAsMap);
```

to
```
        .def(
            "action_probabilities",
            py::overload_cast< const State& >(&Policy::GetStatePolicyAsMap, py::const_),
            py::arg("state"),
            ...
         )
         .def(
            "action_probabilities",
            py::overload_cast< const std::string& >(&Policy::GetStatePolicyAsMap, py::const_),
            py::arg("info_state"),
            ...
         )
```

i.e. the function `get_state_policy_as_map` is gone on the function side and now just an overload of `action_probabilities`. I find this interface clearer, but understand that it may incur some interface breaking changes for some users and should be communicated if you don't want the change reverted.

open_spiel/python/tests/policy_test.py
maichmueller: Should be no problem. Is there a simple (c++) function or class in the codebase that accepts a Policy pointer and can be called from python?

Would this be enough?

```
def test_cpp_policy_from_py(...):

     class DerivedPolicy(pyspiel.Policy):
            ...

     policy_obj = DerivedPolicy()

     that_one_pybound_cpp_func(policy_obj, potential_other_args_needed)
```

lanctot:
Yes please revert that change to the python API `get_state_policy_as_action_map`. We are consistently choosing names that match the C++ API, and cosmetics is mot a good enough reason to make backward-incompatible change.

lanctot:
Oh wait, it was already inconsistent before? Nevermind. :) If this works, it is fine by me.

maichmueller:
Yes, it was inconsistent before so I chose to go with the first name. I can also align the python name of the overload to the c++ name, i.e. `get_state_policy_as_action_map`, to adhere to the naming pattern if you prefer that?

lanctot:
Nah I spoke to soon. What you chose is the better name, let's stick with that.

open_spiel/python/tests/policy_test.py
maichmueller: added the test with calls for each method. It revealed that I had to actually use PYBIND11_OVERRIDE_NAME to get the correct calls from c++ as well!

elkhrt (approved):
LGTM - thanks for the PR

maichmueller:
Done. @elkhrt Please have another look at those changes to verify if this is all correct .

elkhrt (approved):

PR #1067, opened 2023-05-10
Add back cvxopt to extra deps now that Python 3.11 wheels are available

lanctot:
Ref: https://github.com/cvxopt/cvxopt/issues/228. 

- Upgrade cvxopt to version 1.3.1
- cvxopt 1.3.1 now supports Python 3.11
- Re-enable tests that depended on cvxopt

### DIFF
--- a/open_spiel/python/CMakeLists.txt
+++ b/open_spiel/python/CMakeLists.txt
@@ -317,26 +317,21 @@ endif()
 if (OPEN_SPIEL_ENABLE_PYTHON_MISC)
   set(PYTHON_TESTS ${PYTHON_TESTS}
       algorithms/adidas_test.py
-      # Temporarily disabled due to dependency on cvxopt.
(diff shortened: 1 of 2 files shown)


lanctot:
Hello @findmyway, 

I am getting an odd error in Julia CxxWrap compilation seemingly unrelated to this change: https://github.com/deepmind/open_spiel/actions/runs/4941971630/jobs/8843020459

I tried rerunning the job twice and it still produces this same error.

Any ideas? Should I increase the version number of CxxWrap?

findmyway:
Julia@1.9 just got released yesterday. I guess CxxWrap is not updated corresponding yet.

As a temporary fix, we may need to set Julia version to `1.8` in the github action explicitly for now.

```
- uses: julia-actions/setup-julia@v1
  with:
    version: 1.8
```

PR #1071, opened 2023-05-19
Pre-commit setup for the repo

maichmueller:
This PR addresses what has been mentioned in #1070.

What this PR includes:
- pre-commit configuration for **_formatters_**:
   - **clang-forma**t: 
   c++ project files formatting
   - **pyink**: 
   google's python formatter that is close to black, but allows 2-indent formatting (oh the humanity...)
   - **cmake-format**:
  formatting style for cmake files
   - **isort**: 
  sort python imports alphabetically (can use profile=black, so I hope this profile works with pyink too)
   - **black**: 
  the python formatter (commented out but added for when google policy on 2-indent changes)
- **pylint**:
using google's official configuration. Since this config enforces 2-indent, **pyink** is needed.
- common pre-commit **_checks_**:
   - id: check-added-large-files
   - id: check-case-conflict
   - id: check-docstring-first (currently uncommented as some files would need manual changing)
   - id: check-merge-conflict
   - id: check-symlinks
   - id: check-toml
   - id: check-yaml
   - id: debug-statements
   - id: end-of-file-fixer
   - id: mixed-line-ending
   - id: requirements-txt-fixer
   - id: trailing-whitespace
- common **_python checks_**:
   - python-check-blanket-noqa
   - id: python-check-blanket-type-ignore
   - id: python-no-log-warn (uncommented because clashes in multiple places where logging.warn has been used)
   - id: python-no-eval
   - id: python-use-type-annotations 

As a side effect of the pre-commit checks I have moved the setup definition into a `pyproject.toml`, since it allows to also write the configurations for some of these tools. In consequence:
- _`pyproject.toml` now holds the metadata of the project (version, authors, license, etc.)_
- _`setup.py` is only the extension builder_

The dev-tools needed to run these pre-commits have also been added to the the `pyproject.toml` as optional dependencies of the dev group and could then be installed via:
```
pip install 'open_spiel[dev]'
```
or locally 
```
pip install '.[dev]'
```

If this PR is accepted a remark in the docs should be made for developers who wish to contribute the upstream to run `pre-commit install` on their repo in order to activate the hooks.

PR #1072, opened 2023-05-19
Add pre-commit verification to CI

maichmueller:
Builds on #1071.

Adds a CI step that verifies all pre-commit checks pass, otherwise fails the build step and exits. Not built on #1071 yet to keep diff small and the changes easily in sight. Will need #1071 merged into it to become operational.

Currently set as a condition for the build job to run, but can be set as a side job.

Merging this PR will mean that the entire codebase needs to apply the checks and formatting mandated by the pre-commit setup and as such requires one behemoth PR to apply these changes, otherwise all CI steps will constantly fail.

**_I do not recommend merging this until all open PRs in consideration for the project have been merged or closed so that the entire repo can see all of the pre-commit hooks applied to it._**

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,3 +1,5 @@
+
+
 name: build_and_test
 
 on:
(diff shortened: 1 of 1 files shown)


lanctot:
Hi @maichmueller, I think this is great, and in another world where I had enough time to properly look into it and understand how it all works, I would fully embrace the move. I realize it would probably save us time long term but I don't even have the time to spend on the initial cost of learning how it all works.

I also think this is a bit heavy for contributors and would rather keep doing what we have been and keep it easy to contribute.

PR #1078, opened 2023-05-27
Provide complete paths in `developer_guide.md` to avoid confusion

DavidToneian:
Some file and directory names appear in multiple positions in the source and build trees. Stating their full paths should help avoid confusion.


lanctot:
Thanks!

PR #1082, opened 2023-05-30
Refactor using string_view and fix undefined behaviour

maichmueller:
This PR achieves the following things:

### Refactor: `const std::string&` --> `std::string_view` wherever possible and sensible. 

This was done for performance reasons: `const std::string&` automatically converts `const char*` and `std::string` r-values resulting in unnecessary heap allocations. A `string_view` avoids any allocations at the risk of dangling references which is not a problem if common rules are followed: generally only use in function parameters, never return `string_view`, never store `string_view`.
This point alone touches many files of the repo and creates a large diff (sorry).

### Feature: Allow Heterogenous Lookup (HL) in maps using `std::string` as key.

One of the features added by this PR is HL: looking up the value of a st::map/unordered_map associated with a key of type `std::string` via key-simile types `const char*` and `std::string_view`. HL is crucial in ensuring the performance of e.g. `map.find(str)` lookups to not be dominated by the forced conversion to `std::string` (heap allocations). It works by supplying a custom Comparator (ordered maps) or custom Hash and Equality (unordered maps) types for the map templates which allow to compare/hash/eq-check objects within the string family (currently defined as `string, string_view, const char*`) and forego the need to convert every lookup to `std::string`. Two points of interest:
- HL requires a little template metaprogramming machinery to filter out unaccepted types (SFINAE and static asserts). This  is provided in the new files `utils/heterogenous_lookup.h` and `utils/type_traits.h`.
- C++17 does not support HL. Abseil containers however do and since this repo relies on abseil machinery in many parts I have gone ahead and used the drop-in container replacements of abseil for the std counterparts. This may lead to changes in the expected performance of some core parts in the repo (whether for the better or worse I dare not say, abseil claims their containers are better).

### Bugfix: Fix undefined behaviour in `GameParameter`

So far, the class `GameParameter` has held a member of type `std::map< std::string, GameParameter>` which represents the usage of an incomplete type as the value type of a map template. According to my understanding, incomplete value types are not defined behaviour in this context in c++17 (and AFAIK still unchanged in later standards). Only smart pointers and some specific containers are allowed to hold incomplete value types (e.g. `vector`/`list`). Therefore, I refactored the member as a mapping of `std::shared_ptr<GameParameter>` and moved it to an abseil container to also enable HL. This has consequences to nearly the entire repo. For example, the tests construct parameter objects via `GameParameter(args...)` which up until now provided values. These calls are now replaced everywhere by `MakeGameParameter(args...)` which automatically returns a shared ptr of type `GameParameter`. Other places now needed to deref these pointers (`->` instead `.`), comparison of two objects of type `GameParameters` (heed the plural, it is meant) is no longer meaningfully provided by the default comparison of keys and values in a map due to the values being shared pointers; a custom equality function is needed. Many other small changes were necessary. While doing so, I also refactored some of the functions around `GameParameter`. When in Rome...

### Minor refactors of unnecessary copies 

E.g. pass-by-value instances of `std::shared_ptr` and other points in which copies were created without need.

Overall this PR touches a vast amount of the c++ codebase, so apologies for the massive diff.


lanctot:
Hi @maichmueller,

I appreciate this, but since you have already made a few large PRs, could I kindly ask you to touch base with us before you do this? It could save a lot of wasted effort.

There is a reason string_view is not more widely used in OpenSpiel: IMO, it's overkill. It's mostly solving a non-problem, and for 99% of use cases you add nonstandard syntax for a few nanoseconds benefit. Also, I have seen it lead to several nontrivial bugs that would not have happened with string references.

I think this is a clash with the first principle of the design philosophy of the project (keep it simple). IMO string_view complicates things for very miniscule benefit and I would rather keep things simple (and only use it in cases where it is really beneficial).

Anyway. Like I said, I appreciate the enthusiasm regarding PRs but please try to reach out before spending effort on large changes like this. Thanks.

lanctot:
For similar reasons, we usually discourage anything that requires template meta programming (which IMO leads to even bigger potential problems and deeper knowledge of C++).

maichmueller:
Hi @lanctot ,

noted, I will open an issue next time if I consider major work again. However, a PR is communication too IMO and I would expect to discuss and receive feedback there as well. At the least, I am using my work in my own fork so it wouldn't be wasted effort.

Regarding string views, I understand your concerns. However, as function parameters they are harmless and signal intent well I find. Users who need to modify the string can simply make an actual string out of it so that errors like the one in your example wouldn't be possible. Understanding the difference between a string view and a string is quite similar to understanding a raw pointer's relation to smart ptrs in the end (sans implicit conversion).

Anyhow, the commits are split into two parts and I could cut out the fix for the UB if you don't want the view refactoring. If you also don't want the UB fix (or would like to discuss it first in an issue) then feel free to just close the PR.

lanctot:
> Anyhow, the commits are split into two parts and I could cut out the fix for the UB if you don't want the view refactoring. If you also don't want the UB fix (or would like to discuss it first in an issue) then feel free to just close the PR.

More than happy to have a fix for undefined behavior, thanks! :) 

(Though I may have to fullt understand it first, at first glance I see no problem)

lanctot:
Hold on, don't you need the string_view for the heterogenous lookup? 

Also: I'm a bit confused:
> According to my understanding, incomplete value types are not defined behaviour in this context in c++17 (and AFAIK still unchanged in later standards)

I'm not sure what problem is. Wouldn't it just not compile? This map has been used everywhere from the start.. we've never had a problem. Is it easy to construct a concrete example where we'd run into a problem with a `map<std::string, GameParameter>`?

> Only smart pointers and some specific containers are allowed to hold incomplete value types (e.g. vector/list). 
Again, wouldn't it just not compile if it wasn't "allowed" ? Do you have a reference for this. I'd like to understand the problem better. Why is having a map of a templated type problematic? I can imagine the fix is rather involved if it means introducing smart pointers...?

lanctot:
Going to summon @elkhrt to help me out :)

maichmueller:
> Hold on, don't you need the string_view for the heterogenous lookup?

You can have HL for any types you define it for (in this pr: string, string_view, const char*), so removing string_view would not necessarily remove HL as long as you have function overloads. But if the ` const string&` signature is preferred, then taking on HL is pointless and would not need to be needed, yes.    

> Also: I'm a bit confused:
> 
> > According to my understanding, incomplete value types are not defined behaviour in this context in c++17 (and AFAIK still unchanged in later standards)
> 
> I'm not sure what problem is. Wouldn't it just not compile? This map has been used everywhere from the start.. we've never had a problem. Is it easy to construct a concrete example where we'd run into a problem with a `map<std::string, GameParameter>`?

As with all things UB, compilers are free to do whatever they see fit, including making the situation work. Major compiler vendors like clang and GCC do just that it seems. For example, MSVC is not as forgiving AFAIK. Compilers are free to drop supporting this in future versions though which would leave a tricky UB bug to triage. Possible ways of dealing with this is:
- ignore and bet on the compilers to keep playing along
- switch to a container implementation that explicitly allow this (see later in the message)
- Fix by changing the `mapped_type` to a container which allows incomplete types by the standard (`shared/unique_ptr`). 

> > Only smart pointers and some specific containers are allowed to hold incomplete value types (e.g. vector/list).
> 
> Again, wouldn't it just not compile if it wasn't "allowed" ? Do you have a reference for this. I'd like to understand the problem better. Why is having a map of a templated type problematic?

UB is free real estate for compilers, so it may compile just fine.
As for references see e.g. https://stackoverflow.com/a/38060437 ('Container Improvements' and https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4510.html) or https://stackoverflow.com/a/41913654.

Boost equivalent containers do allow incomplete types in line with boost's general aim to provide more functionality than the std library. 
Abseil equivalent containers actively disallow them (static_assert error) which matches their attempt at being standard compliant with better implementations.

>  I can imagine the fix is rather involved if it means introducing smart pointers...?

What I attempted to fix it is described in the initial post. See `GameParameters.h`  as an example of the changes. I have since then noticed on my own branch that the type caster for the pybind11 GameParameter binding needs changing still which isn't in this PR yet.

elkhrt:
Thanks for the contributions & the enthusiasm for improving OpenSpiel!

1 - Now that std::string_view has been in the C++ standard for 5 years, I think it's arguable that `void foo(std::string_view s)` is more idiomatic and more readable than `void foo(const std::string& s)`. For most naive users, either one is just a magic incantation for passing a string from one function to another. Given the usability advantages, I think it's worth changing. I'll attempt to convince Marc in a side channel.

2 - I think you're technically correct about GameParameters, but the fix seems like overkill. Couldn't we just move the definition of GameParameters after the declaration of GameParameter? That means we can't use the type alias inside GameParameter, but has no wider implications.
https://www.youtube.com/watch?v=0ZEuWJ4muYc

3 - The HL sounds it's probably a necessary part of the string_view switch. If the template metaprogramming can be hidden from ordinary users, and this just looks like a set / map at the point of use, then I think this is probably okay. But obviously without string_view there's limited (but not zero!) point.

maichmueller:
Hi @elkhrt 

Regarding 1:
One argument to add to the debate of whether to view or not to view this is that the decision against would probably have to be firm, as the cost of future refactoring only grows with time and the evolution of the library. Tackling all the parts in the repo was a fair bit of work in the end.

Regarding 2:
I agree that the fix is sort of overkill. It appeared to be the least divergent solution at the time since it merely needed to replace values with pointers.
In general, I wasn't quite able to understand the intention behind the design of GameParameter (a tree data structure of further parameters + a union type itself which isn't using a union/variant though). Definitely would have been wiser to discuss this fix in an issue first.
Could you elaborate on how we could move the definition of GameParameters after the definition of GameParameter? At the time of GameParameter's definition, the member types need to be complete or reference types after all (struct's memory layout is otherwise indeterminable)

lanctot:
> 1 - Now that std::string_view has been in the C++ standard for 5 years, I think it's arguable that `void foo(std::string_view s)` is more idiomatic and more readable than `void foo(const std::string& s)`. For most naive users, either one is just a magic incantation for passing a string from one function to another. Given the usability advantages, I think it's worth changing. I'll attempt to convince Marc in a side channel.

I don't think we should change something just because it has been in the standard for 5 years. If it's arguably the new community idiom, then someone will need to convince me. I have searched several times on this topic (including most recently when this PR came up) and all I have seen are people asking the same questions as me and essentially everyone shrugging. 

I don't see how in OpenSpiel we will benefit from performance enhancements (without buying into it quite heavily-- and the benefits are not clear to me). I have only seen hard to fix bugs arise from it so far.

What are the usability advantages compared to a string reference? Will they improve OpenSpiel?

I am open to being convinced, but I am still not seeing the case for it in OpenSpiel.

lanctot:
I have read a few blogs and I am starting too see the benefits. 

But let me ask: the thing I need most is to be convinced that the community is indeed moving away from const string ref, and I am not seeing any evidence of that.

Can you show me some? Has there been any videos on people trying to make the case to move away from const tring ref? To me, it seems quite up in the air still, after 5 years. I tend to be slow on these because frivolous changes for performance don't make much sense if 99% of use cases do not benefit from it and add chances of subtle bugs. The correctness and clarity of the code is so much more important IMO.

elkhrt:
> Regarding 2:
> I agree that the fix is sort of overkill. It appeared to be the least divergent solution at the time since it merely needed to replace values with pointers.
> In general, I wasn't quite able to understand the intention behind the design of GameParameter (a tree data structure of further parameters + a union type itself which isn't using a union/variant though). Definitely would have been wiser to discuss this fix in an issue first.
> Could you elaborate on how we could move the definition of GameParameters after the definition of GameParameter? At the time of GameParameter's definition, the member types need to be complete or reference types after all (struct's memory layout is otherwise indeterminable)

Delete the line ```class GameParameter;```

Move the block

 ```
using GameParameters = std::map<std::string, GameParameter>;
std::string GameParametersToString(const GameParameters& game_params);
GameParameter GameParameterFromString(const std::string& str);
GameParameters GameParametersFromString(const std::string& game_string);
```

to be after the end of the definition of the `GameParameter` class.

Inside the definition of `GameParameter`, either:
a - replace `GameParameters` with `std::map<std::string, GameParameter>` throughout
b - add `using GameParameters = std::map<std::string, GameParameter>` inside the class, and `using GameParameters = GameParameter::GameParameters;` outside it (replacing the other previous `using`)

**Comment by elkhrt** [comment]
> Regarding 2:
> I agree that the fix is sort of overkill. It appeared to be the least divergent solution at the time since it merely needed to replace values with pointers.
> In general, I wasn't quite able to understand the intention behind the design of GameParameter (a tree data structure of further parameters + a union type itself which isn't using a union/variant though). Definitely would have been wiser to discuss this fix in an issue first.
> Could you elaborate on how we could move the definition of GameParameters after the definition of GameParameter? At the time of GameParameter's definition, the member types need to be complete or reference types after all (struct's memory layout is otherwise indeterminable)

The fix I had in mind doesn't work, but how about this? Replace the member variable:
```GameParameters game_value_ = {};``` with ```std::shared_ptr<std::map<std::string, GameParameter>>``` and then make a few local changes in that class to preserve the existing interface as far as possible.

Also move the block:
```
using GameParameters = std::map<std::string, GameParameter>;
std::string GameParametersToString(const GameParameters& game_params);
GameParameter GameParameterFromString(const std::string& str);
GameParameters GameParametersFromString(const std::string& game_string);
```
To after the class definition.
This will only have any impact on call sites using the `GameParameter::GameParameter(GameParameters)` constructor which is pretty rare.

maichmueller:
@lanctot :

> But let me ask: the thing I need most is to be convinced that the community is indeed moving away from const string ref, and I am not seeing any evidence of that.

I am not even sure how to provide such evidence. What community do you mean? 
AI researchers? I doubt they would, but I assume they are most at home with other languages like Python and may tend to use C++ pythonically anyway.
High-performance library devs? I expect them to inspect `const string&` usage more carefully, yes.

[P](https://www.learncpp.com/cpp-tutorial/introduction-to-stdstring_view/)-[l](https://stackoverflow.com/a/40129198)-[e](https://www.youtube.com/watch?v=fj_CF8xK760)-[n](https://www.modernescpp.com/index.php/c-17-avoid-copying-with-std-string-view)-[t](https://quuxplusone.github.io/blog/2018/03/27/string-view-is-a-borrow-type/)-[y](https://www.nextptr.com/tutorial/ta1217154594/cplusplus-stdstring_view-for-better-performance-an-example-use-case) of sources advice using string_view, albeit with caveats of course. And that is also the point, string_view is not meant to simplify your life primarily. It is supposed to add to the string usage performance:
1. no automatic conversion to std::string
2. no double indirection to the underlying chars
3. substring operations are now constant complexity, not linear anymore
4. (not applicable to open_spiel) allows interaction with different string implementations.

Point 1's relevance is arguably dampened by small string optimization (23 chars on clang AFAIK, so potentially even most of all 'const char' usages in open_spiel, less with other compilers). 
Point 2 asserts that you forego many cache misses which double indirections are very prone to.

But these points are **all** _performance_ based, so when you ask

> What are the usability advantages compared to a string reference? Will they improve OpenSpiel?

the answer is 0. Everything that implicitly converts to `string_view` in this repo will do the same to a `const string &`.

Now, I didn't time the performance gain of slapping string_view on function parameters in all honesty. Maybe it doesn't change a thing. This PR is merely a reflection of what **I** consider best practice.

I sympathize with your feeling on string_view and so do [others](https://groups.google.com/a/isocpp.org/g/std-discussion/c/Gj5gt5E-po8). Personally, I find it safe to use in the ways I described above. However, I am not here to convince you of something that you seem all but decided on.

@elkhrt

Correct me if I am wrong, but I _think_ `std::shared_ptr<std::map<std::string, GameParameter>>` is, technically, still UB due to the same map problem. At the point of this definition, the map specialisation is already complete. So, even if shared_ptr accepts incomplete types, the type `std::map<std::string, GameParameter>` is complete and does not.

lanctot:
> But these points are **all** _performance_ based, so when you ask
> 
> > What are the usability advantages compared to a string reference? Will they improve OpenSpiel?
> 
> the answer is 0.

Ed specifically mentioned usability advantages, so I guess that one was mostly directed at him.

> Now, I didn't time the performance gain of slapping string_view on function parameters in all honesty. Maybe it doesn't change a thing. This PR is merely a reflection of what **I** consider best practice.
> 
> I sympathize with your feeling on string_view and so do [others](https://groups.google.com/a/isocpp.org/g/std-discussion/c/Gj5gt5E-po8). Personally, I find it safe to use in the ways I described above. However, I am not here to convince you of something that you seem all but decided on.

I've got my reasons to be hesitant that I've made clear, but mostly trying to understand the motivation + justification for such a large change (more so now that Ed wants to convince me too), and weighing the pros and cons. I've had a hard stance before and managed to be convinced; in those cases, I often learn something. I appreciate the effort and patience, thanks.

Issue #1090, opened 2023-06-30
Documentation of game parameters, observations, actions


raphCode:
During my usage of openspiel, I encountered difficulties finding detailed documentation on game parameters, observations, and actions.
Specifically, there is no clear list of parameters accepted by each game, and I had to search through the C++ code to find them.
Furthermore, it is challenging to determine which observations are available to agents without guessing based on Tensor shapes or content. This makes it very hard to comprehend whether the agent can even learn something reasonable based on the available information.

A very specific case is the game `coop_box_pushing`, which is advertised as a perfect information game by the docs. In the default configuration however, the observation includes only a single field in front of the agent. One needs to discover the game may take an `fully_observable` parameter which actually makes the observation perfect information.

Would it be possible to include this information in the docs? Maybe it is already there, and I was not able to find it?
It would also be acceptable if the documentation pages link to the relevant C++ code with explaining comments.

The bottom-line is: I find it highly surprising that a python-faced project with an `readthedocs.io` page has very little usable documentation*, and expects users to examine various code examples and the C++ source to collect information.

*Part of my frustration is that I started my project before there was even a core API documentation available, requiring hours of reverse engineering the API in an `ipython` console.

lanctot:
Yeah, the documentation can always be improved. 

We have a video tutorial linked from the main page: that, and the API reference, are the best ways to understand the core API. Some other functions may require reading function headers in spiel.h.

Regarding observations, you didn't miss it. They are documented in the headers of the C++ source files. Unfortunately we often don't have the time to improve the docs due to other higher priorities, but could really use the community's help with this! 

I like the idea of linking to the section of the C++ header from the games page. Can you submit a PR?

raphCode:
> I like the idea of linking to the section of the C++ header from the games page. Can you submit a PR?

I will try to find some time. Currently I am busy with my thesis, so that could take a while.

How would this best be done without increasing the maintenance burden? Can we autogenerate docs from the code / using a helper script?

lanctot:
This would be wonderful and still welcome, but I'm closing the issue for now due to inactivity.

PR #1108, opened 2023-08-11
Add some links to the implementation

raphCode:
Like this? #1090

https://github.com/deepmind/open_spiel/pull/1103 just made this easier.

Not completed yet, just added three links by hand.

### DIFF
--- a/docs/games.md
+++ b/docs/games.md
@@ -1,5 +1,7 @@
 # Available games
 
+In general, see the implementation source code for details including game options, observations and actions.
+
 ![](_static/green_circ10.png "green circle"): thoroughly-tested. In many cases,
 we verified against known values and/or reproduced results from papers.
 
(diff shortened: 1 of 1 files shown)


lanctot:
Can you clarify the motivation?

The games are all in the same place, under games/ or python/games. So they are not too hard to find. We could just could have maybe 1-2 links to the parent folders at  the top than rather than have links for all the games?

raphCode:
Oh sorry. The motivation was basically this quote from you in another issue: https://github.com/deepmind/open_spiel/issues/1090#issuecomment-1615998452
> I like the idea of linking to the section of the C++ header from the games page. Can you submit a PR?

However, linking to just the parent folder is also a viable option now. (Also less docs maintenance if the code should ever move)

lanctot:
Closing due to inactivity. Please re-open if you continue working on this.

PR #1115, opened 2023-09-02
Remove unnecessary policy reload in Outcome-Sampling MCCFR

asaf92:
There's a redundant assignment to the `policy` variable.

I think `policy` needs to be re-assigned, but only after the regrets have been updated. This way the avg strategy is updated according to the new strategy, not the previous strategy.

Let me know if you think this is wrong

### DIFF
--- a/open_spiel/python/algorithms/outcome_sampling_mccfr.py
+++ b/open_spiel/python/algorithms/outcome_sampling_mccfr.py
@@ -118,11 +118,8 @@ class OutcomeSamplingSolver(mccfr.MCCFRSolverBase):
     for aidx in range(num_legal_actions):
       value_estimate += policy[aidx] * child_values[aidx]
 
+    # Update regrets and avg strategies
     if cur_player == update_player:
-      # Now the regret and avg strategy updates.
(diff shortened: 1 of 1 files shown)


asaf92:
> > Let me know if you think this is wrong
> 
> Ah yes, this is intentional. The idea being that that you should be computing the average for the strategies as they were before you modify them. The importance sampling term is also computed using these unchanged policies.
> 
> It's basically the analogue to vanilla CFR: it's intentional to include the first uniform strategy in the average.
> 
> Both probably work but I'd argue that what we have now is "more correct" if that makes sense.

OK but even if it's the case, `policy` does not need to be reassigned. It was already assigned and hasn't been modified, right?

lanctot:
You mean here? https://github.com/google-deepmind/open_spiel/blob/894710bbbe806e2738a05db0cb097d92c75ccb94/open_spiel/python/algorithms/outcome_sampling_mccfr.py#L123

Yes, I believe you are correct. 

We must have copied it from the C++ implementation where you do need to reassign because the object reference could change due to the map dat structure being modified.

I think indeed this line can be safely removed.

Issue #1117, opened 2023-09-10
Enhance Bridge State


ZiggerZZ:
Hello,

I would like to propose a bridge state features enhancement to include more tricks from the past play. With 13 tricks every `state.observation_tensor()` would represent the full history of the card playing phase.

Here's my proposal: we introduce two params, `num_tricks` that defaults to 2 (as of the current implementation) and `absolute_order` (default `true`), that we can use like that: `GAME = pyspiel.load_game("bridge(use_double_dummy_result=false, num_tricks=13,absolute_order=false)")`.

`absolute_order=true` would indicate that the tricks in the observation tensor would be stored in the absolute order: 1st trick, 2nd trick, ..., current trick, ..., zeros (for futures tricks).
`absolute_order=false` would indicate that the tricks in the observation tensor would be stored in the relative order: current trick, previous trick, ..., zeros (for futures tricks). I find this representation better for RL algorithms. 

Here's a working implementation of `num_tricks=13, absolute_order=true` (replace lines 340-360 in [bridge.cc](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/games/bridge/bridge.cc#L340)):

```c++
    // Indexing into history for recent tricks.
    int current_trick = num_cards_played_ / kNumPlayers;
    int this_trick_cards_played = num_cards_played_ % kNumPlayers;
    int this_trick_start = history_.size() - this_trick_cards_played;

    // All previous tricks
    for (int j = 0; j < current_trick; ++j) {
      int leader = tricks_[j].Leader();
      for (int i = 0; i < kNumPlayers; ++i) {
        int card = history_[this_trick_start - kNumPlayers * (current_trick - j) + i].action;
        int relative_player = (i + leader + kNumPlayers - player) % kNumPlayers;
        ptr[relative_player * kNumCards + card] = 1;
      }
      ptr += kNumPlayers * kNumCards;
    }

    // Current trick
    if (phase_ != Phase::kGameOver) {
      int leader = tricks_[current_trick].Leader();
      for (int i = 0; i < this_trick_cards_played; ++i) {
        int card = history_[this_trick_start + i].action;
        int relative_player = (i + leader + kNumPlayers - player) % kNumPlayers;
        ptr[relative_player * kNumCards + card] = 1;
      }
    }
    // move point for current and future tricks to have a fixed size tensor
    ptr += kNumPlayers * kNumCards * (kNumTricks - current_trick);
```

[...]

PR #1118, opened 2023-09-18
Bridge

ZiggerZZ:
Hello!

This is the implementation of #1117. It is _not_ ready to be merged yet - I need to change all `binvec`s in `bridge(use_double_dummy_result=false).txt` to reflect the change of the order of the tricks (from `[previous, current]` to `[current, previous]`). As suggested by @elkhrt, I didn't introduce `absolute_order` parameter.
If this approach is fine with you, I'll update the tests and the PR will be ready to be merged.


ZiggerZZ:
Fixed the tests. Since I swap the current and the previous tricks, I needed to swap those in the binary representations.
Here's the script that takes `bridge(use_double_dummy_result=false).txt` and modifies hex values for state observations only for the playing phase (hence condition `idx > 700`). I used this script to patch the tests.

```python
def change_order(hex_value):
    int_value = int(hex_value, 16)  # Convert the hex string to an integer
    binary_str = bin(int_value)[2:]  # Convert the integer to binary and remove the '0b' prefix
    length_required = 571
    binary_str_padded = binary_str.zfill(length_required)
    index_previous_start, index_previous_end = 129, 337
    index_current_start, index_current_end = 337, 545
    new_string_binary = binary_str_padded[:index_previous_start] + binary_str_padded[index_current_start:index_current_end] + binary_str_padded[index_previous_start: index_previous_end] + binary_str_padded[index_current_end:]
    int_value_from_binary = int(new_string_binary, 2)
    hex_str = hex(int_value_from_binary)
    return hex_str

file_path = "./bridge(use_double_dummy_result=false).txt"

with open(file_path, 'r') as file:
    lines = file.readlines()

for idx, line in enumerate(lines):
	# only apply to lines after bidding
    if "binvec" in line and idx > 700:
        hex_value = line.split(', ')[-1].split(')')[0]  # Extract the hex value
        new_hex_value = change_order(hex_value)
        lines[idx] = line.replace(hex_value, new_hex_value)  # Replace old hex value with the new one

with open(file_path, 'w') as file:
    file.writelines(lines)
```

open_spiel/games/bridge/bridge.cc
```diff
@@ -357,13 +350,42 @@ void BridgeState::WriteObservationTensor(Player player,
...
+        kNumBidLevels                           // What the contract is
+        + kNumDenominations                     // What trumps are
+        + kNumOtherCalls                        // Undoubled / doubled / redoubled
+        + kNumPlayers                           // Who declarer is
+        + kNumVulnerabilities                   // Vulnerability of the declaring side
+        + kNumCards                             // Our remaining cards
+        + kNumCards                             // Dummy's remaining cards
+        + num_tricks_ * kNumPlayers * kNumCards // Number of played tricks
+        + kNumTricks                            // Number of tricks we have won
+        + kNumTricks;                           // Number of tricks they have won
```
MattX, line 388: Might be a good idea to make this a function, since you do the same calculation in `GetObservationTensorSize`.

open_spiel/games/bridge/bridge.h
```diff
@@ -221,14 +208,31 @@ class BridgeGame : public Game {
...
+        + kNumOtherCalls                        // Undoubled / doubled / redoubled
+        + kNumPlayers                           // Who declarer is
+        + kNumVulnerabilities                   // Vulnerability of the declaring side
+        + kNumCards                             // Our remaining cards
+        + kNumCards                             // Dummy's remaining cards
+        + num_tricks * kNumPlayers * kNumCards  // Number of played tricks
+        + kNumTricks                            // Number of tricks we have won
+        + kNumTricks;                           // Number of tricks they have won
+    int kObservationTensorSize = kNumObservationTypes + std::max(kPlayTensorSize, kAuctionTensorSize);
+    return kObservationTensorSize;
```
MattX, line 231: nit: You can probably collapse this into a single line; I would also move the entire function inside `ObservationTensorShape` since it isn't used anywhere else.

ZiggerZZ:
Taken @MattX's review into account

lanctot:
Thanks for this and to @MattX for the review.

We should be able to pull this in soon. I'll just quickly run it by Ed early next week.

lanctot:
BTW @ZiggerZZ, you can always simply regenerate those playthroughs given the new implementation using this script: https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/scripts/regenerate_playthroughs.sh

ZiggerZZ:
Thanks @lanctot! I wanted to verify the correctness of the implementation by writing that logic of modifying the existing observation tensors.

PR #1124, opened 2023-10-15
Modularize the CMake build targets of each game

jthemphill:
This PR modularizes the dependencies for `amazons`. If you approve of my general direction, I'll extend this to all the games, which should cut the test folder's size from 600MB to 100MB and reduce test linkage time by 33%.

While trying to build the test for one of the games, I noticed the `Makefile` had me building *every* game:

```
jhemphill@CeilingComp:~/oss/open_spiel/build$ make -j$(nproc) amazons_test
[  0%] Built target absl_spinlock_wait
...
[ 52%] Built target absl_flags_parse
Consolidate compiler generated dependencies of target bots
Consolidate compiler generated dependencies of target utils
Consolidate compiler generated dependencies of target game_transforms
Consolidate compiler generated dependencies of target bridge_double_dummy_solver
[ 52%] Built target tests
Consolidate compiler generated dependencies of target algorithms
[ 52%] Built target bots
[ 54%] Built target utils
[ 54%] Building CXX object game_transforms/CMakeFiles/game_transforms.dir/start_at.cc.o
...
[ 58%] Building CXX object games/CMakeFiles/games.dir/battleship/battleship.cc.o
[ 62%] Building CXX object games/CMakeFiles/bridge_double_dummy_solver.dir/bridge/double_dummy_solver/src/TimerList.cpp.o
[ 62%] Building CXX object algorithms/CMakeFiles/algorithms.dir/evaluate_bots.cc.o
[ 62%] Building CXX object games/CMakeFiles/games.dir/blackjack/blackjack.cc.o
...
```

In addition to building everything, it seems to have linked everything in as well, resulting in a 7.4MB test:

```
jhemphill@CeilingComp:~/oss/open_spiel/build$ ls -lh games/amazons_test 
-rwxr-xr-x 1 jhemphill jhemphill 7.4M Oct 14 17:23 games/amazons_test
```

Since every test links every game in, every test is at least 7.4MB:

```
jhemphill@CeilingComp:~/oss/open_spiel/build$ ls -lh games/
total 608M
-rwxr-xr-x  1 jhemphill jhemphill 7.4M Oct 14 12:36 2048_test
drwxr-xr-x 86 jhemphill jhemphill 4.0K Oct 14 17:15 CMakeFiles
-rw-r--r--  1 jhemphill jhemphill  21K Oct 14 17:15 CTestTestfile.cmake
-rw-r--r--  1 jhemphill jhemphill 312K Oct 14 17:15 Makefile
[...]

PR #1126, opened 2023-10-16
Improve test compilation speed and space usage by linking tests against `libopen_spiel.so`

jthemphill:
This is an alternative approach to #1124. Compared to the status quo, it cuts over 500MB off of the compiled build artifacts, cuts three minutes off the single-threaded full build time, and removes a lot of thinking about which libraries you need to link your game or test against.

Before:

```
$ make clean ; time make -j8
real    5m25.136s
user    34m56.492s
sys     2m8.369s

$ du -sh games
629M    games
```

After:

```
$ ninja clean ; time ninja -j8
[757/757] Linking CXX shared module python/pyspiel.so

real    4m35.704s
user    31m10.678s
sys     1m51.133s

$ du -sh
111M    .

$ du -sh games
34M     games
```

Previously, we conditionally generated `libopen_spiel.so`. Now, we always generate it. And when creating a new executable, you'll want to use `add_open_spiel_executable()` instead of `add_executable(...${OPEN_SPIEL_OBJECTS})`.

### DIFF
--- a/open_spiel/games/CMakeLists.txt
+++ b/open_spiel/games/CMakeLists.txt
@@ -272,355 +272,171 @@ add_library(bridge_double_dummy_solver OBJECT
 target_include_directories (bridge_double_dummy_solver PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 target_compile_definitions(bridge_double_dummy_solver PUBLIC DDS_NO_STATIC_INIT)
 
-add_executable(2048_test twenty_forty_eight/2048_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(2048_test 2048_test)
-
-add_executable(amazons_test amazons/amazons_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(amazons_test amazons_test)
-
-add_executable(backgammon_test backgammon/backgammon_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(backgammon_test backgammon_test)
-
-add_executable(bargaining_instance_generator bargaining/bargaining_instance_generator.cc
-               ${OPEN_SPIEL_OBJECTS})
-add_executable(bargaining_test bargaining/bargaining_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(bargaining_test bargaining_test)
-
-add_executable(battleship_test battleship/battleship_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(battleship_test battleship_test)
-
-add_executable(blackjack_test blackjack/blackjack_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(blackjack_test blackjack_test)
-
-add_executable(blotto_test blotto/blotto_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(blotto_test blotto_test)
-
-add_executable(breakthrough_test breakthrough/breakthrough_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(breakthrough_test breakthrough_test)
-
-add_executable(bridge_test bridge/bridge_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(bridge_test bridge_test)
-
-add_executable(catch_test catch/catch_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(catch_test catch_test)
-
-add_executable(checkers_test checkers/checkers_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(checkers_test checkers_test)
-
-add_executable(chess_test chess/chess_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(chess_test chess_test)
-
-add_executable(cliff_walking_test cliff_walking/cliff_walking_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(cliff_walking_test cliff_walking_test)
-
-add_executable(clobber_test clobber/clobber_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(clobber_test clobber_test)
-
-add_executable(coin_game_test coin_game/coin_game_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>
-               $<TARGET_OBJECTS:algorithms>)
-add_test(coin_game_test coin_game_test)
-
-add_executable(colored_trails_board_generator
-               colored_trails/colored_trails_board_generator.cc
-               ${OPEN_SPIEL_OBJECTS} $<TARGET_OBJECTS:tests>)
-
-add_executable(colored_trails_test colored_trails/colored_trails_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(colored_trails_test colored_trails_test)
-
-add_executable(connect_four_test connect_four/connect_four_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>
-               $<TARGET_OBJECTS:algorithms>)
-add_test(connect_four_test connect_four_test)
-
-add_executable(coop_box_pushing_test coop_box_pushing/coop_box_pushing_test.cc
-               ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(coop_box_pushing_test coop_box_pushing_test)
-
-add_executable(coordinated_mp_test coordinated_mp/coordinated_mp_test.cc ${OPEN_SPIEL_OBJECTS}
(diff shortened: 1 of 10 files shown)


jthemphill:
It turns out a number of tests (`battleship_test`, `dark_chess_test `, `universal_poker_test`, etc...) depend on running a game's static initializers so that the game is registered in the OpenSpiel engine. It also turns out that linking against a `.a` static library will cause none of these initializers to run😑 Changing `libgames.a` to a shared library, `libgames.so`, fixes the problem and further reduces the size of each test to 900kB each. The tests now depend on `libgames.so` existing at runtime, but... they're tests, they don't have to be portable.

lanctot:
The failed wheels test is our fault from an update yesterday, we'll fix that. Apologies.

jthemphill:
@lanctot: If you agree with the direction I've taken here, I think this is ready for another CI run. If you disagree and don't want a shared library here, please let me know!

lanctot:
> @lanctot: If you agree with the direction I've taken here, I think this is ready for another CI run. If you disagree and don't want a shared library here, please let me know!

Hey @jthemphill sorry I've had a busy week and haven't had a chance to take a deep look yet. Was going to mention, too, that I'm about to go on vacation for 10 days. So I'll get back to you start of November. 

If this doesn't lead to any internal maintenance issues (we have some internal code built on top of OpenSpiel that is not open-sourced) nor affects the binary wheels that we distribute for Python users, or e.g. breaks our native Windows build, then I am hugely in favor of the change. I just need to be sure it doesn't disrupt anything. 

Thanks for doing this! We don't normally get contributions that take this form.. takes some rare skills to first notice it and even more so to fix it.

lanctot:
BTW so far it's looking like that wouldn't be the case, so I'm not worried! But I just have to make sure before we import it.

jthemphill:
Thank you very much, and have an awesome vacation!

open_spiel/games/CMakeLists.txt
```diff
@@ -192,18 +192,21 @@ if (${OPEN_SPIEL_BUILD_WITH_ACPC})
   set(GAME_SOURCES ${GAME_SOURCES} universal_poker/universal_poker.cc universal_poker/universal_poker.h)
 endif()
 
-add_library (games OBJECT ${GAME_SOURCES})
-
+add_library (games SHARED ${GAME_SOURCES})
```
lanctot, line 195: Are you sure this won't enforce the change to all binaries built within OpenSpiel, i.e. not just keep the dynamic linking local to the tests?

lanctot: Oh wait, no.. because they still link to them using TARGET_OBJECTS: https://github.com/google-deepmind/open_spiel/blob/a95a9d6589a57e9c7587aa228785aae16f7597f3/open_spiel/CMakeLists.txt#L223

Ok!

open_spiel/games/CMakeLists.txt
```diff
@@ -272,355 +275,120 @@ add_library(bridge_double_dummy_solver OBJECT
...
-add_executable(trade_comm_test trade_comm/trade_comm_test.cc ${OPEN_SPIEL_OBJECTS}
-               $<TARGET_OBJECTS:tests>)
-add_test(trade_comm_test trade_comm_test)
-
-add_executable(ultimate_tic_tac_toe_test ultimate_tic_tac_toe/ultimate_tic_tac_toe_test.cc
-               ${OPEN_SPIEL_OBJECTS} $<TARGET_OBJECTS:tests>)
-add_test(ultimate_tic_tac_toe_test ultimate_tic_tac_toe_test)
+target_link_libraries(games PRIVATE bridge_double_dummy_solver)
+
+macro(add_game_executable)
```
lanctot, line 280: much cleaner
such one line
so neat!
wow

:)

jthemphill: After this PR lands, I have another speedup in the pipeline which [precompiles](https://cmake.org/cmake/help/latest/command/target_precompile_headers.html) `spiel.h` and friends. This macro makes it easy to add `target_precompile_headers(${_NAME} REUSE_FROM open_spiel_core)` to each target😉

jthemphill:
[`nfg_games_test` failed CI on macos-12](https://github.com/google-deepmind/open_spiel/actions/runs/6552214007/job/17908701682?pr=1126#step:6:1954) with error 

```
Spiel Fatal Error: /Users/runner/work/open_spiel/open_spiel/open_spiel/games/nfg_game/nfg_game_test.cc:62 CHECK_TRUE(matrix_game != nullptr)
```

But the test passes locally, and stepping through the code, it's not clear why this would be a `nullptr` at all. Opening a PR to debug this issue further.

jthemphill:
I think `nfg_game_test` failed because `dynamic_cast<MatrixGame>()` failed, and I think *that's* because `matrix_game.cc` was linked in twice...

```sh
# Built with `-fsanitize=address -fsanitize=undefined`
$ games/nfg_game_test 
=================================================================
==43819==ERROR: AddressSanitizer: odr-violation (0x55b4d98283e0):
  [1] size=320 'vtable for open_spiel::matrix_game::MatrixState' /home/jhemphill/oss/open_spiel/open_spiel/matrix_game.cc
  [2] size=320 'vtable for open_spiel::matrix_game::MatrixState' /home/jhemphill/oss/open_spiel/open_spiel/matrix_game.cc
These globals were registered at these points:
  [1]:
    #0 0x55b4d95957da in __asan_register_globals (/home/jhemphill/oss/open_spiel/build/games/nfg_game_test+0x1da7da) (BuildId: 6d7a0b8c5cad382e91a67328dba4709b33c28695)
    #1 0x55b4d968af8b in asan.module_ctor matrix_game.cc

  [2]:
    #0 0x55b4d95957da in __asan_register_globals (/home/jhemphill/oss/open_spiel/build/games/nfg_game_test+0x1da7da) (BuildId: 6d7a0b8c5cad382e91a67328dba4709b33c28695)
    #1 0x7fc0deeeedeb in asan.module_ctor matrix_game.cc

==43819==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global 'vtable for open_spiel::matrix_game::MatrixState' at /home/jhemphill/oss/open_spiel/open_spiel/matrix_game.cc
==43819==ABORTING
```

...and I *that's* because `open_spiel_core` is implicitly linked into every single target we define, via `link_libraries()`:

https://github.com/google-deepmind/open_spiel/blob/a95a9d6589a57e9c7587aa228785aae16f7597f3/open_spiel/CMakeLists.txt#L208

From [the docs](https://cmake.org/cmake/help/latest/command/link_libraries.html):

> Link libraries to all targets added later. Specify libraries or flags to use when linking any targets created later in the current directory or below by commands such as [`add_executable()`](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable) or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library). **Note:** The [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) command should be preferred whenever possible. Library dependencies are chained automatically, so directory-wide specification of link libraries is rarely needed.

I'll open a new PR which replaces this `link_libraries()` call with `target_link_libraries()` calls in the correct places, which will clear the way for me creating new libraries without ODR violations.

jthemphill:
Can you rerun the failed action? The failure complains about a sha256 mismatch in jaxlib.whl, but I downloaded it, ran a `sha256`, and found that the sha256 on my machine matches the one they expect.

https://github.com/google-deepmind/open_spiel/actions/runs/6686304408/job/18167290661?pr=1126

lanctot:
I started the test but.. yeah, just as I suspected, they are now failing because of a JAX compatibility issue. I suspect what happened was that they finally got rid of Python 3.7 on the test machines just very recently because everything was working fine as of two days ago. 

The real fix is rather involved and will take me a week or two, but I have a short-term fix for now that temporarily disables all the JAX tests (#1134) that I'll put through on Monday.

But the other thing is.. this PR is seemingly getting quite complicated with the external library dependencies.. do you think it is still worth the effort? You don't anticipate it causing any issues down the road?

jthemphill:
> this PR is seemingly getting quite complicated with the external library dependencies.. do you think it is still worth the effort? You don't anticipate it causing any issues down the road?

I think the core idea of linking against a shared library works and works well. This PR's transformation is mechanical - every instance of `add_executable()` becomes `add_open_spiel_executable()`, and then every link to `${OPEN_SPIEL_OBJECTS}` and `$<TARGET_OBJECTS:test>` can be removed, because those are already in the `open_spiel` shared library. That's all the PR is now - I abandoned more complex approaches from before.

I wish I had locally tested more by enabling all the different compilation flags and compiling with each of the libraries, instead of relying on CI to throw the PR back to me every time I missed some compilation flag combination. I think CI will pass this time if the jaxlib wheel issue is fixed, but I understand that the size and CI churn of this PR feels risky. I think deciding if you want to go forward with this is your call as the maintainer.

lanctot:
Closed on purpose? (The issues with the jax tests were fixed last week.)

Apologies, I've been busy fixing something else for an impending release in the next few days. I was planning to get back to this eventually, just got a bit busy with work and other maintenance stuff.

jthemphill:
Yeah, I was clearing my PR queue, and based on your last comment I just wasn't sure if you want to go forward with this PR in this form.

I'm still willing to make this green if you agree with the direction.

lanctot:
> Yeah, I was clearing my PR queue, and based on your last comment I just wasn't sure if you want to go forward with this PR in this form.

Ahh, yeah fair enough. But: don't read too much into it yet. I tend to adopt "if it ain't broke, don't fix it" as a default because I need to minimize time spent on maintenance because I have very little time to devote to it.

> I'm still willing to make this green if you agree with the direction.

Yeah, apologies for the lack of response up to now. I've just had a few distractions (vacation, sickness, and a higher-priority thing to address first). But I'll get to it soon, promise!

jthemphill:
Thank you for all the time you've spent on this PR, and I hope you and your family are well. I know how hard it is to maintain an OSS project within a large company, and I don't want you to feel pressured to accept a PR if you think doing so will make your job harder!

lanctot:
Hi @jthemphill,

Thanks for the effort you put into this. Unfortunately with withering resources dedicated to maintenance, it is unrealistic that we will be able to import this, so I will close it for now. 

If you work on it more, please feel free to reopen.

PR #1171, opened 2024-01-29
Adding game german_whist_foregame

willmcgowan:
German Whist is a 2 player imperfect information trick-taking card game. The rules are outlined here https://en.wikipedia.org/wiki/German_whist#:~:text=German%20whist%20is%20a%20variant,is%20probably%20of%20British%20origin. 

Modified from previous pull request, no longer requires x86intrin.h or bmi2 support.


willmcgowan:
I hadn't added the game name to the pyspiel tests expected games.

lanctot:
> @lanctot Should I remove ismcts_gwhist from the PR?

No it's ok, you can leave it. Good to have an example and it's in the right place.

lanctot:
I don't know what is going on but the tests keep failing repeatedly after your last commit. Can you pull changes from master and push the merge commit? (maybe make an inconsequential change to one of your files if necessary to trigger the tests)

Apologies on the lateness in looking at this, I will take a look as soon as the tests pass.

willmcgowan:
My apologies for the delay in getting back to you, I have issues getting the python tests to run on my device due to some issues with absl. 
api_test.py and the playthrough_test.py are failing. With regard to api_test.py I have added an assert in the informationstatestring implementation.
I can add some more extensive comments and linting by monday.

willmcgowan:
Sorry for taking so long to address some of the issues.
On my device the playthrough test is failing throwing this error:

**Traceback (most recent call last):
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 90, in test_all_games_tested
    missing_games = set(_AVAILABLE_GAMES) - set(_MISSING_GAMES) - set(
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 91, in <genexpr>
    _playthrough_match(os.path.join(path, basename), _SHORTNAME)[1]
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 57, in _playthrough_match
    data = f.read()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 3131: invalid start byte

======================================================================
ERROR: test_playthrough_.DS_Store (__main__.PlaythroughTest)
PlaythroughTest.test_playthrough_.DS_Store
----------------------------------------------------------------------
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 108, in <lambda>
    test_func = lambda self, basename=basename: self.run_test(path, basename)
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/../integration_tests/playthrough_test.py", line 76, in run_test
    expected, actual = generate_playthrough.replay(file_path)
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/algorithms/generate_playthrough.py", line 526, in replay
    original, kwargs = _read_playthrough(filename)
  File "/Users/thomasmcgowan/Documents/Github/open_spiel/open_spiel/python/algorithms/generate_playthrough.py", line 519, in _read_playthrough
    original = f.read()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 3131: invalid start byte

Investigating with hexdump:
xxd -l 16 -s 3131 ./open_spiel/integration_tests/playthroughs/german_whist_foregame.txt 
00000c3b: 3139 3630 3738 292c 2028 3134 2c30 2e30  196078), (14,0.0
This is the segment (where the chance outcomes are being printed).
I thought perhaps because my infostate string/observation state string made use of "/n" that might be causing the issue.

lanctot (approved):

open_spiel/games/german_whist_foregame/german_whist_endgame.cc
```diff
@@ -0,0 +1,729 @@
+// Source Code for an Executable Generating an Endgame Tablebase for German
+// Whist
+
+#include <cassert>
+#include <thread>
```
lanctot, line 5: I did not realize you were using threads. This is not an approved header. Can you remove the dependence on threading in this implementation? From what I can tell this is only used for the end game solver? We have something similar in Bridge but it's not threaded.

open_spiel/games/german_whist_foregame/german_whist_foregame.cc
```diff
@@ -0,0 +1,697 @@
+
+#include "open_spiel/games/german_whist_foregame/german_whist_foregame.h"
+
+#include <filesystem>
```
lanctot, line 4: Similarly this is not an approved header. Could you use the utilities library [here](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/utils/file.h)? I assume that should have everything you need.

lanctot (changes requested):
Hi @willmcgowan, sorry but I missed a few things that are critical. I had started the import but the changes required are non-trivial. Apologies for the delay.

open_spiel/games/german_whist_foregame/german_whist_foregame.cc
willmcgowan: I think it only uses approved headers now.

lanctot (approved):

lanctot:
Apologies for the delays.. the summer ended up a bit busier than I expected!

Should finally be merged some time soon (today or next few days).

lanctot:
Merged into master in https://github.com/google-deepmind/open_spiel/commit/08543bd9b773bdb7a9246066a3f3bd10d2a25199

lanctot:
Hi @willmcgowan, we had to remove the game because it was crashing in the playthrough test. 

I tried a few things but ultimately could not fix it. I believe it might be due to the game construction trying to read the transposition table from disk or maybe taking up too much memory constructing the initial TTable.

Here's the commit: https://github.com/google-deepmind/open_spiel/commit/a66063d2d4b83f26ed9a361d14bbad95f7d2b8d7

We should have a version of the game that will work without a transposition table, i.e. make it optional by default. 

If you want to take a look please start from the fixed version in this commit above. I had to make a number of changes for it to fit our internal style and the version above includes all of them. If you find and fix it, then please resubmit as a new PR.

Issue #1167, opened 2024-01-18
Question about Bridge observation tensor types.


zizhang-qiu:
Hi, I read the code in [bridge.cc](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/games/bridge/bridge.cc) and find the observation tensor only use three types ,
https://github.com/google-deepmind/open_spiel/blob/d8a3babc2703d58dfd7a86974c91a28b9593b90a/open_spiel/games/bridge/bridge.cc#L317

https://github.com/google-deepmind/open_spiel/blob/d8a3babc2703d58dfd7a86974c91a28b9593b90a/open_spiel/games/bridge/bridge.cc#L400

However, in [bridge.h](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/games/bridge/bridge.h), four types of observation are declared.
https://github.com/google-deepmind/open_spiel/blob/d8a3babc2703d58dfd7a86974c91a28b9593b90a/open_spiel/games/bridge/bridge.h#L55

lanctot:
Hi, thanks. @elkhrt any ideas?

elkhrt:
Yes, this looks like a bug - thanks for reporting it. Line 317 should be something like:

```
  const bool defending = (partnership != Partnership(contract_.declarer));
  if (phase_ == Phase::kPlay) ptr[2 + defending] = 1;
```

Do you fancy fixing it?

In general, we've only really used the bidding phase of the bridge implementation, so there may be some small issues with the play phase.

zizhang-qiu:
Thanks for the reply.
To be honest, I don't know how to fix it on github. Should I create a pull request or something? Are there any tutorials?

elkhrt:
Yes, submit a pull request. You'll need to sign an contributor agreement also.
There's no OpenSpiel specific tutorial, but plenty of general ones!

lanctot:
Fixed by PR #1172 . Thanks @zizhang-qiu !

Commit here: https://github.com/google-deepmind/open_spiel/commit/931e39a99ee73412500def0227925f8f19f033fe

PR #1172, opened 2024-01-31
Fix bridge observation tensor

zizhang-qiu:
Fix the bridge observation tensor in #1167.


lanctot:
Thanks @zizhang-qiu !

lanctot:
Can you regenerate the playthroughs?

(run [this script](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/scripts/regenerate_playthroughs.sh))

PR #1175, opened 2024-02-05
Added JAX implementation of CFR

kubicon:
Implementation of CFR that uses JAX. This allows running CFR with GPU acceleration. Speed up over python CFR implementation is ~10-times on CPU only.

The goal was to make it interchangable with the python CFR.


lanctot:
Hi, thanks for the PR!

The reason the wheel tests are failing is because you need an empty __init__.py file in evert directory. Your new one is missing it. Can you add it?

(Don't forget the license header even if the file is empty .. e.g. just copy one of the other ones)

lanctot (approved):

PR #1183, opened 2024-02-23
Add Python 3.12 to list of tests

lanctot:


lanctot:
Will revisit this in a new PR soon.

PR #1192, opened 2024-03-26
Remove nox

morLev:


lanctot:
Ok, as of this commit: https://github.com/google-deepmind/open_spiel/commit/eb18a3b011b90a56f696c942adf10bc4bcc79888 we've now removed Python 3.8 support.

This will hopefully unblock this PR. Can you pull changes from master?

docs/install.md
```diff
@@ -132,14 +131,6 @@ In a nutshell:
     ./open_spiel/scripts/build_and_run_tests.sh
     ```
 
-    **Building and testing using PIP**
-
-    ```bash
-    python3 -m pip install .
```
lanctot, line 138: Leave this section because OpenSpiel can still be installed using pip this way. Just remove the next two lines

lanctot (changes requested):

lanctot (changes requested):
Also can you remove the reference to nox in the comment here: https://github.com/google-deepmind/open_spiel/blob/eb18a3b011b90a56f696c942adf10bc4bcc79888/setup.py#L122

lanctot (changes requested):
Let's please also remove the TRAVIS_USE_NOX environment variable from .github/workflows/* and from open_spiel/scripts/*

lanctot (changes requested):
Great, thanks!

lanctot (approved):

PR #1211, opened 2024-04-18
Added Middle-Eastern dominoes as independent game

Brunozml:
Added 2v2 dominoes, also known as [ME dominoes](https://en.wikipedia.org/wiki/Dominoes#Middle_Eastern_Version), to the repo as an independent game. 

The current PR could be modified so that this is just a separate configuration of block dominoes instead of a whole new game. The game implementations vary sufficiently as to require different `State` classes in my opinion, but it could be done in a manner similar to `tiny_bridge`. 

If this second course of action is suggested, I would appreciate advice on how to implement it correctly according to OpenSpiel's guidelines.


lanctot (changes requested):
Great thanks, a few quick things:

Can you add the test to the Python CMakeLists.txt ([here](Here: https://github.com/google-deepmind/open_spiel/blob/e5d1f0954bed354366967d0d5bf68c64c6bef02a/open_spiel/python/CMakeLists.txt#L218)) 

so that it runs on Github Actions CI?

lanctot (changes requested):
Also, can you add it to the list of games pyspiel_test.py here: https://github.com/google-deepmind/open_spiel/blob/e5d1f0954bed354366967d0d5bf68c64c6bef02a/open_spiel/python/tests/pyspiel_test.py#L109

(That is needed for general simulation tests and to check the list of games matches what is registered.)

lanctot (changes requested):
Can you add the game to the games list?

This one: https://github.com/google-deepmind/open_spiel/blob/master/docs/games.md

Brunozml (commented):
I have updated the boilerplate code and made a few changes to the references (the Middle Eastern version is slightly different from the Latin American version so I referenced the latter).

lanctot:
> I have updated the boilerplate code and made a few changes to the references (the Middle Eastern version is slightly different from the Latin American version so I referenced the latter).

Ok, thanks. I have now triggered the tests on Github Actions.

If this is a specific variant of the more popular game, we should probably not call it just `dominoes` unless it supports the most widely-known "default" version of the game. What do you think about calling it `la_dominoes` or `team_dominoes`?

Brunozml:
You're right, and I like `team_dominoes` ! I will change that as soon as the tests are done.

On a separate note, I'm now trying to work on training the best possible agent based on available OpenSpiel algorithms (or expand to other algorithms if necessary). Due to the size of the game, I'm a bit overwhelmed as to how to go about this process. Do you have any tips/resources that could be useful?

I developed `tiny_dominoes` locally (with two domino blocks per player instead of 7) for OpenSpiel to perhaps perform some more rapid testing, but even basic algorithms such as CFR are incredibly slow when being run locally. I'm thinking of using Google Colab's GPU service, but I haven't managed to install my local OpenSpiel version so that I can test algorithms on `tiny_dominoes.` Any advice helps!

Brunozml (commented):
updated name to `team_dominoes`.

Issue #1216, opened 2024-04-26
dqn_torch_test build failure


tacertain:
I am trying to build Open Spiel for the first time. Using a new Ubuntu 22.04 install under WSL2. I am building with Cuda and libtorch. After a bunch of tinkering, I have gotten down to everything builds and only a single test failure:

```
terminate called after throwing an instance of 'c10::Error'
  what():  masked_fill_ only supports boolean masks, but got mask with dtype int
Exception raised from masked_fill_impl_cpu at ../aten/src/ATen/native/TensorAdvancedIndexing.cpp:1910 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x6c (0x7f3af2ac42ac in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xfa (0x7f3af2a6dcbc in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libc10.so)
frame #2: <unknown function> + 0x20b2d4d (0x7f3ad8cb2d4d in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libtorch_cpu.so)
frame #3: at::native::masked_fill__cpu(at::Tensor&, at::Tensor const&, c10::Scalar const&) + 0x49 (0x7f3ad8cb2df9 in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libtorch_cpu.so)
frame #4: at::_ops::masked_fill__Scalar::call(at::Tensor&, at::Tensor const&, c10::Scalar const&) + 0x16f (0x7f3ad99ad1df in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libtorch_cpu.so)
frame #5: at::native::masked_fill(at::Tensor const&, at::Tensor const&, c10::Scalar const&) + 0xd1 (0x7f3ad8cd6ae1 in /home/certain/GitHub/open_spiel/open_spiel/libtorch/libtorch/lib/libtorch_cpu.so)
[...]

PR #1219, opened 2024-04-28
Update for libtorch 2.0

tacertain:
PyTorch 2.0 deprecated passing a int-valued mask to masked_fill. This change updates one usage in dqn.cc to use bools.

This commit fixes issue #1216

### DIFF
--- a/open_spiel/scripts/global_variables.sh
+++ b/open_spiel/scripts/global_variables.sh
@@ -84,7 +84,7 @@ export OPEN_SPIEL_BUILD_WITH_LIBTORCH="${OPEN_SPIEL_BUILD_WITH_LIBTORCH:-"OFF"}"
 # Note: there are currently known problems with the C++ PyTorch: inteferences
 # with pybind11 versions. Until it is properly fixed, there is a workaround:
 # https://github.com/deepmind/open_spiel/issues/966#issuecomment-1322982393
-export OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL="${OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL:-"https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-1.10.1%2Bcpu.zip"}"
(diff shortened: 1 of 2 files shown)


lanctot:
Nice!!! 

Thanks so much for digging into this!

lanctot:
Two questions:

1. What version of LibTorch did you use? Should we update the link [here](https://github.com/google-deepmind/open_spiel/blob/1208f832568063c36d0e3076069e103d5b00cf5b/open_spiel/scripts/global_variables.sh#L87) to refelct a newer version?
2. Did you run into the pybind11 conflict mentioned on line 86 of the global_variables.sh file linked in (1), or is that now "fixed" as a result of using a newer version of libtorch (issue #966 )?

tacertain:
1. Sure - I'll add that to the PR
2. I did run into that issue and fixed it the same way - I didn't spend any time looking whether there was a better fix.

open_spiel/scripts/global_variables.sh
```diff
@@ -84,7 +84,7 @@ export OPEN_SPIEL_BUILD_WITH_LIBTORCH="${OPEN_SPIEL_BUILD_WITH_LIBTORCH:-"OFF"}"
 # Note: there are currently known problems with the C++ PyTorch: inteferences
 # with pybind11 versions. Until it is properly fixed, there is a workaround:
 # https://github.com/deepmind/open_spiel/issues/966#issuecomment-1322982393
-export OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL="${OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL:-"https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-1.10.1%2Bcpu.zip"}"
+export OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL="${OPEN_SPIEL_BUILD_WITH_LIBTORCH_DOWNLOAD_URL:-"https://download.pytorch.org/libtorch/cu121/libtorch-win-shared-with-deps-2.3.0%2Bcu121.zip"}"
```
lanctot, line 87: We'll need to put the Linux equivalent one here

tacertain:
Third time's a charm?

PR #1221, opened 2024-05-04
Make action IDs relative to pawn position

### LINKED ISSUE #1158: Quoridor Movement Action IDs keep changing
So, I'm playing quoridor, and I was trying to figure out which action IDs corresponded to moving the agent. Therefore, after I placed all the walls, I went to look at the legal actions thinking these should be the movement action IDs. 

However, I noticed that these IDs keep changing based on where I am on the board/the wall configuration nearby me. That seems really weird as I would expect `up` to always be the same action id each time. 

For example, here are the available actions at this board state: 

image

Meanwhile if I then move up (found with trial and error), I have a whole new set of actions to choose from that don't seem in any way related to the previous movement actions. 

image

I don't see how an agent could ever learn to play this game if their movement IDs are not fixed. Is there some scheme I'm just missing?

tacertain:
The original implementation of Quoridor mapped action IDs to absolute board positions.
This meant that moving north had a different ID for every pawn position. This commit
adds the option to make the action IDs be consistent for the same move relative to
pawn position.

This commit fixes #1158.

### DIFF
--- a/open_spiel/games/quoridor/quoridor.cc
+++ b/open_spiel/games/quoridor/quoridor.cc
@@ -52,6 +52,7 @@ const GameType kGameType{
          GameParameter(GameParameter::Type::kInt, /*is_mandatory=*/false)},
         {"ansi_color_output", GameParameter(false)},
         {"players", GameParameter(kMinNumPlayers, false)},
+        {"relative_moves", GameParameter(GameParameter::Type::kBool, false)},
     }};
 
 std::shared_ptr<const Game> Factory(const GameParameters& params) {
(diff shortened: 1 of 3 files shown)


tacertain:
I still need to add some tests with the relative_moves flag set to true, but opening the PR now to get feedback on the method. I have tested it by hand running interactive games.

tacertain:
Also, thoughts on making the default relative? I'm generally against breaking changes, and as pointed out in issue #1158, there may be already-trained models that expect absolute action IDs.

tacertain:
I realized that there's no reason not to have the action IDs for absolute and relative be the same, so I'll update that.

tacertain:
I'm ready to submit the rest of the PR, but want to get advice on what the default should be.

PR #1223, opened 2024-05-09
Normalize the received GTP responses to lowercase

plamentotev:
Some programs return uppercase responses (`PASS`, `G5`, etc). For maximum compatibility normalize the responses to lowercase.


lanctot:
Thanks @plamentotev !

PR #1230, opened 2024-05-18
Add missing bindings in Julia wrapper

findmyway:
Hi @lanctot , this should be ready to merge.


lanctot:
Thanks! Will do this week

PR #1231, opened 2024-05-20
[BUG FIX} fixed small positional argument bug

Brunozml:
problem: Jax implementation of DQN wouldn't work for buffer size >= 1000 (min buffer size for learning) because of small bug in line 282.
```bash
open_spiel/open_spiel/python/jax/dqn.py", line 282, in _loss max_next_q = jax.numpy.where( TypeError: where() got some positional-only arguments passed as keyword arguments: 'x, y' (.venv) brunozorrilla:.../GitHub/open_spiel$
```

Solution:  get rid keyword of arguments:

```python
    max_next_q = jax.numpy.where(
        1 - are_final_steps, max_next_q, jnp.zeros_like(max_next_q))
```

### DIFF
--- a/open_spiel/python/jax/dqn.py
+++ b/open_spiel/python/jax/dqn.py
@@ -280,7 +280,7 @@ class DQN(rl_agent.AbstractAgent):
         (1 - legal_actions_mask) * ILLEGAL_ACTION_LOGITS_PENALTY,
         axis=-1)
     max_next_q = jax.numpy.where(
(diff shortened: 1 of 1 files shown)


lanctot:
Thanks!

lanctot:
Thanks @Brunozml for this.

I was running into an error when importing because it said the changes resulted in an empty import.

Turns out it was fixed recently internally: https://github.com/google-deepmind/open_spiel/commit/f33d30fc2dcda547b361412e037f62a043a5c50d

I guess your copy of master had not been updated yet. Thanks for spotting anyway!

PR #1232, opened 2024-05-21
Convert quoridor movement action IDs to relative

### LINKED ISSUE #1158: Quoridor Movement Action IDs keep changing

tacertain:
The original implementation of Quoridor used absolute position numbering for pawn moves, so moving to b3 was always the same action ID, regardless of where the pawn was. This commit changes to using relative action IDs, so moving directly north is always the same action ID, regardless of what square that is moving to.

See PR #1229 for a deeper discussion of what this change does and why it's only a partial solution to optimzing the game for reinforcement learning.

Fixes #1158


lanctot:
Wow there's really something going on with the wheels test. 

I've restarted the failed jobs, hopefully this does it.

PR #1233, opened 2024-05-22
Adding Spades

i-Madsen:
Spades implementation is largely based off of the Bridge implementation. Runs a single round and returns the points earned/lost from the round (not the overall team's score). Note that currently the parameters use_mercy_rule, mercy_threshold, and win_threshold are not actually used at the moment. They will most likely be moved to and managed by the training script. However, score_partnership_0 and score_partnership_1 are needed in order to determine if a bag penalty is earned from the round.

### DIFF
--- a/open_spiel/games/CMakeLists.txt
+++ b/open_spiel/games/CMakeLists.txt
@@ -161,6 +161,10 @@ set(GAME_SOURCES
   skat/skat.h
   solitaire/solitaire.cc
   solitaire/solitaire.h
+  spades/spades.cc
+  spades/spades.h
+  spades/spades_scoring.cc
+  spades/spades_scoring.h
   stones_and_gems/stones_and_gems.cc
   stones_and_gems/stones_and_gems.h
   tarok/tarok.cc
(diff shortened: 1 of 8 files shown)


lanctot:
Can you generate a new playthrough and add it to the PR?

https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/scripts/generate_new_playthrough.sh

i-Madsen:
Ok, I ran generate_new_playthrough.sh and added the resulting spades.txt - that's the only file needed from this, right?

lanctot:
Correct, yes. Thanks!

lanctot:
Seems like there's a seg fault, there might be a memory issue (buffer overrun or similar). Try building in debug mode (see the flag in the main CMakeLists.txt) and running the test under  valgrind or gdb, it should catch memory out-of-bound issues.

i-Madsen:
@lanctot Going to run some more tests, but after setting build mode to Debug and running build_and_run_tests.sh, it shows a SEGFAULT error with the dou dizhu test:

image

(If I run with --test_only=spades everything passes.)

i-Madsen:
If this is helpful at all, this is from running dou_dizhu_test in gdb:

image

lanctot:
Huh, we'll that's interesting. Ok let me do some checks on my end and get back to you. 

(Actually I can even do those checks with the Spade implementation imported.)

I'm quite loaded at the moment though, so it might be a few days.

lanctot:
Hmm, ok I did some memory sanitizing checks and there appear to be no problems with Spades nor Dou Dizhu. Thinking this is a fluke. Sometimes github CI has random issues. I've imported it and will push to github momentarily. I expect it'll be fine. Thanks again for the great addition!

i-Madsen:
Cool! I'll make another PR eventually as I've been adding game specific functions and tweaking the code a bit to deal with a adding a reward bonus for winning to the returns and such, but glad to have a working version in for now!

PR #1243, opened 2024-06-30
Update Deep CFR implementations to implement Policy

plamentotev:
The `player_id` argument is not need, but still needs to be present in order to implement `Policy`. For example without it PolicyBot breaks as it passes three arguments (self, the game state and the player id).


lanctot:
Thanks!

I am off on vacation at the moment but will take a look when I get back.

(Note: the test failure seems unrelated.. cause by a backward incompatibility in numpy the we have to fix)

lanctot:
Great, looking good! 👍 

(The test failing is something we have to fix in our side.)

PR #1244, opened 2024-07-05
Update heuristic_payoff_table.py: replace np.math.factorial with math…

lanctot:
….factorial


lanctot:
Merged in https://github.com/google-deepmind/open_spiel/commit/490892cf7f570ffea95f429f2896cfa03dc0b9d4

PR #1255, opened 2024-07-30
Improve Windows compatibility

WFKala:
Improve windows installing instruction, add alternative windows methods for File I/O, add Windows SDK specific header option of max(), fix struct initialization

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -56,7 +56,13 @@ CMake, and choose `C:\Users\MyUser\open_spiel\open_spiel\CMakeLists.txt`. CMake
 will then run; once you see `CMake generation finished`, choose Build -> Build
 All. The files will be available in
 `C:\Users\MyUser\open_spiel\open_spiel\out\build\x64-Debug`, when the build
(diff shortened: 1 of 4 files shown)


WFKala:
Hi, I found currently open_spiel cannot be compiled under Windows when following the installing instruction. There are some small necessary fixes to let it work under Windows ( Windows 11 / Visual Studio 2022 Community). 

* `(STRUCT){a,b,c}`  => `STRUCT{a,b,c}` : the original struct initialization method with extra () does not work in MSVC. 
* `realpath` / `PATH_MAX` : there are Linux-only file IO functions/constants and an alternative method on Windows must be provided to let it work.
* `std::max`: Windows SDK has a macro "max" which conflicts with std::max, so one need the Windows-only switch `#define NOMINMAX` to turn off it.
* Other compilation options: to use the correct C++ standard,  deal with chess characters, deal with large object files in pyspiel, one need options `"/std:c++17 /utf-8 /bigobj` in MSVC.

Issue #1266, opened 2024-08-19
EFR code issue with NumPy


lanctot:
CI tests are failing on the EFR code: https://github.com/google-deepmind/open_spiel/actions/runs/10438874702/job/28948184064

Here is the problem:

```
  EFRTest.test_efr_kuhn_poker4 ('csps')
  test_efr_kuhn_poker('csps')
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/absl/testing/parameterized.py", line 323, in bound_param_test
      return test_method(self, testcase_params)
    File "/project/open_spiel/python/algorithms/efr_test.py", line 69, in test_efr_kuhn_poker
      efr_solver = efr.EFRSolver(
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 497, in __init__
      super(EFRSolver, self).__init__(game, deviation_sets)
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 117, in __init__
      self._initialize_info_state_nodes(self._root_node, hist, empty_path_indices)
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 188, in _initialize_info_state_nodes
      self._initialize_info_state_nodes(
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 188, in _initialize_info_state_nodes
      self._initialize_info_state_nodes(
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 212, in _initialize_info_state_nodes
      info_state_node.relizable_deviations = self._deviation_gen(
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 818, in return_cs_partial_sequence
      external = return_all_external_modified_deviations(
    File "/tmp/tmp.qKWdCXq1hQ/venv/lib/python3.10/site-packages/open_spiel/python/algorithms/efr.py", line 1236, in return_all_external_modified_deviations
[...]

PR #1267, opened 2024-08-19
Fix for issue #1266

Jamesflynn1:
I removed an initial None element for a few of the prior_actions_weight arrays. They had no purpose within the code and must have been an artifact from a previous version. I have tested it locally with the newest version of Numpy and it passes the failing test fine.

I am also getting the following deprecation warning with the newer version (which I will also take a look at).

open_spiel/open_spiel/python/algorithms/efr.py:280: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

#1266

### DIFF
--- a/open_spiel/python/algorithms/efr.py
+++ b/open_spiel/python/algorithms/efr.py
@@ -808,7 +808,7 @@ def return_cs_partial_sequence(num_actions, history, prior_legal_actions):
     information set.
   """
   prior_actions_in_memory = history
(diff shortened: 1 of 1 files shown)


lanctot:
Thanks for the quick response!

Jamesflynn1:
No problem at all!

PR #1278, opened 2024-09-12
Add player turn information to Breakthrough's serialization and deserialization function

waterhorse1:
Fix the bug in the original Breakthrough's serialization and deserialization function that misses the player turn information.


lanctot:
This looks good, thanks!

PR #1279, opened 2024-09-26
Fix #1254: Convert pyspiel game state to dict

### LINKED ISSUE #1254: Converting a `pyspiel` game state to a dictionary of array-likes
Would it be possible to convert a `pyspiel` game's `State` object to a dictionary of array-likes and back again, in an efficient way? If that is currently not supported, would it be possible to add this feature?

At the moment, it seems to me that this is not possible. `pyspiel` states are implemented in C++ and bound to Python with pybind11, and it doesn't look like any of the bound methods or properties provide a dict-of-arrays representation of the state.

I'm asking about this because I am looking into adding an environment wrapper class for OpenSpiel to TorchRL. Ideally, the wrapper would be stateless, so the state would need to be provided to the wrapper's step function as part of a TensorDict, which is a dictionary of array-likes.

Some other RL environment libraries support dict-of-arrays representations, like Brax and Jumanji. Just to give an example:

```python
import jumanji
import jax
env = jumanji.make('Snake-v1')
key = jax.random.PRNGKey(0)
state, _ = env.reset(key)

def state_to_dict_of_arrays(state):
    res = {}
    for key, value in state.items():
        if hasattr(value, '_fields'):
            res[key] = {}
            for field in value._fields:
                res[key][field] = jax.numpy.asarray(value)
        else:
            res[key] = jax.numpy.asarray(value)
    
    return res

state_to_dict_of_arrays(state)
```

```
{'body': Array([[False, False, False, False, False, False, False, False, False,
         False, False, False],
         ...
        [False, False, False, False, False, False, False, False, False,
         False, False, False]], dtype=bool),
 'body_state': Array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
...
  'col': Array([2, 4], dtype=int32)},
 'length': Array(1, dtype=int32),
 'step_count': Array(0, dtype=int32),
 'action_mask': Array([ True,  True,  True,  True], dtype=bool),
 'key': Array([2467461003,  428148500], dtype=uint32)}
```

Ayah-Saleh:
This pull request introduces the following changes based off issue #1254:

**Added ToDict() and FromDict() methods to the State class in spiel.**h:

- These methods allow converting the game state to a dictionary of array-like structures and restoring the state from such a dictionary.
- ToDict() converts the current state to a dictionary, and FromDict() restores the state from the dictionary representation.

**Added a new test case in spiel_test.cc**:

- The test verifies that the ToDict() and FromDict() methods work correctly by:
- Converting the state of a Tic-Tac-Toe game to a dictionary.
- Creating a new game state and restoring it using the FromDict() method.
- Making sure that the original and restored states are the same.

### DIFF
--- a/open_spiel/spiel.h
+++ b/open_spiel/spiel.h
@@ -211,6 +211,17 @@ class State {
  public:
   virtual ~State() = default;
 
+  // convert from current state to a dictionary of array-likes
+  virtual std::unordered_map<std::string, std::vector<float>> ToDict() const {
+    SpielFatalError("ToDict is not implemented for this game state.");
+  }
+
+  // method to restore the state from a dictionary of array-likes.
+  virtual void FromDict(const std::unordered_map<std::string, std::vector<float>>& dict) {
+    SpielFatalError("FromDict is not implemented for this game state.");
+  }
+
+
   // Derived classes must call one of these constructors. Note that a state must
   // be passed a pointer to the game which created it. Some methods in some
   // games rely on this and so it must correspond to a valid game object.
--- a/open_spiel/tests/spiel_test.cc
+++ b/open_spiel/tests/spiel_test.cc
@@ -339,6 +339,28 @@ void PolicySerializationTest() {
 }  // namespace testing
 }  // namespace open_spiel
 
+void TestStateToDictAndFromDict() {
+  // Load Tic-Tac-Toe
+  std::shared_ptr<const Game> game = LoadGame("tic_tac_toe");
+  std::unique_ptr<State> state = game->NewInitialState();
+
+  // apply some moves to change the state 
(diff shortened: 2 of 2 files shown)


lanctot:
Thanks @Ayah-Saleh !

Based on the test and description, I suspect that you forgot to include changes to to `tic_tac_toe.h` and `tic_tac_toe.cc`. Can you add them to the PR?

lanctot:
Also can you take a quick look at this: https://github.com/google-deepmind/open_spiel/issues/1254#issuecomment-2379127019. Curious to get your thoughts.

lanctot:
Hello, thanks for this! 

We will be including something similar in the new core API in OpenSpiel 2.0 so I will close this one for now. 

Key your eye out for the release! https://github.com/google-deepmind/open_spiel/issues/1340

PR #1284, opened 2024-10-23
Update dqn.py

eugenevinitsky:
If you do not make this change for recent torch you get the error

 raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'SonnetLinear' object has no attribute 'weight'. Did you mean: '_weight'?


lanctot:
Thanks!

eugenevinitsky:
Temporarily closing, need to double-check something.

PR #1289, opened 2024-11-17
Add game: EinStein würfelt nicht!

giogix2:
Here's my implementation of the game EinStein würfelt nicht!

I've implemented only the basic version of the game with a 5x5 board, as described in the wiki https://en.wikipedia.org/wiki/EinStein_w%C3%BCrfelt_nicht!. I'm not familiar with the other versions, so for now I decided to stick with the simplest version.

There is a common variant where a player wins the game after winning a total of 3 games. In this implementation, winning a single match means winning the game.

Feedbacks are appreciated. Hopefully this game is a fit for open_spiel.


lanctot:
Error looks unrelated to your changes and broken in some of our code. We'll look to get this fixed and let you know when master is up to date with fixes!

giogix2:
Thank you for the heads up. I was planning to dive deeper into this in the next few days, but I was also confused, because tests pass for macos-14, but not for the others. 
I'll wait for the changes in the master branch.

giogix2:
> > Thank you for the heads up. I was planning to dive deeper into this in the next few days, but I was also confused, because tests pass for macos-14, but not for the others. I'll wait for the changes in the master branch.
> 
> Ok fixes applied. Can you pull the most recent changes from master and push the merge commit?

Done.

giogix2:
Some jobs are still failing. Could this still be problem unrelated to these changes? It's a bit difficult to understand what is the root cause from the logs.

Though, there is one test failing for which the reason is clear, specifically: **build_and_test / build (ubuntu-22.04, 3.10, OFF, OFF, OFF)**. This test fails because the initial board setup differs from the one in the playthrough file. In this game implementation, the position of the cubes on the board is initialized randomly, controlled by a random seed passed as an argument. If the seed is set to -1, it’s taken from the clock; otherwise, the default value is 42. 
When running a playthrough with the default parameters, I’d expect the initial setup to be the same (the playthrough tests always pass locally, note: I am using macOS and Python 3.10.15). However, the playthrough tests fail on Ubuntu 22.04 with Python 3.10. Could the random generator produce different values on different operating systems?

If this is the case, I might need to modify the implementation so that the board starts empty, and the first random action is to set up the board. This concept seems similar to what you're doing in Backgammon [here](https://github.com/google-deepmind/open_spiel/blob/f68f2a388a8bf41181b3a323f65fd2d3414ebb63/open_spiel/games/backgammon/backgammon.cc#L422).

giogix2:
I modified the logic for the initial board setup. Previously, the board was initialized randomly using std::shuffle() within the constructor. Now, the allocation of cubes is determined by the values from the chance outcomes, removing the random behavior based on the environment. 
I retested the playthrough on both macOS and Ubuntu, and it passes in both cases.

@lanctot could you please let the CI/CD run again to see if it passes this time?

lanctot:
> I modified the logic for the initial board setup. Previously, the board was initialized randomly using std::shuffle() within the constructor. Now, the allocation of cubes is determined by the values from the chance outcomes, removing the random behavior based on the environment. I retested the playthrough on both macOS and Ubuntu, and it passes in both cases.
> 
> @lanctot could you please let the CI/CD run again to see if it passes this time?

Sorry I had missed your last email. This is the right way to do the random starting board (via chance nodes).

giogix2:
> > I modified the logic for the initial board setup. Previously, the board was initialized randomly using std::shuffle() within the constructor. Now, the allocation of cubes is determined by the values from the chance outcomes, removing the random behavior based on the environment. I retested the playthrough on both macOS and Ubuntu, and it passes in both cases.
> > @lanctot could you please let the CI/CD run again to see if it passes this time?
> 
> Sorry I had missed your last comment. This is indeed the right way to do the random starting board (via chance nodes).

Yes. I didn't realised this initially. But the integration issue above forced me to look at implementation of other similar games and I understood what was wrong.

lanctot:
Something weird is happening with Github Actions CI. The output is odd... Python 3.11 is listing as available via brew but then won't install. 

Not sure what's going on. I'll try it again in a few days. 

(Could be that something changed in the images..)

lanctot:
> Something weird is happening with Github Actions CI. The output is odd... Python 3.11 is listing as available via brew but then won't install.
> 
> Not sure what's going on. I'll try it again in a few days.
> 
> (Could be that something changed in the CI images, but likely just a blip)

Ahh, I know why... [one of my hacks](https://github.com/google-deepmind/open_spiel/blob/f68f2a388a8bf41181b3a323f65fd2d3414ebb63/open_spiel/scripts/install.sh#L292) stopped working on github actions CI. That happens 😅

I'm a bit busy at the moment but I'll fix it mid week and update you when it's done. Apologies.

giogix2:
> > Something weird is happening with Github Actions CI. The output is odd... Python 3.11 is listing as available via brew but then won't install.
> > Not sure what's going on. I'll try it again in a few days.
> > (Could be that something changed in the CI images, but likely just a blip)
> 
> Ahh, I know why... [one of my hacks](https://github.com/google-deepmind/open_spiel/blob/f68f2a388a8bf41181b3a323f65fd2d3414ebb63/open_spiel/scripts/install.sh#L292) stopped working on github actions CI. That happens 😅
> 
> I'm a bit busy at the moment but I'll fix it mid week and update you when it's done. Apologies.

No problem at all. Thank you for looking into it.

lanctot:
> I'm a bit busy at the moment but I'll fix it mid week and update you when it's done. Apologies.

Ok, the fix is now in the master branch. Can you pull changes from master and push the merge commit to trigger the CI tests again?

open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.cc
```diff
@@ -0,0 +1,557 @@
...
+  SPIEL_CHECK_GT(cols_, 1);
+  board_.fill(Cube{Color::kEmpty, -1});
+
+  winner_ = kInvalidPlayer;
+  cubes_[0] = cubes_[1] = kNumPlayerCubes;
+}
+
+void EinsteinWurfeltNichtState::SetupInitialBoard(
+    Player player, Action action) {
+    auto perms = GetAllPermutations();
```
lanctot, line 150: Generally, avoid using auto when you don't have to. It hides the type information unnecessarily.

open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.cc
```diff
@@ -0,0 +1,557 @@
...
+      prev_player_(kBlackPlayerId) {
+  SPIEL_CHECK_GT(rows_, 1);
+  SPIEL_CHECK_GT(cols_, 1);
+  board_.fill(Cube{Color::kEmpty, -1});
+
+  winner_ = kInvalidPlayer;
+  cubes_[0] = cubes_[1] = kNumPlayerCubes;
+}
+
+void EinsteinWurfeltNichtState::SetupInitialBoard(
```
lanctot, line 148: This function generates all 6! permutations and takes factorial space. In this case it's a vector of 720 vectors.

I think we should change this to one of two options:

1. There are known constant-space unranking functions for permutations.  (See this: https://stackoverflow.com/questions/31702845/permutations-unrank)

2. Instead of a single chance node with all the permutations, you can use 12 chances nodes (each with <= 6 outcomes). The chance outcomes labeled 0-5 correspond to the die that is placed in the next position. At first, all of the outcomes are "legal". Next, only the 5 remaining are legal. Next, only the 4 remaining are legal.. etc.

I don't have a strong preference. (2) is a bit less cryptic and consistent with other games, and (1) makes the code a bit more cryptic, but is completely fine as long as it's correct. If you chose 1, I would add a test to make sure the unrank(p) is in the set of permutations and that it uniquely generates all of them if you loop from 0 to 6! - 1.

lanctot (changes requested):
The code looks very good, thanks! 

I have only one major requested change about how the chance node works.

open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.cc
lanctot: Hi @giogix2, 

I completely forgot that we already have the permutation unrank in the code base! ... https://github.com/google-deepmind/open_spiel/blob/b8225eb95a357bd9ab13d098e5e3cddcc4698ff1/open_spiel/utils/combinatorics.h#L139

So let's go with that one (it has been tested independently, so you can assume it works).

giogix2: > Hi @giogix2,
> 
> I just remembered that we already have the permutation unrank in the code base! ...
> 
> https://github.com/google-deepmind/open_spiel/blob/b8225eb95a357bd9ab13d098e5e3cddcc4698ff1/open_spiel/utils/combinatorics.h#L139
> 
> So let's go with that one (it has been tested independently, so you can assume it works).

Oh nice. This helps. I should have asked earlier about this. I missed that in the utils.

Looking into the combinatorics utils I noticed that there is an implementation of Factorial https://github.com/google-deepmind/open_spiel/blob/b8225eb95a357bd9ab13d098e5e3cddcc4698ff1/open_spiel/utils/combinatorics.cc#L56, which is used in pathfinding. I'm wondering if we also should remove this hardcoded value for the permutations https://github.com/giogix2/open_spiel/blob/einstein_wurfelt_nicht/open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.h#L48.

I'll apply the suggestions in the next few days (likely the weekend).

lanctot: I think it's fine to leave this hard-coded since it's a constant and the game doesn't support different board sizes.

open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.cc
giogix2: Done. Since I'm now using UnrankPermutation(), this line has gone. Now all the new types are explicit.

open_spiel/games/einstein_wurfelt_nicht/einstein_wurfelt_nicht.cc
giogix2: Done. I have replaced the logic with UnrankPermutation().

lanctot (approved):

PR #1294, opened 2024-12-02
Fix issue with Python importlib_metadata requirements when building Dockerfile.base

giogix2:
This PR fixes the following issue https://github.com/google-deepmind/open_spiel/issues/1293.
In particular, when building the docker, the following error is thrown
`AttributeError: module 'importlib_metadata' has no attribute 'EntryPoints'`
This solution was suggested in this this [Stackoverflow post](https://stackoverflow.com/questions/64494526/jupiter-notebook-run-error-attributeerror-module-importlib-metadata-has-no-a).

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -21,6 +21,8 @@ COPY . .
 RUN DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata
 RUN ./install.sh
 RUN pip3 install --upgrade setuptools testresources
+# Line below is a workaround for the issue https://github.com/google-deepmind/open_spiel/issues/1293
+RUN pip install importlib_metadata --force-reinstall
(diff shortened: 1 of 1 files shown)


lanctot:
Thank you!

giogix2:
Let me know if this one is failing because of my change. I guess it's not the case, as from what I understand this Dockerfile is not tested in the CI/CD. I was wondering whether it's still the same issue as lat time, but the other to PRs have worked after the latest changes 🤔.

lanctot:
Oh, that `file_test` periodically fails..  I'm never sure why, it's non-deterministic (maybe a race condition). 

So yes, certainly not due to your PR. I've restarted the CI tests. Maybe I can file an issue to remind us to fix it.

PR #1311, opened 2025-02-23
Clean TicTacToe games up

odanrc:
Hey!

I wanted to contribute to another project in my spare time, and this one seemed interesting; however, the website suggested checking out tic tace toe (TTT) for perfect information, which was a great suggestion, but then recommended poker and other games for references of imperfect information. I believe this is counterproductive: TTT is much simpler, and if one uses it to learn how to set up a perfect info game, then it should be a smaller step to follow through with phantom TTT, specially given that the poker code is not easily readable for a newcomer.

The problem is that phantom TTT is not a commonly known variant - can we even find a description longer than a paragraph online? - so I decided to clean it up a bit as I understood its code. I also noticed that TTT and variant's code embedded multiple assumptions about TTT's behavior. My goal with this set of commits (and PRs) is to generate abstractions and make TTT (and variants) more generic and easier to understand. These are the results of these commits:
- Generalized and deduplacted the code in TTT and variants
- Added GridBoard abstraction, which represents a board made of a grid of tiles
- Allowed TTT to be instantiated with other dimensions
- Started implementing a GridBoard::Tile and a Component class

For future PRs I plan to apply these same steps to other games, making implementing games more trivial and standardized.

Possible long-term goals:
- Break the Game class down into multiple parts. It is doing way too much.
- Change int typedefs (Action, Player, etc) to real classes, so that they can have properties, instead of relaying that info to the wrong class.
- Generalize Game::ToString(). There should be an abstract Display/Render class that is able to convert the game state to a visual representation. This representation should not be constrained to a string - i.e., we should be able to polymorphically define if the output will be shown textually, through textures, or other means.
- Make Phantom a transformer class, not specializations of games

Regarding this PR, it gives an overview of how the TTT and variants look like after the changes. I will likely break it down into multiple PRs, but I want to know what is the reviewers' preference first. I tried to make every commit as self-contained and small as possible, so they should be quite easy to review individually  (and I recommend doing so). Do you prefer PRs per commit (or group of very similar commits), per feature, or per number of lines?

Note: On the bright side, the tests caught bugs quite well; on the dark side, though, I often had no idea where to look at to figure out what was breaking, specially when having to deal with python's dynamic typeness. I managed to get through by dumping variables' contents until I found which txt was breaking. Is there a document/tutorial describing how to find what/where things broke? If so, then maybe it should be made more visible in the Developer Guide section. If not, I suggest creating one.


odanrc:
I am confused by this section of the CLA check: ' Uncheck it and re-create the offending commit, ´. What does re-create mean in this context? Force push, or delete the PR and push again?

lanctot:
Hi @odanrc, 

Thanks! 

Can I double-check that you have followed the link at the top of this screenshot and then, once you've signed the CLA, followed the link at the bottom of this screenshot?

image

odanrc:
Hey, @lanctot . It looks like my CLA was voided because I didn't make clear my position in the company. Once I get a confirmation of acceptance I will get back to this PR.

lanctot:
Hi @odanrc ,

I had to fix a breakage in the wheel tests. Could you pull changes from master and push the merge commit?

PR #1315, opened 2025-02-26
Fix pybind11 to a specific commit on smart_holder branch

lanctot:


lanctot:
@rwgk tagging you here. Do you have a suggested change to my pinning to a specific commit on the smart_holder branch in this PR?

lanctot:
Dropping in favor of #1320

PR #1323, opened 2025-03-15
Changes to support Python 3.13

lanctot:
- Add upgraded Python extra deps for version 3.13
- Remove support for Tensorflow
- Move TF1 algorithms to algorithms/tf/* and mark as unmaintained
- Move TF1 examples in examples/tf and add README to warn as unmaintained
- Disable any tests from CI that require TF1 and mark then as no longer tested


lanctot:
Closing in favor of #1333

PR #1324, opened 2025-03-16
Add Cribbage game

lanctot:
- Currently only contains placeholders.
- TODO: everything :)


lanctot:
Having trouble importing this one, will submit it internally instead.

PR #1331, opened 2025-04-19
Add board class in TTT

odanrc:
Add a class for the board in TTT. This helps encapsulate behavior, and drafts a class that could be later reused in other games. In the future it will be expanded with other functions, once more cleanup is merged.


lanctot:
Thanks!

(Apologies for the delays...)

odanrc:
Hey, @lanctot . There is a mismatch between my expectations and the maintenance/discussion level of the repo. As such, I am no longer willing to continue supporting this contribution and https://github.com/google-deepmind/open_spiel/pull/1311. Given that I am apparently unable to delete/decline pull requests (I can't find a button to do so), would you kindly delete/decline them? Otherwise, feel free to take ownership over. Thanks!

lanctot:
No problem. Sorry, but we have a skeleton crew and we had a number of higher-priority PRs to merge and other maintenance to take care of first. Large refactors of core classes like this consume a lot of my time and so it has to be put at the lowest priority. Appreciate the effort nonetheless.

PR #1337, opened 2025-05-14
Fix some build errors with MSVC and update the windows.md

lhamot:
- In **windows.md** remove the `-b smart_holder` parameters since this branch of `pybind11` no more exists
- In **open_spiel/CMakeLists.txt** add the build flags which are required for MSVC and already listed in **windows.md**
- In **open_spiel/games/bargaining/bargaining_instances1000.cc** and **chess/chess960_starting_positions.cc** :
    - Splits raw strings since they are loo long for MSVC
- In **open_spiel/games/hive/hive.cc** fix this error :
    - `error C2397: conversion from 'int64_t' to 'open_spiel::hive::HiveTile' requires a narrowing conversion`
- In **HiveBoard::IsCovered**, replace `find_if` by `any_of` since `find_if` returns an iterator and `IsCovered` expect a `bool`.

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -42,7 +42,7 @@ dependencies (commands adapted from open_spiel/scripts/install.sh)
 cd C:\Users\MyUser
 git clone https://github.com/deepmind/open_spiel.git
 cd open_spiel
-git clone -b smart_holder --single-branch --depth 1 https://github.com/pybind/pybind11.git pybind11
+git clone --single-branch --depth 1 https://github.com/pybind/pybind11.git pybind11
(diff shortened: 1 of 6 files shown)


lanctot:
Thanks @lhamot . It seems like there were some troubles with the tests.

lanctot:
Thank you again @lhamot for this fix!

I wanted to float an idea by you. It would be great to more properly support OpenSpiel on Windows. There is a lot of work to be done there, but one thing that would help is building and testing the pip wheel on Windows.

I know that the package we use to build it (cibuildwheel) does support building and testing Windows pip wheels, but we have never tried looking into that direction. I wonder if it would be possible. Anyway, just wanted to mention it in case you have every done that before and have any idea how easy it is.

PR #1338, opened 2025-05-20
Added a test for `sequence_form_lp` 

alexunderch:
Added a test for `sequence_form_lp` to check the right way of computing the `explotablity` of the solution.
Should I run local tests and if so, how? I didn't find that in the documentation.


alexunderch:
I didn't read everything well but I have just signed the CLA.

PR #1347, opened 2025-06-04
Add Azul game implementation

fenghaolw:
Azul: https://boardgamegeek.com/boardgame/230802/azul

Please review carefully :) I haven't written C++ in years.

I am using Apple M4 so forked #1335 to get it compiling. The .gitignore change is to ignore clangd caches (am I the only one who ran into this?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ Package.resolved
 # Visual Studio generated files
 open_spiel/.vs
 /.env
+.cache/
--- a/open_spiel/games/CMakeLists.txt
+++ b/open_spiel/games/CMakeLists.txt
@@ -1,6 +1,8 @@
 set(GAME_SOURCES
   amazons/amazons.cc
   amazons/amazons.h
+  azul/azul.cc
+  azul/azul.h
   backgammon/backgammon.cc
   backgammon/backgammon.h
   bargaining/bargaining.cc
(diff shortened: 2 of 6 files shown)


lanctot:
Wow.. thanks!

This is great. 

It might be a bit hard to merge right away due to reasons listed in https://github.com/google-deepmind/open_spiel/issues/770 

But, we recently had a success story with Hive and was able to pull it in as a result (we contacted the publishers and they gave us permission). It'd be great if we could do it again here. There are 20+ publishers are listed on the BGG page.. I wonder if you know a "main" one that we can contact.. do you know?

lanctot:
Looked a bit into the publisher.. Asmodee is listed. This could be good news. Hanabi was one we previously got permissions for and it's also published by Asmodee. So it's possible that we _could_ get permission for this one too.

But, it is something that we would need your help to take on. Let me know what you think.

fenghaolw:
Makes sense, hadn't considered the legal implications 😢 

I am happy to take on this if you can give me some pointers. (I work at Google too, if that helps anything :) I assume the primary part is contacting the publishers though)

lanctot:
Please revert `install.sh` as the new abseil still causes some tests to fail on Github CI. (I will try to find time to fix that properly in master--  next week I hope -- as you're the second person to have been blocked by this)

PR #1370, opened 2025-09-30
New staging release: 1.6.2

lanctot:


open_spiel/scripts/test_wheel.sh
visheshrwl, line 1: Hi @lanctot ,

I noticed you added a wheel installation line for cp312 in `test_wheel.sh`:

```bash
- ${PYBIN} -m pip install wheelhouse/open_spiel-*-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
+ ${PYBIN} -m pip install wheelhouse/open_spiel-*-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

I'm curious about the reasoning behind installing that specific wheel. In my testing, this approach causes "not a supported wheel on this platform" errors when the Python version doesn't match the hardcoded wheel versions.

Would a dynamic approach that detects the actual Python version and installs the matching wheel be more appropriate? Something like:
```
PYTHON_VERSION=$(${PYBIN} --version 2>&1 | awk '{print $2}' | cut -d. -f1,2 | tr -d '.')
${PYBIN} -m pip install wheelhouse/open_spiel-*-cp${PYTHON_VERSION}-cp${PYTHON_VERSION}-*.whl
```
This would support all Python versions (3.9-3.13) that we build wheels for, rather than being limited to specific hardcoded versions.

What was the intended behavior with the dual installation approach?

visheshrwl: Got it, thanks for clarifying the context around the dual-level tests and the Ubuntu 24.04 upgrade.

I agree with your point on readability, the dynamic version detection is a bit cryptic at first glance, but it future-proofs the script against similar issues if/when Python version changes again.

PR #1383, opened 2025-10-10
Improve performance of Tensor object creation.

gislan:
For some reason libtorch is very slow when constructing Tensor objects by assigning values one by one. On my box, it could take 200ms to handle a batch of 128 connect-four states (just creating Tensors, not doing inference) on CUDA and 30ms on CPU. With this change, it's <100us.


lanctot:
Thanks! Apologies for the delay in response, we'll pull this in soon!

PR #1396, opened 2025-11-08
Fix two correctness bugs in AlphaZero

gislan:
This fixes two important bugs in the AlphaZero C++ implementation. Both bugs are about using returns for the wrong player. One happens during MCTS, the other during learning.

Without these fixes both the self-play and learning process are very incorrect and the model never converges on anything even remotely usable, at least in my experience.

### DIFF
--- a/open_spiel/algorithms/alpha_zero_torch/vpevaluator.cc
+++ b/open_spiel/algorithms/alpha_zero_torch/vpevaluator.cc
@@ -72,8 +72,12 @@ LRUCacheInfo VPNetEvaluator::CacheInfo() {
 
 std::vector<double> VPNetEvaluator::Evaluate(const State& state) {
   // TODO(author5): currently assumes zero-sum.
-  double p0value = Inference(state).value;
-  return {p0value, -p0value};
+  double value = Inference(state).value;
+  if (state.CurrentPlayer() == 0) {
+    return {value, -value};
+  } else {
+    return {-value, value};
+  }
 }
 
 open_spiel::ActionsAndProbs VPNetEvaluator::Prior(const State& state) {


lanctot:
Thank you! Just a quick note to let you know that there'll be a delay until we can take a look as I'm a bit overloaded at the moment. But appreciate the contribution and will take a look as soon as I can.

jin404861445lan:
[open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/evaluator.py](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/evaluator.py)

Python's alphazero also has the same implementation, so is Python wrong too?

lanctot:
> [open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/evaluator.py](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/evaluator.py)
> 
> Python's alphazero also has the same implementation, so is Python wrong too?

Note: I have not verified/confirmed that they are bugs yet. 

I'd say it's unlikely, though. The Python implementation was quite heavily tested (internal contribution, which we verified with the AlphaZero team) whereas the C++ one was external. The C++ also has not been maintained, so it's possible that there are also issues with it due to several versions of PyTorch upgrades happening since the code was last changed.

lanctot:
I'll check with @tewalds but based on:

https://github.com/google-deepmind/open_spiel/blob/20eeb7232706bd5f30a10a999de270fe82d21ba9/open_spiel/python/algorithms/alpha_zero/evaluator.py#L61

My guess is that there's an explicit assumption that the network is always outputting the value in terms of the first player. (i.e. the indices of this array correspond to player id's.. but I couldn't be sure without looking at the rest of the code.)

gislan:
If you look at, for example, the chess implementation in games/chess/chess.cc the observation vector does not differ depending on which player is making the move. The only difference is one bit, who's to move next. So, with the current code, network will learn to output the odds of winning for white rather than for current player. So everything will work. The network gets correct outcome in the training process (p1_outcome, so whether white wins or not) and the evaluator will output correct value as well (whether p1, that's white, wins).

However, as I understand it, the contract for ObservationTensor() method is that it should return a position from perspective of current player (hence the player argument). This is what happens in, for example, connect4 - the position is flipped depending on who the current player is. Yet the training examples provided to the learning process will label every state from given game with the same outcome. So half of the states will be labeled correctly and half incorrectly (with a flipped sign) and similar issue will happen during MCTS - if state is terminal game will output {p1_outcome, p2_outcome}, but if it's not terminal the evaluator will output {p1_outcome, -p1_outcome} or {p2_outcome, -p2_outcome} depending on which player's turn this was.

So either these are two bugs that flip a sign on the outcome sometimes, or ObservationTensor() contract needs to change to always return observations from perspective of first player and include who's turn is it to move as one of the observations, like chess implementation does.

If your testing was done on chess then current code would work just fine, but that's because chess does not return the observations from current player perspective.

lanctot:
> However, as I understand it, the contract for ObservationTensor() method is that it should return a position from perspective of current player (hence the player argument). 

Actually this is not generally not true in OpenSpiel (and not required by observation tensor).

For most (all?) perfect information games you should get the same observation (including Connect Four, btw: https://github.com/google-deepmind/open_spiel/blob/20eeb7232706bd5f30a10a999de270fe82d21ba9/open_spiel/games/connect_four/connect_four.cc#L275, so I'm curious in what way you're getting different perspectives.)

Some games support a flag "egocentric" that you can pass in when creating the game (e.g. Goofspiel), but not many of them. So at least in OpenSpiel it makes sense to predict one player's value and then just invert it to get the other player's.

lanctot:
Nevermind.. my bad; I missed the `PlayerRelative` part of Connect Four observation tensor code I cited above. 

That's the actual bug. Most of the observations in OpenSpiel are not player-relative and the code is designed under that assumption, so it makes sense that AlphaZero not work well in Connect Four. 

We will fix that to be consistent with the other games.. I'm not sure why that's there. Apologies!

gislan:
It's not just connect4. I see the same thing in checkers:
    // Observation Tensor Representation:
    //  Plane 0: 1's where the current player's pieces are, 0's elsewhere.
    //  Plane 1: 1's where the oppponent's pieces are, 0's elsewhere.
    //  Plane 2: 1's where the current player's crowned pieces are, 0's elsewhere.
    //  Plane 3: 1's where the oppponent's crowned pieces are, 0's elsewhere.
    //  Plane 4: 1's where the empty cells are, 0's elsewhere.

the documentation on State::ObservationTensor():
    // Returns the view of the game, preferably from `player`'s perspective.
    //                                      
    // Implementations should start with (and it's tested in api_test.py):
    //   SPIEL_CHECK_GE(player, 0);
    //   SPIEL_CHECK_LT(player, num_players_);                                    
    virtual void ObservationTensor(Player player,               
                                   absl::Span<float> values) const {

Don't get me wrong, I think both approaches are fine as far as perfect information games are concerned. It's just that right now the codebase is not consistent and if you want to make ObservationTensor() be player-independent you should probably remove that parameter. Any game that uses it will not work correctly without this fix.

lanctot:
Ok, yes - thanks for highlightling those.

Yeah those docs are not consistent with the code :( 

The observations (in perfect information games) have been player-independent for years and we never changed that. In fact the AlphaZero implementations (both, it seems) were built on that assumption. So we won't change that now, but we will definitely fix inconsistent docs and implementations that deviate from it.

I've already fixed the Connect Four one internally (I added a flag "egocentric_obs_tensor" but set it to false by default), and I'll fix these now too.

Sorry that you were affected by this. It's a bit tough to keep docs up to date with the code movement, and it's super frustrating from a user's perspective when the docs don't reflect the code... so identifying them really helps us and others; if you see any more please let us know. 

Thanks again!

lanctot:
Hi @gislan I have update github with recent commits addressing the problems you mentioned above.

Can you pull from master and try running AlphaZero (wihtout the modifications in this PR) on these versions to see if it works (e.g. maybe just on Connect Four) ?

In the case of Connect Four and Checkers I've added a new `egocentric` flag to the game in case people want player-relative observations. 

Thanks again for helping out here.

PR #1399, opened 2025-11-14
Ensure at least one parallel job in build

Aniket-pd:
Replaces os.cpu_count() with a safe minimum of 1 when determining the number of parallel build jobs, and adds a log message showing how many jobs are used. This prevents build failures on platforms where os.cpu_count() returns None or 0, which would otherwise produce an invalid make -jNone flag and abort the build.


lanctot:
Thanks!

Aniket-pd:
Hope it helps!

Issue #1398, opened 2025-11-10
Bug: IndexError in PSROSolver.update_agents when symmetric_game=False


Ynkcc:
### Environment Details

  * **OpenSpiel Version:** 1.6.9
  * **Python Version:** 3.13.5
  * **File:** `open_spiel/python/algorithms/psro_v2/psro_v2.py`

### Problem Description

When running `PSROSolver` with the setting `symmetric_game=False`, an `IndexError: list index out of range` occurs during the first call to `solver.iteration()`.

This bug is triggered inside the `update_agents` method. It appears the logic path for non-symmetric games that *do not* sample from marginals (the default, `sample_from_marginals=False`) is broken.

### Root Cause Analysis

The bug is in the `update_agents` method, specifically in the `if self.sample_from_marginals:` block (approx. lines 329-338).

The `_training_strategy_selector` (called on line 326) appears to *always* return per-player lists, regardless of the `sample_from_marginals` setting. The structure is:

  * `used_policies` = `[ [P0_policies], [P1_policies] ]`
  * `used_indexes` = `[ [P0_indexes], [P1_indexes] ]`

The `if` branch (lines 329-330) correctly handles this structure.

However, the `else` branch (lines 331-338) incorrectly assumes these variables contain lists of *joint tuples*, like `[ (P0_policy, P1_policy), ... ]`.

This false assumption causes two `IndexError`s:

1.  **First (and immediate) Error:** At line 332, `currently_used_policies = [ joint_policy[current_player] for joint_policy in used_policies ]`. When `current_player = 1`, this attempts to access `P0_policies[1]`, causing an `IndexError`.
2.  **Second (Latent) Error:** If the first error were fixed, the code would fail at line 337, `current_indexes = used_indexes[current_player]`. This also incorrectly tries to index the outer list (e.g., `used_indexes[1]`) instead of accessing the per-player sub-list.

The entire `else` block (lines 331-338) seems to be based on a flawed assumption and should not be used. The logic in the `if` block (lines 329-330) is correct for *both* cases.

### How to Reproduce

1.  Initialize `PSROSolver` with any game.
2.  Set `symmetric_game=False` during initialization.
3.  Ensure `sample_from_marginals=False` (which is the default).
4.  Call `solver.iteration()`.
5.  The program will crash with `IndexError: list index out of range` at `psro_v2.py:332`.

### Proposed Solution

[...]

PR #1401, opened 2025-11-17
Fix incorrect policy indexing in PSRO update_agents

AneesPatel:
Remove buggy if/else conditional that incorrectly built currently_used_policies as a joint policy list comprehension. The code now correctly indexes used_policies[current_player] directly, which works for all cases since used_policies is already structured as a list per player.


lanctot:
Thanks!

Can I confirm that this is a fix for https://github.com/google-deepmind/open_spiel/issues/1398 ?

Apologies for any delays in importing as I've been a bit swamped.

AneesPatel:
Yes, this PR should fix #1398. The else branch was incorrectly treating used_policies as joint tuples when it's actually structured as per-player lists. Thanks for your response!

PR #1402, opened 2025-11-25
Add antichess game implementation

mykhail0:
Add an antichess (also known as losing chess) implementation (with tests) based on existing standard chess files. Slightly change chess_board.{h,cc} files to make pawn promotion to a king possible. The behaviour for other chess games must remain unchanged, achieve this with adding an allow_king_promotion flag. This is similar to other existing flags which allow chess_board's logic to be used for other chess variants. Also add various small fixes across the repository.

On a side note, I was using clang-format with Google styleguide as a parameter in order to comply with contribution guidelines, and it changed quite a lot of lines in chess_board.{h,cc} files as I modified them, for example stuff like "const std::string &param" to "const std::string& param" (moving the reference sign). Ultimately, I decided against including those changes since they would clutter the pull request, however I would appreciate some guidance on this as to what to do since it's my first time contributing.


lanctot:
Thanks!

PR #1404, opened 2025-12-05
Support SciPy sparse matrices for alpharank code, implement Stochasically Stable Distributions

l-ouis:
Evolutionary dynamic Markov chains are sparse matrices, so using SciPy sparse matrices and operations can save memory.

Also implements SSD: https://arxiv.org/pdf/1207.1424, which is an exact algorithm that computes the stochastically stable distribution of a Markov chain (and can be used similarly as alpharank)

- [x] move Kosaraju.py into ssd.py and fix formatting
- [x] support sparse matrices in the SSD reduce step 

@Eric-Ewing


lanctot:
Thanks @l-ouis ! Can you sign the CLA?

lanctot (changes requested):
Can you fix the indentation to be consistent? (two-space indent)

E.g kosaraju.py appears to use 4-space indent but there might be others.|

Edit: `original_ssd.py` as well. Or maybe you don't intend to keep that one in the PR? (i.e. is it the original implementation that you're fitting into OpenSpiel, for instance)

l-ouis:
Fixed indentation / moved kosaraju logic into ssd.py. Also implemented more correct sparsity logic to the ssd portion.

lanctot (changes requested):
Finally getting to this. Apologies for the delays.

Can you:

1. Revert pokerkit_wrapper.py. That check has now been to `__init.py__`
2. Please add `ssd_test.py` to the Python tests (in python/CMakeLists.txt)
3. Pull changes from master and push the merge commit.

Thanks!

PR #1411, opened 2025-12-18
refactor(examples): add type hints to tic_tac_toe_qlearner

SH20RAJ:
refactor(examples): add type hints to tic_tac_toe_qlearner

Added type hints to the following functions:
- `pretty_board`
- `command_line_action`
- `eval_against_random_bots`
- `main`

Cleaned up imports and improved docstrings.
Verified by running the script locally.

### DIFF
--- a/open_spiel/python/examples/tic_tac_toe_qlearner.py
+++ b/open_spiel/python/examples/tic_tac_toe_qlearner.py
@@ -27,6 +27,9 @@ from absl import app
 from absl import flags
 import numpy as np
 
+from typing import List, Sequence
+
+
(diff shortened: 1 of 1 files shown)


lanctot:
Thanks! (for all the PRs). 

On type hints: we didn't have them for a long time because the minimum Python version we supported dated back to before type hints were common.  Very happy to add them so long as the minimum version that we supports them (we're currently >= 3.10).

I remember there was a switch for some things like `dict` vs. `typing.Dict` and `list` vs. `typing.List` but I think we're past that now with `>= 3.10`.

PR #1416, opened 2025-12-20
docs: add uv build instructions

### LINKED ISSUE #1363: Add `uv` building system and add `gpu` support/Makefile
Whom it may concern,

Do you think that:
1) adding [`uv`](https://docs.astral.sh/uv/) in addition to pip to support the integration with the modern dependency systems
2) with addition of `jax` versions and for the sake of build simplification, would it be useful to add a `Makefile` for different builds and update the dockerfile with the gpu requirements

Thank you!

SH20RAJ:
Fixes #1363.

This PR adds `uv` build instructions to the documentation. `uv` is a faster alternative to `pip` and is becoming widely used in the Python ecosystem.

Changes:
- Added `uv pip install` instructions for Python-only installation.
- Added `uv venv` and `uv pip install -r requirements.txt` instructions in the source installation section.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -62,6 +62,21 @@ On MacOS, you can install the dependencies via `brew install cmake python3`. For
 clang, you need to install or upgrade XCode and install the command-line
 developer tools.
 
+### Python-only installation via uv
+
+[uv](https://github.com/astral-sh/uv) is an extremely fast Python package manager
+and a drop-in replacement for pip. It can be used to install OpenSpiel:
+
+```bash
+uv pip install open_spiel
+```
+
+Or from source:
+
+```bash
+uv pip install --no-binary=:open_spiel: open_spiel
+```
+
 ## Installation from Source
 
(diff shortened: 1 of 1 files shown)


lanctot:
Have you seen https://github.com/google-deepmind/open_spiel/issues/1363?

This has been proposed before.

It would be good to discuss this first. Do we know it even works? I'd be surprised if it just works out of the box like this? Did you test it? Please follow-up on the thread above.

alexunderch:
it partly works but it doesn't really do what it has to do

however, the build itself needs to be ported with uv, just adding `uv` to the pip wheels will do absolutely nothing

lanctot:
> it partly works but it doesn't really do what it has to do
> 
> however, the build itself needs to be ported with uv, just adding `uv` to the pip wheels will do absolutely nothing

Thanks, that's what I suspected as well. @SH20RAJ did you test these commands?

PR #1419, opened 2025-12-21
DQN refactor

alexunderch:
- [x] Refactored DQN versions with `jax` and `torch` that pass the tests
- [x] Simpler tree-based replay buffers
- [x] Changed the example name to `dqn_` to improve visibility 

Still TBD:
- [ ] `jax` implementation is still very slow with plain python, see https://github.com/google/flax/issues/4045, will fix with different improvements, like caching
- [ ] the only tests are on breakthrough, @lanctot are there any other representative games: tic-tac-toe, hanabi?..

### DIFF
--- a/open_spiel/python/jax/boltzmann_dqn.py
+++ b/open_spiel/python/jax/boltzmann_dqn.py
@@ -88,7 +88,6 @@ class BoltzmannDQN(dqn.DQN):
     epsilon: float,
   ):
     """Returns a selected action and the probabilities of legal actions."""
-
     if epsilon == 0.0:  # greeddy evaluation
       # Soft-max normalized by the action probabilities from the previous
       # Q-network.
--- a/open_spiel/python/jax/dqn.py
+++ b/open_spiel/python/jax/dqn.py
@@ -53,7 +53,8 @@ class ReplayBufferState(NamedTuple):
 
 class ReplayBuffer:
   """ReplayBuffer of fixed size with a FIFO replacement policy.
-    Stored transitions can be sampled uniformly.
+
+  Stored transitions can be sampled uniformly.
     The underlying datastructure is a ring buffer, allowing 0(1) adding and
     sampling.
   """
@@ -85,13 +86,14 @@ class ReplayBuffer:
     experience: chex.ArrayTree,
   ) -> ReplayBufferState:
     """Potentially adds `experience` to the replay buffer.
+
     Args:
       state (ReplayBufferState): current state of the buffer
       experience (chex.ArrayTree): data to be added to the reservoir buffer.
+
     Returns:
       An updated `ReplayBufferState`.
     """
-
     chex.assert_trees_all_equal_dtypes(experience, state.experience)
 
     index = state.entry_index % state.capacity
@@ -118,14 +120,15 @@ class ReplayBuffer:
     rng: chex.PRNGKey, state: ReplayBufferState, num_samples: int
   ) -> Transition:
     """Returns `num_samples` uniformly sampled from the buffer.
+
     Args:
       rng: (chex.PRNGKey): a random state
       state: (ReplayBufferState): a buffer state
       num_samples: (int): number of samples to draw.
+
     Returns:
       An iterable over `num_samples` random elements of the buffer.
     """
-
     max_size = jnp.where(state.is_full, state.capacity, state.entry_index)
     indices = jax.random.randint(
       rng, shape=(num_samples,), minval=0, maxval=max_size
@@ -255,7 +258,6 @@ class DQN(rl_agent.AbstractAgent):
     allow_checkpointing: bool = True,
   ) -> None:
     """Initialize the DQN agent."""
-
     # This call to locals() is used to store every argument used to initialize
     # the class instance, so it can be copied with no hyperparameter change.
     self._kwargs = locals()
@@ -458,6 +460,8 @@ class DQN(rl_agent.AbstractAgent):
 
     Args:
       time_step: an instance of rl_environment.TimeStep.
+      greedy: if to take action greedily of sample it from
+        the logits
 
     Returns:
       A `rl_agent.StepOutput` containing the action probs and chosen action.
@@ -480,6 +484,7 @@ class DQN(rl_agent.AbstractAgent):
     return rl_agent.StepOutput(action=action, probs=probs)
 
   def act_epsilon_greedy(self, info_state, legal_actions, epsilon):
+    """A public wrapper for `_act_epsilon_greedy`."""
     return self._act_epsilon_greedy(
       nn.state(self._q_network),
       jnp.asarray(info_state),
@@ -601,14 +606,15 @@ class DQN(rl_agent.AbstractAgent):
     Action probabilities are given by a softmax over legal q-values.
 
     Args:
+      network_state: `nnx.State` of the Q-network
       info_state: hashable representation of the information state.
       legal_actions: list of legal actions at `info_state`.
+      rng: `chex.PRNGKey` to control randomness
       epsilon: float, probability of taking an exploratory action.
 
     Returns:
       A valid epsilon-greedy action and valid action probabilities.
     """
-
     q_values = self._jittable_inference(network_state, info_state)
     masked_q = jnp.where(legal_actions, q_values, ILLEGAL_ACTION_LOGITS_PENALTY)
     greedy_action = masked_q.argmax()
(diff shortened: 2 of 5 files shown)


alexunderch:
I decided to use different `flax` and `jax` versions to get access to the latest `nnx` api.
```
jax==0.8.1
flax==0.12.1
optax==0.2.6
orbax-checkpoint==0.11.31
```

lanctot:
> I decided to use different `flax` and `jax` versions to get access to the latest `nnx` api.
> 
> ```
> jax==0.8.1
> flax==0.12.1
> optax==0.2.6
> orbax-checkpoint==0.11.31
> ```

Are each of them still supported on Python >= 3.10?

alexunderch:
Docs say that for >=3.10 current top versions are fully compatible...

alexunderch:
Yes, it now fails because of the versions.

@lanctot, what's better for you, to upgrade versions in a separate PR or with this/AlphaZero PR?

lanctot:
> Yes, it now fails because of the versions.
> 
> @lanctot, what's better for you, to upgrade versions in a separate PR or with this/AlphaZero PR?

I'm still not sure if updating versions will work. Sometimes the cascade of dependencies leads to a problem and it makes it impossible if the Python versions are too old. But let me try in a separate PR.

lanctot:
> jax==0.8.1

This one requires >= 3.11, from https://pypi.org/project/jax/0.8.1/

> flax==0.12.1

This one too (>= 3.11), from https://pypi.org/project/flax/0.12.1/

> optax==0.2.6

This one is ok (>= 3.10),  https://pypi.org/project/optax/0.2.6/

> orbax-checkpoint==0.11.31

This one is ok too (>= 3.10), from: https://pypi.org/project/orbax-checkpoint/0.11.31/

lanctot:
Ok...... so what do we do?

Well, 3.10 is causing other problems too ( see https://github.com/google-deepmind/open_spiel/pull/1424 ).

I just checked. Seems like Colab is now using 3.12. Normally I wait until EOL before removing support for a version, but that's quite far away (October '26). In this case it's causing multiple issues so I'd be happy to remove it early. 

Give me a few weeks. I have to check with a few people (most notably the Kaggle Game Arena who are relying on a stable OpenSpiel for their environments). And I could have sworn that the version of Colab I used for the LLM imitation learning was 3.10, so maybe it's lower for the TPU kernels or maybe I was just mistaken.

lanctot:
@alexunderch 

The order I suggest is the following:

- Wait for me to confirm that removing 3.10 support is ok
- I will then remove 3.10 support and update github, will let you know when that's done
- Then, let's update the AlphaZero PR, add it to the tests, and merge it since it's been around for a really long time and looking good
- Then, we come back to this one and the other ones that use the newer nnx (like deep CFR etc.)

Sound good to you?

lanctot:
Moving the discussion of removing Python 3.10 to a new issue: https://github.com/google-deepmind/open_spiel/issues/1425

alexunderch:
Okay, I will do EVA as well. Haven't heard of the algorithm

alexunderch:
DQN breakout (`torch` and `jax`)
image
DQN tic-tac-toe
image

For Kuhn poker, however, NFSP converges to different exploitabilty values, but it's for SGD:
image

alexunderch:
@lanctot I took away `eva` test for `pytorch` because I am not actually sure that the implementation is OK. It may trigger a sampling from an empty buffer which is incompatible to my implementation of it. Not to go deeper into the algo, I decided to come back to it later if it's necessary.

Moreover, jax versions for DQN and NFSP aren't the fastest, to he honest. This morning, I've checked, they don't trigger additional recompilations, so there might be a bottleneck due to interchangable use of pure python and jax. I think we'll get there, I asked a Flax team member if we can do much about that. But we should match pytorch pefr sooner or later across all impls.

alexunderch:
@lanctot for Kuhn poker, exploitabilty of both, jax and torch implementations, is below 0.06 (0.02-0.04). Also, I stripped it a little, reusing some of `DQN` modules.

also, for colabs, I replaced `!pip` (installs the dependencies in a default local env) with `%pip` (that should utilise the venv where the kernel is ran from). Could've been a problem if run the notebooks locally.

Technically, if the tests pass, the PR is in mergable condition. I don't think that I can marginally improve the performance without future `flax` development updates or restructuring the code, which I don't think that might be a good idea to do right away.

alexunderch:
I am sorry, I am not yet used run ALL tests before commits. Will be better

lanctot:
> I am sorry, I am not yet used run ALL tests before commits. Will be better

Don't worry about it -- I just have to press a button. But there might be a delay sometimes if I'm in meetings etc.

lanctot:
Hi @alexunderch,

Thanks for this PR, it's great!

However, I need to request one thing going forward, before we import we need to catch all the Goolge lint errors. This one has many and it's costing us too much time.

Specifically, can you apply Step 9 from this: https://github.com/google-deepmind/open_spiel/blob/master/docs/developer_guide.md#adding-a-game

I will highlight a few of the common issues so you can cross-check that the Python linter with the most recent pylintrc catches them.

open_spiel/python/jax/dqn.py
```diff
@@ -13,57 +13,230 @@
 # limitations under the License.
 """DQN agent implemented in JAX."""
 
-import collections
+from typing import Iterable, NamedTuple, Callable, Any
+from functools import partial
+from enum import StrEnum
```
lanctot: This causes a lint error.

Change to `import enum` and use 1enum.StrEnum` below.

alexunderch:
I see, I was worried about the linter comments, thank you for the correction. I update this and my other active PRs ASAP.

alexunderch:
@lanctot also sorry for reiterating, but can we add pylint and some precommit stuff with a separate PR, kind of finishing this effort https://github.com/google-deepmind/open_spiel/pull/1071

lanctot:
> @lanctot also sorry for reiterating, but can we add pylint and some precommit stuff with a separate PR, kind of finishing this effort #1071

Yes I replied on that thread too. Very open to it at this point if someone gets a minimal version working. See the reply on that thread.

alexunderch:
@lanctot I checked all the files I've changed with the official google's [pylintrc](https://google.github.io/styleguide/pyguide.html)

No errors now, only some formatting issues...

lanctot:
> @lanctot I checked all the files I've changed with the official google's [pylintrc](https://google.github.io/styleguide/pyguide.html)
> 
> No errors now, only some formatting issues...

I think you have to fix all the formatting issues too. They turn to errors preventing the import on our side.

I highlighted one example I saw in the commit.

alexunderch:
Like trialling spaces and tabs, right?

open_spiel/python/jax/boltzmann_dqn.py
```diff
@@ -54,7 +54,7 @@ def _boltzmann_action_probs(
     """Returns a valid soft-max action and action probabilities.
 
     Args:
-      params: State of the Q-network.
+      params: NNX state of the Q-network.
```
lanctot: This won't work. It has to be named the same way as in the function above. Should be e.g. 

`network_state: NNX state of the Q-network.`

Is this one of the minor errors still outstanding? Those turn into actual lint errors (preventing the import) on our side.

lanctot:
> Like trialling spaces and tabs, right?

Those need to go too. I commented on an example above.

alexunderch:
sorry, I should read before I click various buttons

alexunderch:
@lanctot I think you can check internally, I fixed all trailing lines and argument inconsistensies with a more advanced linter, it now should match the doc standard (more or less)...

lanctot:
> @lanctot I think you can check internally, I fixed all trailing lines and argument inconsistensies with a more advanced linter, it now should match the doc standard (more or less)...

Still a lot of issues. It's ok, I'll do it for this one and we'll prioritize ensuring that #1071 catched everything and importing it going forward.

alexunderch:
Wait, so you want to say that missing doc-strings also count? I think I can fix them.

```
D101 Missing docstring in public class
  --> open_spiel/python/jax/nfsp.py:47:7
   |
47 | class MODE(enum.Enum):
   |       ^^^^
48 |   BEST_RESPONSE = 0
49 |   AVERAGE_POLICY = 1
   |
```

Sorry, I just try to understand what should I do to make your life (and my in the end) easier... Because the only things I have ever done is black formatting because I was reliant on external workflows...

PR #1433, opened 2026-01-08
fix install.sh bug with a command failing and not printing the intended error message

Turu-Tamas:
When which didn't find apt-get, the script exited because of set -e and did not print the message afterwards.


lanctot:
Nice, thank you!

PR #1440, opened 2026-01-12
Crazyhouse Issue 1439

gmweinberg:
This is an implementation of the crazyhouse chess variant. It is mostly a clone of
chess but add new piece types (because promoted pieces unpromote when captured)
and support for the "pockets".
The original chess tests pass, except I had to get rid of the perfs and there is no 50 move rule.
I threw in support for the king of the hill variant.

### DIFF
--- a/open_spiel/games/crazyhouse/crazyhouse_board.cc
+++ b/open_spiel/games/crazyhouse/crazyhouse_board.cc
@@ -321,7 +321,7 @@ std::string Move::ToSAN(const CrazyhouseBoard &board) const {
     board.GenerateLegalMoves([&](const Move &move) -> bool {
       if (move.IsDropMove()) {
         return true;
-			}
+      }
       if (move.piece.type != piece.type) {
         return true;  // Continue generating moves.
       }
@@ -665,7 +665,7 @@ void CrazyhouseBoard::GenerateLegalMoves(const MoveYieldFn &yield,
 
           auto ks = king_square;
           if (!(move.IsDropMove()) && at(move.from).type == PieceType::kKing) {
-            ks =  move.to;
+            ks = move.to;
           }
           if (board_copy.UnderAttack(ks, color)) {
             return true;
@@ -1063,7 +1063,7 @@ absl::optional<Move> CrazyhouseBoard::ParseMove(const std::string &move,
 absl::optional<Move> CrazyhouseBoard::ParseSANMove(
     const std::string &move_str) const {
   std::string move = move_str;
-	auto drop_move = ParseDropMove(move);
+  auto drop_move = ParseDropMove(move);
   if (drop_move) {
     return drop_move;
   }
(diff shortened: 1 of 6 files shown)


lanctot:
Thanks!

lanctot (commented):
Since you're adding a pyspiel API specific to crazyhouse, might be good to just add 1-2 basic tests for it, i.e. that call some of the function you're exposing? We normally put these in python/tests/games_<GAME>_test.py so you can see the other examples.

open_spiel/games/crazyhouse/chess960_starting_positions.cc
```diff
@@ -0,0 +1,996 @@
...
+
+#include <string>
+#include <vector>
+
+#include "open_spiel/abseil-cpp/absl/strings/str_split.h"
+
+namespace open_spiel {
+namespace crazyhouse {
+
+constexpr const char* kChess960StartingFens =
```
lanctot: Is this an identical copy ot the one in games/chess? If so, why not just use it directly rather than copying it?

gmweinberg:
Is there anything else I need to do here?

lanctot:
> Is there anything else I need to do here?

Nope, all good. I'm just busy, apologies for the delay. This will be the next one I import, probably in the next few days.

lanctot:
I am getting a crash (memory out of bounds or illegal linstruction), likely due to a memory issue with the following history:

```
3572, 3009, 674, 90, 1842, 263, 1212, 20, 3136, 4698, 1781, 4178, 656, 3576, 3009, 4301, 1386, 3576, 2426, 4302, 2977, 4117, 2467, 1841, 1869, 674, 2109, 1225, 4725, 2393, 30, 652, 1258, 4703, 3110, 397, 4798, 1780, 1198, 1869, 4177, 2540, 601, 308, 3739, 615, 875, 3110, 1598, 3036, 3594, 4756, 364, 1599, 4802, 4692, 745, 4708, 1091, 2393, 691, 1841, 4727, 1549
```

Specifically, calling `IsTerminal` with that history.

I had to make two modifications to the code for it to compile on our side, I will highlight them separately. (I doubt they are the cause, but they might be?)

open_spiel/games/crazyhouse/crazyhouse_board.cc
```diff
@@ -0,0 +1,2470 @@
...
+    if (ptype == PieceType::kKing) return absl::nullopt;
+
+    auto to = SquareFromString(move.substr(2, 2));
+    if (!to) return absl::nullopt;
+
+    // Construct drop move
+    Move drop;
+    drop.from = Square{
+      static_cast<int8_t>(board_size_),
+      static_cast<int>(Pocket::Index(ptype))};
```
lanctot: This was not compiling because:

```
error: non-constant-expression cannot be narrowed from type 'int' to 'int8_t' (aka 'signed char') in initializer list [-Wc++11-narrowing]
```

So I changed it to:

```
drop.from = Square{
      static_cast<int8_t>(board_size_),
      static_cast<int8_t>(Pocket::Index(ptype))};
```

Issue #1441, opened 2026-01-13
Linux arm64 wheels build broken


lanctot:
Seems related to scipy version upgrade from a few days back.

Full link to CI failure: https://github.com/google-deepmind/open_spiel/actions/runs/20957953304/job/60227228831

```
Collecting scipy>=1.10.1 (from open-spiel==1.6.11)
    Downloading scipy-1.17.0.tar.gz (30.4 MB)
       ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 30.4/30.4 MB 107.8 MB/s  0:00:00
    Installing build dependencies: started
    Installing build dependencies: still running...
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
    Installing backend dependencies: started
    Installing backend dependencies: finished with status 'done'
    Preparing metadata (pyproject.toml): started
    Preparing metadata (pyproject.toml): finished with status 'error'
    error: subprocess-exited-with-error
    
    × Preparing metadata (pyproject.toml) did not run successfully.
    │ exit code: 1
    ╰─> [83 lines of output]
        + meson setup /tmp/pip-install-3i5__aj4/scipy_299b03c183864cc59ef3a233fdbb6f3c /tmp/pip-install-3i5__aj4/scipy_299b03c183864cc59ef3a233fdbb6f3c/.mesonpy-zlaow2i6 -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=/tmp/pip-install-3i5__aj4/scipy_299b03c183864cc59ef3a233fdbb6f3c/.mesonpy-zlaow2i6/meson-python-native-file.ini
        The Meson build system
        Version: 1.10.0
        Source dir: /tmp/pip-install-3i5__aj4/scipy_299b03c183864cc59ef3a233fdbb6f3c
        Build dir: /tmp/pip-install-3i5__aj4/scipy_299b03c183864cc59ef3a233fdbb6f3c/.mesonpy-zlaow2i6
        Build type: native build
        Project name: scipy
        Project version: 1.17.0
        C compiler for the host machine: cc (gcc 10.2.1 "cc (GCC) 10.2.1 20210130 (Red Hat 10.2.1-11)")
        C linker for the host machine: cc ld.bfd 2.35-5
[...]

PR #1442, opened 2026-01-13
Update wheels.yml: disable Linux arm64 wheel builds

lanctot:
See #1441 for details


lanctot:
@alexunderch @visheshrwl fyi will merge asap

PR #1443, opened 2026-01-13
Remove Linux wheels from CI tests

lanctot:
See #1441 and log here: https://github.com/google-deepmind/open_spiel/actions/runs/20958555334

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -28,18 +28,18 @@ jobs:
     strategy:
       matrix:
         include:
-        - os: ubuntu-24.04
-          NAME: "Linux"
-          OS_TYPE: "Linux"
-          CI_PYBIN: python3.12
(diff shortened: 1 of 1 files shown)


alexunderch:
@lanctot additionally: macos_15_python_3.13 doesn't run full tests `Skipping full tests for Python 3.14 wheel.` but it's python 3.13.

I think it's about this is due to this in `python_extra_deps.sh`
`elif verlt $PY_VER 3.14; then
  echo "Detected Python version 3.13"`


my tests with updated linux image ran successfully, so maybe this is unnecessary
https://github.com/google-deepmind/open_spiel/actions/runs/20960627184

lanctot:
Yes, this is on purpose: we're not fully supporting Python 3.14 yet. We build the wheel and test it inside the VM (basic tests) but not the full tests. We'll enable that in the future.

PR #1449, opened 2026-01-17
Rcfr patch

alexunderch:
TODO:
- [x] align with `ReservoirBuffer` api 
- [x] patch `pytorch` implementation to pass the tests (wrong broadcasting)
- [x] refactor a lot of legacy `pytorch` bloat
- [x] add `jax` implementation, possibly different tree traversal
- [x] fix `rcfr` colab


alexunderch:
the test is fine, I forgot about `neurd`

also, need to make the traversals less ugly

alexunderch:
This implementation is ugly, for `jax` version, I'd redo it from scratch, keeping the API.

lanctot:
> This implementation is ugly, for `jax` version, I'd redo it from scratch, keeping the API.

Probably not worth doing this: it's unclear how much use it gets. We might delete it in the future if there are no indicators of regular use.

alexunderch:
> Probably not worth doing this: it's unclear how much use it gets. We might delete it in the future if there are no indicators of regular use.

I think that `rcfr` should stay. Motivation:
1. there're not many implementations of the algorithm I've found, however,
2. the algorithm takes an important (especially, currently with LLMs) niche of online no-regret learners that could be a very powerful baseline to have. Plus, the algorithm can use a plug-in training function, which can make the algorithm really versatile.

I want it to stay and make it a bit more future-proof. Not as the first priority so far, but I would rework it for the 2.0. We can scrap `NeuRD`, though, or move it to the `examples`, because it looks as a trivial extenstion of `replicator dynamics` + `rcfr`.

alexunderch:
@lanctot I added a sample `jax` implementation. As I recently found out, there would a need to fully rewrite all env step functions, so I decided not to go deeper with the impl and follow the existing one to let your people incorporate traversals properly. As the tests pass, I can fix all the docs, linting, and such.

- [x] add `jax` implementation, possibly different tree traversal
- [x] fix `rcfr` colab

lanctot:
> @lanctot I added a sample `jax` implementation. As I recently found out, there would a need to fully rewrite all env step functions, so I decided not to go deeper with the impl and follow the existing one to let your people incorporate traversals properly. As the tests pass, I can fix all the docs, linting, and such.

Great, thanks! 

Yeah don't go that deep yet because it'll go to waste. We've already got a number of JAX environments alread coded. We need to push clean some of the simpler ones up and release them. Should be soon, like within the next few weeks. As you hint there's a few important design considerations we need to figure out for JAX-first algorithm implementations that we want to properly discuss among the team first.

alexunderch:
If had a standalone `jax` env wrapper, it's vice versa whould be simpler because traverse logic has to be more or less the same for all the algorithms...

Issue #1434, opened 2026-01-09
Python examples and games (what should we do)


alexunderch:
1. game `atari.py` depends on `stable-baselines3`, which isn't in any of main requirements
2. `bridge_supervised_learning.py` depends on `haiku` Refactor? Keep as it is?
3. `bridge_wb5.py` same
4. `eva.py` is only implemented in `TFv1` Refactor? Delete?
5. `exploitability_descent.py` -- same
6. Do we keep both `pytorch` and `jax` examples given they share the same API, or one of is enough?
7. To fully delete TF, there the following algos left uncovered: `masked_softmax`, `eva`, `neurd` -- is there any priority for them or can delete and include later?

I reckon, we need to locate this kind of issues before you release `2.0`.
I think, it's find to keep haiku and flax together for some time, but it'd be nice to eliminate `TF` and do some housekeeping with requirements...

We address current refactoring with the following:
1. deep CFR and RCFR: https://github.com/google-deepmind/open_spiel/pull/1408
2. deep DQN, Boltzman DQN and NFSP: https://github.com/google-deepmind/open_spiel/pull/1419
3. policy gradients: TBD
@lanctot wdyt

lanctot:
Hey, 

Agree, we should do some cleanup in preparation of OpenSpiel 2.0.

Atari was cool to have but I'm not sure it really fits, it was imported for one specific project, but I don't think it gets much use. It kind of doesn't fit. We can deprecate/remove it.

The bridge examples example should be moved to flax linen or nnx. If nobody does it, we can deprecate / remove. In these two cases I'd rather move to a deprecated/ subdirectory because I think they are easy fixes.

Any examples based on TF1 code still lingering in the examples directory (eva, exploitabiliy descent) should be removed.

Not sure what you mean by the pytorch and jax "sharing the same API". Can you elaborate?

Let's remove everything TF-based that has not been ported to JAX or PyTorch.

Are you feeling generous and looking to help with this cleanup? Because right now, we're quite behind on PR imports which we should do first. But if you submit PRs to do any of the above it'd help a lot!

alexunderch:
I meant the algorithms' implementations have the same API in both frameworks. Just replacing `from open_spiel.python.jax/pytorch import ALGORITHM_NAME` could be used interchangeably.

Yes, I do all I can and can review help with others' unfinished contributions.

lanctot:
[...]

PR #1455, opened 2026-01-22
Remove TF1-based examples (eva.py, exploitability_descent.py)

pantha704:
## Summary

Removes TF1-based examples from the `open_spiel/python/examples/` directory as requested by @lanctot in #1434.

## Changes

- Deleted `eva.py` - uses `tensorflow.compat.v1`
- Deleted `exploitability_descent.py` - uses `tf.disable_v2_behavior()`

## Rationale

Per maintainer guidance in #1434:
> "Any examples based on TF1 code still lingering in the examples directory (eva, exploitability_descent) should be removed."

A PyTorch version of EVA exists at `open_spiel/python/pytorch/eva.py`, so the TF1 example removal is safe.

## Verification

- Verified no other files import these examples via grep
- Clean deletion of 194 lines

Addresses #1434 (partial)


alexunderch:
why only these two....

lanctot:
> why only these two....

Agreed. Probably time to drop this whole directory:

https://github.com/google-deepmind/open_spiel/tree/master/open_spiel/python/algorithms/tf

Can we add it to this PR?

alexunderch:
yes, let's get rid of that. I won't back to `eva` any time soon, almost everything else we have in pytorch/jax or I can re-add it, nothing urgent atp.

then, need to get rid of https://github.com/google-deepmind/open_spiel/blob/09b2873f0a42542a2f81d37be78c150aaa4fb96f/open_spiel/scripts/python_extra_deps.sh#L57

and https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/scripts/find_tensorflow.sh

I guess that's it!

you may feel sceptical about that, because some components related to `tf` may be re-added because, per se, of dataset processing with `grain`, but it would be fine to perform these things with `jax` stack, and not reintroduce new dependency line

lanctot:
Oh yes, good point!

`find_tensorflow.sh` can go for sure. 

I do believe `tensorflow_datasets` may still be used somewhere (maybe also `tensorflow_probability`) but I don't remember exactly. Maybe we can try removing all of those lines and see what happens, and just add back in the ones we still need.

alexunderch:
Yes, let's try, I tried to remove those deps wherever I could. But still could miss smth. `tensorflow_probability` is being installed with `distrax` in the `jax` stack, so should be all good there.

alexunderch:
@lanctot, I will aslo need to delete some other tensorflow things that don't break the tests but stay redundant because they were need to be compatible with `tf.Sess` and models after I am done with refactoring. Should I create a separate PR or we can concentrate on deletion here?

pantha704:
Got it! Since @alexunderch was already working on the broader TF cleanup and refactoring, I'll leave the 
_algorithms/tf/_ directory removal to them to avoid duplicating effort.
Happy to have this PR merged as-is for the examples cleanup, or closed if you'd prefer to consolidate everything in one PR. Let me know what works best!

Issue #1452, opened 2026-01-20
Feature request: Add method that returns the max length of a serialization string.


pranavraj575:
Is there a way to retrieve the maximum possible length of the string that .serialize() method can return for a given game? If not can this be implemented?
We are working with vectorized openspiel states in JAX, and having this feature would be useful to set how long the vectors must be.

Thanks!

sriramsowmithri9807:
@pranavraj575 hey i would like to work on this.

amyssnippet:
@pranavraj575 hope this pr #1456 helps solving the issue

lanctot:
Apologies for the late response @pranavraj575 .

Are you sure it's the serialize() method that you want to use? The serialization is generally linear in the history of actions.  There's no easy way to bound that, and it's probably not something you want to store in tensors because they'll be huge. I don't think it makes sense to have a bound on that and force the games to export it.

Why would you not be use the information state strings, or observation string, or even state->ToString()? Also the information state / observation tensors do have constant-size tensors that are obtainable by calling their Size or Shape functions.

Either way, I think it's probably something that makes more sense to expose on your own branch rather than have it in the master repos.

lanctot:
Note: we're also working on JAX environments currently, which we will be exposing soon too.  We currently have a number of them already done but we're still discussing the right design. Stay tuned, should see them some time in March I would expect.

carlosgmartin:
@lanctot We need to pass PySpiel states through JAX boundaries, so we need to represent them as pytrees. Currently, we do that by representing them via their serialization strings, encoded as a data structure with a fixed-size `buffer` field (a vector) and a `length` field (a scalar).

PR #1456, opened 2026-01-23
feat: Add MaxSerializationLength() to Game API

amyssnippet:
i added a new virtual method `MaxSerializationLength()` to the `Game` class which allows consumers (especially in JAX/vectorized environments) to pre-allocate fixed-size string buffers for state serialization without guessing or over-allocating.

**Changes:**
* Added default implementation in `spiel.cc` based on `MaxGameLength` and action digits.
* Overridden implementation for **Chess** (accounts for FEN + history).
* Overridden implementation for **Tic-Tac-Toe** (fixed constant).

This should help unblock the JAX vectorization work mentioned in #1452.

### DIFF
--- a/open_spiel/games/chess/chess.cc
+++ b/open_spiel/games/chess/chess.cc
@@ -592,5 +592,17 @@ int ChessGame::MaxChanceOutcomes() const {
   }
 }
 
+int ChessGame::MaxSerializationLength() const {
+  // "FEN: " + fen + "\n" + StrJoin(history, "\n") + "\n"
+  // Max FEN length is approximately 100 characters.
+  // Max history length is MaxHistoryLength().
+  // Each action is up to 5 digits (NumDistinctActions() ~ 4674).
+  int max_fen_len = 100;
+  int max_history_len = MaxHistoryLength();
+  int max_action_len = 5;  // log10(4674) + 1
+  return std::strlen("FEN: ") + max_fen_len + 1 +
+         max_history_len * (max_action_len + 1) + 1;
+}
+
 }  // namespace chess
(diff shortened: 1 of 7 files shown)


lanctot:
Thanks!

amyssnippet:
Thanks @lanctot! Since all the checks have passed, is this ready to be merged? Please let me know if there is anything else you need me to adjust.

lanctot:
Oh it will be a while before it's merged, sorry. I haven't even looked closely yet. There are several PRs ahead of this one, and I have to discuss with the team if this is the best way to introduce it. 

It'll be difficult to provide a sensible bound on serialized strings in most games. I'm not sure why the bound should be on the serialized string rather than e.g. the observations. The serialized string for most games will be linear in the history, which can be quite large. I suspect the authors of #1452 didn't have that in mind and don't plan to pre-allocated vectors of maximum history size (they might not realize the serialized strings are linear in the history, they're probably thinking of the serialized strings as observations.. and in that case it would make more sense to have a bound for `State::ToString`). So I need to respond on that thread to clarify this first.

And we definitely won't be able to add the missing ones, so it'd probably be functionality that would have to exists locally for whoever uses them anyway (unless they get contributed for each game, which I seriously doubt).

amyssnippet:
@lanctot Thanks for the feedback. I completely agree that for many games, the serialized string grows linearly with history and can be large.

so could we keep the default implementation safely returning "Not Supported" (or a calculated estimate based on MaxGameLength where possible), and only override it for the specific games (like Chess/TicTacToe) where we know the bounds? This unblocks the specific use case in #1452 without forcing a rewrite of every game in the library.

PR #1459, opened 2026-01-24
Tensforflow-v1, bye you'll be missed!

alexunderch:
* a follow-up for https://github.com/google-deepmind/open_spiel/pull/1455
* removed `open_spiel/python/algorithms/tf`
* removed `open_spiel/scripts/find_tensorflow.sh`
* removed a legacy deps thereof
* some minor docs adjustments to avoid confusion
* `open_spiel/python/algorithms/alpha_zero` is still present due to the separate PR


alexunderch:
`tensorflow` dependencies still present because it's actually tough to reincarnate them again, there's an evidence that they might be necessary.

lanctot:
Cool, thanks!

Can you try also removing the tensorflow deps lines from `python_extra_deps.sh` ?

PR #1477, opened 2026-02-08
Add Gomoku (Issue 1476)

gmweinberg:
This is an implementation of Gomoku. It allows variation of board size, number of dimensions, and the numnber of connected stones you need to win.  Supports wrapping and an "anti" option (getting n in a row loses).
The observation tensor is from a neutral perspective. Also includes a symmetric hash function.

### DIFF
--- a/open_spiel/games/gomoku/gomoku.cc
+++ b/open_spiel/games/gomoku/gomoku.cc
@@ -360,6 +360,41 @@ std::unique_ptr<State> GomokuState::Clone() const {
   return std::unique_ptr<State>(new GomokuState(*this));
 }
 
+std::string GomokuState::Pretty() const {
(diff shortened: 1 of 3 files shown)


lanctot:
Thanks! Might be a while before I get to taking a deeper look but on quick glance it looks great.

lanctot:
Merged in https://github.com/google-deepmind/open_spiel/commit/d8993634a83bad52ba1b429a892fd769971ba4fa

lanctot:
@gmweinberg I imported it before the last two commits, so I think they are missing.

We removed the Undo because it didn't seem to undo the zobrist hashing. 

Feel free to open a new PR if you want those last two additions to be added to the game.

PR #1479, opened 2026-02-10
Removed cvxopt, upgraded to cvxpy, added python3.13-3.14 tests

alexunderch:
Addresses https://github.com/google-deepmind/open_spiel/pull/1472.

* Replaces `cvxopt` with `cvxpy` ~~with additional `GLPK`~~ solver. Tests pass on python3.13 locally, unsure of `3.14`. 
* I decided not to delete `cvxopt` impl for a while, just renamed it and added a guardrail, because `cvxopt` doesn't provide a much low level control and sparsity...
* There is an unedited piece in `jspro`, but it's not covered by tests.

cc @lanctot

### DIFF
--- a/open_spiel/python/algorithms/lp_solver.py
+++ b/open_spiel/python/algorithms/lp_solver.py
@@ -451,7 +451,7 @@ def is_dominated(
     for r in range(num_rows):
       if r != action:
         lp.set_obj_coeff(r, -1)
-    mixture = lp.solve(**options)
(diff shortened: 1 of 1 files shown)


alexunderch:
However, `jspro` requires to replace the solvers. Shall I do so?
https://github.com/google-deepmind/open_spiel/blob/5b63d725f9016c1fe238f5d9ac32de61931232f7/open_spiel/python/algorithms/jpsro.py#L74

lanctot:
Awesome!! Thanks

PR #1485, opened 2026-02-22
Fix TypeError when logging advantage losses in Deep CFR PyTorch

julianmcdonald:
When running Deep CFR PyTorch on Kuhn Poker, calculating advantage losses resulted in a TypeError because integers were being subset using list notation. This changes the logging statement to not assume string concatenation against lists, fixing the logging bug. It provides a simple output of the object.


lanctot:
Thanks!

PR #1475, opened 2026-02-07
Improve error handling in rl_environment.py with custom exceptions and validation

abhaygoudannavar:
This PR improves error handling in the Python RL environment by replacing generic assertions with custom, descriptive exceptions and adding parameter validation.

## Changes
- **Added Custom Exceptions**: Introduced [RLEnvironmentError](cci:2://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/open_spiel/python/rl_environment.py:58:0-60:8), [IllegalActionError](cci:2://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/open_spiel/python/rl_environment.py:63:0-65:8), [InvalidStateError](cci:2://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/open_spiel/python/rl_environment.py:68:0-70:8), and [InvalidParameterError](cci:2://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/open_spiel/python/rl_environment.py:73:0-75:8).
- **Replaced Assertions**: Replaced bare `assert` statements with proper check-and-raise logic to prevent issues from being masked by `-O` flags.
- **Added Validation**:
  - [discount](cci:1://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/verify_fixes.py:20:4-33:20) must now be in the range [0, 1].
  - [step()](cci:1://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/open_spiel/python/rl_environment.py:281:2-321:31) validates action count.
- **Improved Messages**: Error messages now provide more context (e.g., "Expected 1 action, got 2").

## Verification
- Added verification script [test_error_handling.py](cci:7://file://wsl.localhost/Ubuntu/home/abhay/openspiel/open_spiel/test_error_handling.py:0:0-0:0) to confirm new exceptions are raised correctly.
- Verified existing tests pass.
- Verified discount boundary checks work as expected.

closes #1471

PR #1488, opened 2026-03-03
Fix/alphazero libtorch bottlenecks 1487

### LINKED ISSUE #1487: Fix 3 bottlenecks in LibTorch AlphaZero inference/training ( vpnet.cc )
## Bottleneck 1 — Redundant .clone() after .to(device)
**Description**
In both VPNetModel::Inference() and VPNetModel::Learn(), tensors are created via torch::from_blob(), moved to the target device with .to(), and then immediately cloned:

``` cpp
// vpnet.cc, lines 172–176 (Inference) and 235–254 (Learn)
torch::Tensor torch_inf_inputs =
    torch::from_blob(raw_observations.data(),
                     {inference_batch_size, flat_input_size_})
        .to(torch_device_)
        .clone();   // ← Redundant
```
**Why this is a problem**
- torch::from_blob() creates a non-owning tensor that references the raw std::vector data — it must be copied before the vector goes out of scope.
- However, .to(device) already returns a new tensor whenever the source and destination devices differ (CPU → CUDA), or even on the same device since the from_blob tensor doesn't own its memory.
- The subsequent .clone() allocates a second new tensor and copies the data again unnecessarily.
- This pattern appears 6 times in the file (3 in 
Inference()
, 3 in 
Learn()
), doubling memory allocation and data movement on every single call.
**Suggested fix**
Replace .to(torch_device_).clone() with .to(torch_device_) when the device differs from CPU, or use .clone().to(torch_device_) to be safe in all cases (clone from the fast CPU-side blob, then move once):

``` cpp
// Option A: Clone first (from CPU blob), then move to device (single device transfer)
torch::Tensor torch_inf_inputs =
    torch::from_blob(raw_observations.data(),
                     {inference_batch_size, flat_input_size_})
        .clone()
        .to(torch_device_);
// Option B: If always crossing devices, .to() alone suffices
torch::Tensor torch_inf_inputs =
    torch::from_blob(raw_observations.data(),
                     {inference_batch_size, flat_input_size_})
        .to(torch_device_);  // Already creates a new tensor
```
## Bottleneck 2 — Missing torch::NoGradGuard during inference
**Description**
VPNetModel::Inference() correctly sets the model to eval mode but does not disable gradient tracking:

``` cpp
// vpnet.cc, lines 184–187
model_->eval();
// ⚠️ Missing: torch::NoGradGuard no_grad;
std::vector<torch::Tensor> torch_outputs =
    model_(torch_inf_inputs, torch_inf_legal_mask);
```
**Why this is a problem**
- Without NoGradGuard, LibTorch constructs the full autograd computation graph for every forward pass during inference.
- This wastes memory (storing intermediate activations for backprop that will never happen) and compute (graph bookkeeping overhead).
- This is the C++ equivalent of forgetting with torch.no_grad(): in Python — a well-known performance pitfall.
- The overhead is particularly significant here because inference is called thousands of times per second from actor and evaluator threads during AlphaZero self-play.
Suggested fix
``` cpp
model_->eval();
torch::NoGradGuard no_grad;  // ← Add this single line
std::vector<torch::Tensor> torch_outputs =
    model_(torch_inf_inputs, torch_inf_legal_mask);
```

## Bottleneck 3 — Per-element .item<>() calls in output extraction loop
**Description**
After inference, results are extracted from GPU tensors using per-element .item<>() calls inside a nested loop:

``` cpp
// vpnet.cc, lines 195–206
for (int batch = 0; batch < inference_batch_size; ++batch) {
    double value = value_batch[batch].item<double>();   // Sync point #1
    ActionsAndProbs state_policy;
    state_policy.reserve(inputs[batch].legal_actions.size());
    for (Action action : inputs[batch].legal_actions) {
        state_policy.push_back(
            {action, policy_batch[batch][action].item<float>()});  // Sync point #N
    }
}
```
**Why this is a problem**
- Each .item<T>() call on a CUDA tensor triggers a GPU-to-CPU synchronization barrier, blocking the calling thread until the GPU operation completes and the scalar value is transferred.
- For a typical batch of 128 states with ~10 legal actions each, this creates ~1,408 individual synchronization points per inference call.
- In the AlphaZero pipeline, VPNetEvaluator::Runner() (
vpevaluator.cc
) calls 
Inference()
 continuously in a tight loop, making this a dominant source of latency.
**Suggested fix**
Move the entire tensor to CPU in a single transfer, then use accessor<> or data_ptr<> for zero-overhead element access:

``` cpp
// Move to CPU once (single sync point)
torch::Tensor value_cpu = value_batch.to(torch::kCPU).contiguous();
torch::Tensor policy_cpu = policy_batch.to(torch::kCPU).contiguous();
auto value_acc = value_cpu.accessor<float, 2>();   // 2D: [batch, 1]
auto policy_acc = policy_cpu.accessor<float, 2>(); // 2D: [batch, num_actions]
for (int batch = 0; batch < inference_batch_size; ++batch) {
    double value = static_cast<double>(value_acc[batch][0]);
    ActionsAndProbs state_policy;
    state_policy.reserve(inputs[batch].legal_actions.size());
    for (Action action : inputs[batch].legal_actions) {
        state_policy.push_back({action, policy_acc[batch][action]});
    }
    output.push_back({value, state_policy});
}
```

abhaygoudannavar:
## Changes
1. Reorder .to(device).clone() → .clone().to(device) (6 occurrences)
torch::from_blob() creates a non-owning tensor over a raw std::vector. The previous pattern .to(device).clone() performs two allocations: .to() creates a new tensor on the target device, then .clone() duplicates it again. By cloning first (to take ownership on CPU), then moving to device, we eliminate the redundant allocation.

``` diff
-  torch::from_blob(data, shape).to(torch_device_).clone();
+  torch::from_blob(data, shape).clone().to(torch_device_);
```
Applies to 2 tensors in Inference() and 4 tensors in Learn().

2. Add torch::NoGradGuard during inference Inference() set model_->eval() but never disabled gradient tracking. Without NoGradGuard, LibTorch builds the full autograd computation graph on every forward pass — wasting memory and compute for operations that never require backpropagation. This is the C++ equivalent of Python's torch.no_grad().

``` diff
model_->eval();
+ torch::NoGradGuard no_grad;
  auto outputs = model_(inputs, mask);
```
3. Replace per-element .item<>() with batch .to(kCPU) + accessor<>
The output extraction loop previously called .item<>() on individual tensor elements inside a nested loop. Each .item<>() on a CUDA tensor triggers a GPU→CPU synchronization barrier. For a batch of 128 states with ~10 legal actions each, this was ~1,400 sync points per inference call.
Now the entire output tensor is moved to CPU in a single transfer, and elements are accessed via accessor<> with zero overhead.

``` diff
- double value = value_batch[batch].item<double>();
- policy_batch[batch][action].item<float>();
+ torch::Tensor value_cpu = outputs[0].to(torch::kCPU).contiguous();
+ auto value_acc = value_cpu.accessor<float, 2>();
+ double value = static_cast<double>(value_acc[batch][0]);
Reduces sync points from O(batch_size × avg_legal_actions) to O(1).
```

## Testing
These changes are covered by the existing alpha_zero_torch_vpnet_test. No new tests are needed since the fixes are purely performance-related with identical outputs.

```bash
cmake --build build --target alpha_zero_torch_vpnet_test
./build/algorithms/alpha_zero_torch/alpha_zero_torch_vpnet_test
```

closes #1487


lanctot:
Thanks!

lanctot:
> Can you please revert `rl_environment_test.py` and `rl_environment.py` so that they don't show up in the PR (that were brought in by a separate PR, I think)

Similarly for `full_test.py`, `verify_fixes.py`, and `verify_game.py`. I suspect these were included by mistake or beloing in #1475  rather than this PR.

lanctot (approved):

PR #1489, opened 2026-03-04
Adding support of subgame equilibirum computation in IIEFGs.

alexunderch:
- [x] Implemented optimisation problem from the [paper](https://www.itu.dk/~trbj/papers/seqeqsoda.pdf) for 2p0s games in `python`
- [x] Added games "Guess the ace" and "Kuhn poker with raise action"
- [x] Add more tests to understand if the implementation computes the *sequential* equilibrium
- [ ] ~~Add the implementation for C++ ortools version~~

Now the implementation just extends current solver with noise in accordance with the paper. However, there're 2 questions, whether I need explicitly state the slack variables (currently they are under control of the solver, AIs told me this) and to implement all the tests + pygambit eval, showing that it works. To make it simpler to criticise, I make the PR public.

Current simplest tests (just computing values and policies pass)
ETA: 2-3 weeks.

@lanctot is it okay to start with that simple of extenstion? Because we don't actually converge to the equilibrium at the limit, only approximately. For the former, iirc, I need to implement one of: symbolic solver or Lemke algo.

### DIFF
--- a/open_spiel/python/algorithms/lp_solver.py
+++ b/open_spiel/python/algorithms/lp_solver.py
@@ -139,10 +139,10 @@ class LinearProgram:
       ub: an upper-bound value for this variable
     """
     var = self._vars.get(label)
-    # if var is not None:
-    #   # Do not re-add, but ensure it's the same
-    #   assert var.lb == lb and var.ub == ub
-    #   return
+    if var is not None:
+      # Do not re-add, but ensure it's the same
+      assert var.lb == lb and var.ub == ub
+      return
     var = _Variable(len(self._var_list), lb, ub)
     self._vars[label] = var
     self._var_list.append(var)


lanctot:
Thanks!

lanctot:
BTW about this: "Add the implementation for C++ ortools version": we've remove the testing of the OpenSpiel build with OR-Tools from GitHub Actions and will probably mark it as unmaintained as of OpenSpiel 2.0. So might not be worth the effort. Having the Python implementation is enough -- if anybody wants to extend it to C++, they can. :)

alexunderch:
@lanctot , I made several substantial changes:

1. Slack variables are really handled by the solver, however, it zeroes out its coefficients in the objective. Thus, I explicitly introduced slack variables, `u` and `v` to solve the dual problem. The problem construction is updated to support recursion over slacks as well.
2. I added an explonential noise schedule with a degree proportional to the depth of the tree.
3. Added 2 games: `Guess the Ace` and `Poker with raise` that are tested for convergence and rationality. The only thing is, afaik, open spiel's efg format has some subtle differences with `gambit`'s one. Thus, I've generated `efg`'s with AI, testing them with `graphviz`. SO, you might want to give them a look.
4. Tests (that I've got) reproduce the paper's findings, I guess.

alexunderch:
related to https://github.com/google-deepmind/open_spiel/pull/1467, however, I don't understand why the solution is like that. There's some logic behind the comments, but there's a question if all together can be viewed as some method.

PR #1507, opened 2026-03-25
Fix DeepCFR advantage network learning

martinboca:
Fix a comparison bug in `_learn_advantage_network` that prevented the network from ever being trained.

Instead of checking whether `info_state` is empty, it compares each element of the array against zero (producing a boolean array), and then checks the length of that boolean array. Since this boolean array always has the same number of elements as the original, the condition is true (except when the original is empty) and the function returns early without training the advantage network.

### DIFF
--- a/open_spiel/python/pytorch/deep_cfr.py
+++ b/open_spiel/python/pytorch/deep_cfr.py
@@ -562,7 +562,7 @@ class DeepCFRSolver(policy.Policy):
         samples = self._advantage_memories[player].experience
 
       # Ensure some samples have been gathered.
-      if len(samples.info_state == 0):
+      if len(samples.info_state) == 0:
         return None
 
       self._optimizer_advantages[player].zero_grad()


lanctot:
Thanks! Do you happen to have any exploitability graphs on, say, Kuhn or Leduc poker that we can look at? It'd be great to verify that it's working as expected on those benchmark games.

alexunderch:
yes, it's a trivial typo -- well-noticed!
it's really a corner case and shouldn't affect learning outcomes for poker, I reckon, because other than the 0th iteration, buffer should contain some data and the condition wouldn't be triggered

martinboca:
@alexunderch Maybe I'm misunderstanding it, but I think this affects learning more than it seems. Since `samples.info_state == 0` returns another array of the same size, the `len(...)`returns the number of samples, always falling into the early return when the buffer contains data.

alexunderch:
@martinboca it would be the case if `samples.info_state` was a list:
```Python
>>> [0, 1, 0, 1, 0] == 0
False
```
However, with `np/jnp.array` (which I believe buffer samples are) it triggers element-wise comparison:
```Python
>>> np.array([0, 1, 0, 1, 0]) == 0
array([ True, False,  True, False,  True])
```

So, it shouldn't really hurt that much, I think.

Update: I missread your message, I am sorry it might be an issue.

alexunderch:
@lanctot 

Benchmarks for Kuhn Poker:
```
I0325 19:29:50.936964 8478404864 deep_cfr_pytorch.py:88] Deep CFR in kuhn_poker - NashConv: 0.036619428131200844
I0325 19:29:50.937295 8478404864 deep_cfr_pytorch.py:94] Computed player 0 value: -0.06 (expected: -0.06).
I0325 19:29:50.937314 8478404864 deep_cfr_pytorch.py:98] Computed player 1 value: 0.06 (expected: 0.06).
```

For Leduc Poker:
```
I0325 19:32:34.604614 8478404864 deep_cfr_pytorch.py:88] Deep CFR in leduc_poker - NashConv: 0.6801337377649047
I0325 19:32:34.633373 8478404864 deep_cfr_pytorch.py:103] Computed player 0 value: -0.14
I0325 19:32:34.633446 8478404864 deep_cfr_pytorch.py:104] Computed player 1 value: 0.14
```

Which are almost indentical to the original benchmarks https://github.com/google-deepmind/open_spiel/pull/1408#issuecomment-3679664851

I think that it is definitely the issue, rightfully found by @martinboca. I think I've got lucky because the algorithm re-initialised value networks really oftern (due to small game tree sizes), and the advanatage updates didn't really have too much impact. 

TLDR; the same line is correctly done in the strategy update correctly, so it should be like that. The results with fixed hyperparams stay consistent. Recommed merging

Update: `flax` implementation doesn't have this check, and maybe - for good.

I am sorry!

lanctot:
Great, thanks for checking. I'll import it!

alexunderch:
This one for sure, but it motivated me to re-check if the advantage updates do their job. If something is wrong, I report separately.

martinboca:
@lanctot @alexunderch Wait. I was running a test and after the fix I got some errors.
```
OpenSpiel exception: /Users/runner/work/open_spiel/open_spiel/open_spiel/algorithms/history_tree.cc:210 policy_prob >= -(ProbabilityDefaultTolerance())
policy_prob = nan, -(ProbabilityDefaultTolerance()) = -1e-09
```

It seems that replacing the old `torch.FloatTensor` with `torch.tensor(..., dtype=torch.float32, ...)` solves this, but let me test more.

alexunderch:
Hmm, I don't get this exception, and the test ran fine, didn't they

@martinboca, can you give something to reproduce?

lanctot:
Ok I'll wait on the import. Take your time. Yes, the CI tests passed.

martinboca:
@lanctot Never mind that `nan` error I just mentioned. It turned out to be an issue with my tests. The PR looks good to me.

Issue #1515, opened 2026-04-02
rbc.cc requires C++20 when building with MSVC


erik-anderson:
I tried building with MSVC, but see that [rbc.cc](https://github.com/google-deepmind/open_spiel/blob/8665ebe9161c20c4d5dae3e9d1b4b763343f5710/open_spiel/games/rbc/rbc.cc#L608) and [rbc.h](https://github.com/google-deepmind/open_spiel/blob/8665ebe9161c20c4d5dae3e9d1b4b763343f5710/open_spiel/games/rbc/rbc.h#L161) are using designated initializers, e.g.:
```
  return SenseWindow{
      .min_corner = chess::Square{.x = min_col, .y = min_row},
      .max_corner = chess::Square{.x = max_col, .y = max_row},
  };
```

Since this project is using C++17, it fails with a `error C7555: use of designated initializes requires at least '/std:c++20'` error.

Apparently gcc and clang support this pre-C++20 but would warn if using `-pedantic`.

I see prior issue #864 which was fixed by removing the use of the language feature. And #1494 fixed some similar C++20 usages in these files.

Things appear to build fine with C++20, so maybe worth considering moving forward to that?

lanctot:
Thanks, good point. Might be time to bump up now. Will take a look when back from vacation.

Can I ask you to submit a PR to update the Windows documentation to add a mention of this? (windows.md)

erik-anderson:
To build with C++20, I needed to modify CMakeLists.txt to change `CMAKE_CXX_STANDARD` to 20 and modify the `if(MSVC)` `CMAKE_CXX_FLAGS` to `/std:c++20`.

What sort of mention were you thinking of? Just calling out that MSVC needs C++20? Or also give a hint to modify CMakeLists.txt to make that happen?

lanctot:
I kind of meant a temporary mention to change it manually in the Windows instructions.

But, actually, how about we just try changing to C++20 in the CMakeLists.txt and see if the GitHub Actions tests pass. If they do, we can likely update the whole repos now.

lanctot:
We were mostly keeping C++17 around for older compilers but I think all the platforms that built the older Python wheels, but they should all support C++20 now.

lanctot:
Fixed by https://github.com/google-deepmind/open_spiel/commit/b7234e314f55f7d4493fa12040f2002a90149ccb , thanks @erik-anderson !

PR #1518, opened 2026-04-05
Compile using C++20 and clean up Windows instructions

erik-anderson:
Per #1515, this change moves the project to compile with C++20.

It also refreshes the Windows build instructions to work and simplifies the WSL (now WSL2) directions by pointing to, after WSL setup, the primary install.md Linux directions.


lanctot:
Thanks! Tests passed. I'll import this when back from vacation (mid to late April).

lanctot:
Hi @erik-anderson , this thread may interest you and would love if you could test Windows via pip: https://github.com/google-deepmind/open_spiel/issues/1531

PR #1520, opened 2026-04-25
Add Snake game (2 or 4 players, simultaneous moves)

bovard:
Multi-player Snake on a rectangular grid. All alive players simultaneously choose a direction; snakes die on wall, head-on, or body collisions; eating the fruit grows the snake and triggers a chance node placing a new fruit on a random empty cell. The episode ends when fewer than two snakes are alive or the horizon is reached.

Parameters: rows (10), columns (10), players (2; must be 2 or 4), horizon (100).


lanctot:
Nice!! Thanks

lanctot:
The Python sim tests failed because 3 players is not supported. Could you add snake to the exclusion list here?

https://github.com/google-deepmind/open_spiel/blob/feab31d385e4aac7093954f6f6656f3d8baa32c9/open_spiel/python/tests/games_sim_test.py#L75

That should fix it.

PR #1532, opened 2026-05-01
Fix multiprocessing fork start method on Windows and Python 3.8+

visheshrwl:


alexunderch:
I think because `fork` is in general outdated and worth replacing due to only being compatible with linux, it's better to let this thing fail and do a cross-platform replacement for `>=python3.12` (`3.14` is an interesting case...)

lanctot:
That failed test is probably unrelated, we've been seeing it recently.. I'll try to fix it this weekend.

alexunderch:
@lanctot it doesn't look like a test failure

it's more of a connection error https://github.com/TheAssassin/AppImageLauncher/issues/548

lanctot:
Ok github issues fixed. Can you pull changes from master and push the merge commit (to retrigger the tests)?

PR #1533, opened 2026-05-01
Replace hardcoded /tmp paths with tempfile.gettempdir()

visheshrwl:
Add import os and import tempfile where missing, then replace every
/tmp/... default with os.path.join(tempfile.gettempdir(), ...) across
15 example files. On Linux this is a no-op (/tmp), on macOS it uses
the system temp dir, and on Windows it resolves to %TEMP% so the
examples no longer fail at startup.


lanctot:
That build failure is unrelated. We've been having issues with it recently. I'll try to address it this weekend.

lanctot:
Ok github issues fixed. Can you pull changes from master and push the merge commit (to retrigger the tests)?

PR #1550, opened 2026-05-30
feat: Implement zero-copy AsyncVectorEnv via shared memory for parallel execution

shivansh023023:
### Description
This PR introduces `AsyncVectorEnv` to `open_spiel/python/vector_env.py`, enabling true parallel execution of environments across multiple CPU cores. 

### Key Architectural Details
* **Zero-Copy Shared Memory:** To avoid severe Inter-Process Communication (IPC) pickling bottlenecks when transmitting large observation tensors (e.g., in Chess or Backgammon), this implementation utilizes `multiprocessing.sharedctypes.RawArray`.
* **Direct Memory Writing:** Worker processes instantiate the C++ environments natively and write `time_step.observations["info_state"]` directly into pre-allocated shared NumPy views.
* **Lightweight Pipes:** The `multiprocessing.Pipe` is strictly reserved for small control payloads (rewards, done flags, step types), completely eliminating the need to serialize massive multi-dimensional arrays.
* **Performance:** In a 16-worker benchmark on `chess` running 1,000 steps, this shared memory architecture achieved a **~2.87x speedup** over the synchronous `SyncVectorEnv`, fully bypassing the IPC overhead that typically throttles complex state transmission.

### DIFF
--- a/open_spiel/python/vector_env.py
+++ b/open_spiel/python/vector_env.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 """A vectorized RL Environment."""
 
+import multiprocessing as mp
+
 
 class SyncVectorEnv(object):
   """A vectorized RL Environment.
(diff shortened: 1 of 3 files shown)


lanctot:
Thanks!

lanctot (changes requested):
- Can you add a test that shows it working on a simple game?
- Can you add the test to the python tests in python/CMakeLists.txt

shivansh023023:
> * Can you add a test that shows it working on a simple game?
> * Can you add the test to the python tests in python/CMakeLists.txt

Done !

lanctot:
The tests failed. Can you take a look? Did they pass locally?

open_spiel/python/vector_env.py
```diff
@@ -76,3 +78,207 @@ def reset(self, envs_to_reset=None):
...
+  def close(self):
+    """Cleanly shuts down all worker processes."""
+    if self.closed:
+      return
+    if self.waiting:
+      for remote in self.remotes:
+        remote.recv()
+    for remote in self.remotes:
+      remote.send(("close", None))
+    for p in self.processes:
```
lanctot: Is there a race condition here?

lanctot: (lines 279-280)

shivansh023023: hi @lanctot there was indeed a potential race condition during shutdown if a worker was stuck on an IPC boundary when close() was triggered.

I've pushed a fix that handles process termination safely. I also resolved an underlying multi-player data alignment bug where info_state arrays were hitting dimension mismatches when unpacked at the master process level.

shivansh023023:
@lanctot i have implemented some chages , can you re run the checks

shivansh023023:
hi @lanctot all the checks are passing now !

lanctot (approved):

lanctot:
Hmm, unfortunately there is an issue with how multiprocessing is handled in Python (internally), so we probably can't support this because it does not work internally. Thanks, though.. we will leave it open for reference here if people want to use it, and please feel free to point people to it!

shivansh023023:
> Hmm, unfortunately there is an issue with how multiprocessing is handled in Python (internally), so we probably can't support this because it does not work internally. Thanks, though.. we will leave it open for reference here if people want to use it, and please feel free to point people to it!

Sure ! @lanctot

PR #1553, opened 2026-06-09
games: add pursuit-evasion game (2-player zero-sum, continuous 2D space)

aeyjeyaryan:
This PR adds a new Python game, `python_pursuit_evasion`, to address issue #843 (Call for New Games).

**Game overview**

A 2-player zero-sum pursuit-evasion game in bounded 2D continuous space:
- **Player 0 (Pursuer):** chooses a move direction each turn from 9 discrete actions (8 compass directions + stay). Movement is unit-length in any direction.
- **Player 1 (Evader):** moves automatically according to a parameterized strategy. Has exactly one no-op legal action.
- **Capture condition:** Euclidean distance < `capture_radius` (default 1.0).
- **Reward:** +1 for pursuer if capture occurs within `max_steps` rounds, −1 otherwise (zero-sum).
- **Game parameters:** `grid_size` (default 10.0), `max_steps` (default 50), `capture_radius` (default 1.0), `evader_strategy` (0–3).

**Evader strategies**

| ID | Name | Behavior |
|----|------|----------|
| 0 | Random | Moves in a uniformly random direction each turn |
| 1 | Constant-velocity | Moves East every turn |
| 2 | Zigzag | Alternates between NE and SE |
| 3 | Adaptive | Moves directly away from the pursuer (unit vector of evader − pursuer) |

**Implementation details**

- Follows the standard OpenSpiel Python game pattern (`pyspiel.Game`, `pyspiel.State`, observer class, registration).
- Observation/information-state tensor: 5-element normalized vector `[pursuer_x/grid_size, pursuer_y/grid_size, evader_x/grid_size, evader_y/grid_size, step/max_steps]`.
- Sequential turn structure: pursuer moves → evader moves → capture check → repeat.

**Tests**

12 tests covering API conformance (`pyspiel.random_sim_test`), all 4 evader strategies, both terminal conditions, legal action counts, zero-sum property, observation tensor shape/range, and deterministic replay.

**Additional context**

This game accompanies a research paper comparing NEAT and PPO under non-stationary opponent strategies (IEEE Access, under submission). The adaptive evader strategy shifts the evasion policy during an episode, creating a non-stationary environment where population-based methods (NEAT) outperform gradient-based methods (PPO). The game is designed for use with OpenSpiel's Python algorithms and can serve as a benchmark for multi-agent RL under strategy distribution shifts.


open_spiel/python/games/pursuit_evasion.py
```diff
@@ -0,0 +1,273 @@
...
+"""Pursuit-evasion game in 2D continuous space.
+
+A 2-player zero-sum game where a pursuer (Player 0) attempts to capture an
+evader (Player 1) within a bounded 2D grid. The evader follows one of four
+parameterized strategies (random, constant-velocity, zigzag, or adaptive).
+The pursuer moves using one of 9 discrete actions (8 directions + stay).
+Capture occurs when the Euclidean distance between pursuer and evader is
+less than the capture radius.
+
+Designed to accompany research on NEAT vs PPO under non-stationary opponent
```
lanctot, line 27: Interesting domain! Will import it now. 

It would be great if you can update this file to add the reference once the paper is public (in a future PR).

lanctot (approved):

open_spiel/python/games/pursuit_evasion.py
aeyjeyaryan: Sure thing! thank you so much

lanctot:
@aeyjeyaryan could you add the game to the games list? (in `docs/games.md`)?

open_spiel/python/games/pursuit_evasion.py
```diff
@@ -0,0 +1,286 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Pursuit-evasion game in 2D continuous space.
+
+A 2-player zero-sum game where a pursuer (Player 0) attempts to capture an
+evader (Player 1) within a bounded 2D grid. The evader follows one of four
```
lanctot: Since the game takes place in a continuous space and the players don't snap to discrete cells, consider changing "grid" to "box" or "space" to avoid confusion.

lanctot: Similarly elsewhere in the file (grid_size_ should be renamed)

open_spiel/python/games/pursuit_evasion.py
```diff
@@ -0,0 +1,286 @@
...
+    provides_observation_string=True,
+    provides_observation_tensor=True,
+    parameter_specification=_DEFAULT_PARAMS,
+)
+
+
+class PursuitEvasionGame(pyspiel.Game):
+  """2-player zero-sum pursuit-evasion game in continuous 2D space.
+
+  The pursuer (Player 0) takes discrete directional actions each turn. The
```
lanctot: Hardcoding the opponent's strategy into the game state and hiding stochasticity behind a dummy action breaks OpenSpiel's tree-search assumptions.

Because _EVADER_RANDOM uses self._rng, applying action 0 is stochastic. In OpenSpiel, stochastic transitions should be explicitly modeled as chance nodes so search algorithms can branch correctly. Furthermore, without a custom clone() method, duplicating this state in a search tree will share the same RNG object or fail to copy it correctly.

If this is meant to be a 2-player game, consider exposing the 9 actions to Player 1 and moving these hardcoded strategies (random, zigzag, etc.) into external pyspiel.Bot implementations. If it's meant to be a 1-player environment against a fixed behavior, Player 1's turn should be formally implemented as a chance node.

lanctot (changes requested):
Comments from internal review thanks to @dhennes

open_spiel/python/games/pursuit_evasion.py
aeyjeyaryan: Thanks for the feedback, Marc! I've addressed both points in the most recent push.

open_spiel/python/games/pursuit_evasion.py
```diff
@@ -0,0 +1,411 @@
...
+    self.evader_strategy = game.evader_strategy
+    self.pursuer_x = 0.0
+    self.pursuer_y = 0.0
+    self.evader_x = self.space_size
+    self.evader_y = self.space_size
+    self.step = 0
+    self._is_chance = False
+    self._is_terminal = False
+    self.pursuer_reward = 0.0
+    self._rng = np.random.RandomState(game.seed)
```
lanctot: There should be no random number generator in the state or the game. All randomness should be controlled externally and change events handelled by chance nodes.

lanctot: Seems to no longer be used, right? Please remove.

open_spiel/python/games/pursuit_evasion.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+class PursuitEvasionGame(pyspiel.Game):
+  """Single-player pursuit-evasion game with stochastic opponent transitions.
+
+  The pursuer (Player 0) takes discrete directional actions each turn. After
+  each pursuer action the evader moves via a chance node whose probability
+  distribution is determined by the ``evader_strategy`` parameter. The game
+  ends when the pursuer captures the evader (distance < capture_radius) or
+  after ``max_steps`` rounds.
```
lanctot: I don't understand this. 

First: why make this a single-player game instead of a two-player game?

Second: in the case where it is a single player game, then bots should not be acting at chance nodes (i.e. see the test where several different types of evaders are tested). Chance nodes have fixed distributions: you should only sample from their (categorical) distribution.

I think the better choice here is keep the evader bots and make it proper two-player game. Then use the bots when it's the evader's turn.

lanctot (changes requested):

open_spiel/python/games/pursuit_evasion.py
aeyjeyaryan: Thanks for the clear direction,  addressed all three points in the recent push!

lanctot (approved):

PR #1558, opened 2026-06-14
Add MCP Server example

lanctot:


lanctot:
Note: big credit to @dhennes who wrote the original version!

Issue #1559, opened 2026-06-20
games: Add game "Skull"


Popret:
I'm writing my Bachelors Thesis on using CFR variants to learn Skull, and would like to add my implementation.
Note: It is a commercial game (released 2011)

[Wikipedia](https://en.wikipedia.org/wiki/Skull_(card_game))
[16min Example YouTube Video](https://www.youtube.com/watch?v=rK2locyo-Pc)

It's an elegant multiplayer (>2) imperfect information game with bidding & bluffing, so it is in a similar space to Liar's Dice, but larger and with more 'agency' around what is being bluffed rather than just responding to the roll of dice.

There's already been some academic work looking at the game:
Using DQN: `Playing the Game of Skull - a Multi-Agent Deep Reinforcement Learning approach` By Maria Kapusheva 
Analyzing game-size: `A bounds-driven analysis of “Skull and Roses” cards game` By Gragera Aguaza, Baffier, Suppakitpaisarn

lanctot:
Cool, looks like a nice game.. definitely one that would be great to have in OpenSpiel.

Please take a look at point 3 in #1511. We have gotten a few PRs of games in this category lately, so I just wanted to point it out so you know ahead of time.

Do you happen to have any contact with the publisher, or know whether the authors of these past papers do?

Popret:
I don't have any contact, and don't believe previous authors have either.

I am unsure wether it makes more sense to reach out to Lui-même (the original publisher) or Asmodee ( Who acquired them in 2019 as far as I can tell ) directly. The original website still list an email to `infos@lui-meme.net` but I'm unsure if its in use.

I'm also unsure the legal ramifications but it also seems like a 'canonical' game in the sense that so many of the mechanics are so widespread, that a thematically separate but mechanically a superset/subset could be created for research purposes( See: comparing Yacht to Yahtzee )

lanctot:
> I'm also unsure the legal ramifications but it also seems like a 'canonical' game in the sense that so many of the mechanics are so widespread, that a thematically separate but mechanically a superset/subset could be created for research purposes( See: comparing Yacht to Yahtzee )

The reference given here to Wikipedia names a commercial game and that name is used in the implementation in #1560. The recommendation from our legal team has been not to import games in this circumstance. 

Is there perhaps version of the game that is clearly part of the public domain, i.e. analgously to Yacht?

PR #1560, opened 2026-06-24
Implementation of the game "Skull" from Issue #1559

Popret:
Here's a first crack at implementing #1559 , with a couple of lingering questions:

- Most importantly, I'd like some help implementing `MakeObserver()`. I'm rather new to large-codebase C++, being more used to plain C. Right now I am creating the Observers manually in the constructor, but if I do the same via MakeObserver I get heap corruption errors. Actually I get those when an implementation `MakeObserver()` exists, I think regardless of if I call it in the `SkullGame` constructor. I've tried, and [my attempt is commented out](https://github.com/Popret/open_spiel/blob/5970ed41545bc5d0df26b0ccfa1f9873025a4f3d/open_spiel/games/skulll/skull.cc#L809) right now, but any guidance here would be much appreciated.

- I'd also love feedback on the structure of `Returns()`. Especially on how best to add intermediate rewards without incentivizing some suboptimal behaviour. I'm going to do some literature review in general, to try to understand in what situations it matters. I was considering implementing a `kReturnsType` similar to how[ goofspiel currently does it](https://github.com/google-deepmind/open_spiel/blob/6c8edc829962967730e5ff353340df75847fa184/open_spiel/games/goofspiel/goofspiel.h#L80-L84), but I wanted to get a better understanding for for the reasons behind any approach before implementing any specifics.

- A note on the Tests, they could be cleaned up somewhat but before committing to any approach I had some lingering questions regarding policies. I am using preferred action policy and custom policies in `skull_test.cc` because depending on the game parameters skull can be a rather large game (e.g. many players, and perfect recall) that doesn't lend itself to a full tabular policy. 
    1. Is a use case like this a reasonable use of policies, or for situations like this are bots more appropriate? (I chose to use policies over bots because it seemed to me policies are more core to OpenSpiel?) I know bots were added after the main paper, maybe some more documentation could be of use disambiguating bots vs policies? 
    2. If this is a reasonable way to use policies could it make sense to maybe expand PreferredActionPolicy to also have optional Fallback Policies like partial_tabular, rather than failing by default?
        - How `PreferredActionPolicy` works:
        - https://github.com/google-deepmind/open_spiel/blob/6c8edc829962967730e5ff353340df75847fa184/open_spiel/policy.cc#L533-L542
        - Compare to `PartialTabularPolicy`
        - https://github.com/google-deepmind/open_spiel/blob/6c8edc829962967730e5ff353340df75847fa184/open_spiel/policy.cc#L255-L263
         - I could submit a PR if this is wanted.

Overall, I tried to be relatively thoughtful with my implementation, and tried to keep it simple. If there's any places stylistically that could be improved as well I'd welcome any feedback. 
The next steps from my end would be to properly implement `MakeObserver()`, adapt this to use the new ActionStructs instead of this flat representation, and adapt the policies / bots into their own file after hearing back what path you recommend.


Popret:
Hey, just saw the workflow was run and failed because of a goofy typo, sorry about that... Just pushed again.

Issue #1511, opened 2026-03-30
Please Read Before Submitting a Pull Request


lanctot:
Dear potential contributors,

Thank you for your interest in OpenSpiel! OpenSpiel has had the benefit of growing organically over the years due to a lot of great engagement with the community, and we're always happy to have more features, games, algorithms, bug fixes, and other improvements.

To ensure that everyone benefits from the contributions, please do the following before submitting a Pull Request:

1. **Avoid duplicating work.** Check to see that there is not already a PR in the [Pull Requests queue](https://github.com/google-deepmind/open_spiel/pulls) achieving what your pull request would. 
2. **Please talk to us before any large PRs.** There are some issues that are less important than others, and some types of contributions we may prefer to have some discussion about before you do any work. So, please check with us first. This could take the form of replying to an existing issue or opening a new issue.
3. **Please read the guidelines.** Please read over the [Developer Guide](https://github.com/google-deepmind/open_spiel/blob/master/docs/developer_guide.md) before submitting any code.

Ok, so once those are all taken into account, what happens after you've submitted a PR? 

- The GitHub Action CI tests will get triggered. Please take a look to ensure that your code passes all the tests. If the tests don't pass, please make changes until they do. 
- Someone will take a look at your code, likely ask you to make a few changes.
- Once the code passes all the tests, and is clear to be imported and merged by OpenSpiel maintainers, it will get almost certaily be merged. 

Please not that these steps can take a few weeks, even for a small PR, depending on how busy the maintainers currently are with their primary work responsibilities. Rest assured, though: if your code is cleared for import, it will quite likely be imported and merged into the master branch.. but this may take some time and may depend on the importance of the contribution (e.g. critical fixes will be prioritized).

Thanks so much for your interest in OpenSpiel and we look forward to your contributions!

Issue #979, opened 2022-12-20
How to implement games with interruptions


rezunli96:
Hi. In some card games (e.g., UNO, Mahjong), when a player plays a certain hand, then it is possible for another player to interrupt, do some action, and become the next player. Usually if multiple players interrupt it is the first one who shouted becomes the next player.

But I think it is hard to implement in OpenSpiel as it has a real-time flavor. I can see one feasible solution is to determine some pre-fixed round robin order and ask each player whether it wants to interrupt. 

Another possible solution might be to implement it as a simultaneous node. And for those who choose to interrupt just randomly decide who is the next player.

But these may not be perfectly satisfactory as they may change the vanllia game mechanic.

I wonder was there a similar situation for the existing implemented games? And what were their solutions?

lanctot:
Magic: The Gathering would be another example (but I didn't think UNO would be?)

The only way I see this being feasible is as you say: some kind of regular prompt to ask if the others want to interrupt. I like the idea of a simultaneous node to avoid the round-robin querying of the players.

I agree it's not the best solution and ideally there would be some for of support for real-time games. 


There are no games in OpenSpiel of this form, currently. Is this coming up in Mahjong? Can you give an example of a situation where it occurs that you're currently trying to figure out? Because if it's in Mahjong, that game has been used for AI before and then OpenSpiel could adopt what is commonly done in the Mahjong engines to address this.

rezunli96:
Oh for UNO I just checked [wikipedia](https://en.wikipedia.org/wiki/Uno_(card_game)), turns out yes the official rule doesn't have interruption, while there is a jump-in rule in some house-rule.

Wow that makes implementing UNO much easier than I thought! I think I am also interested in implementing UNO now, if there isn't a copy-right related issue?

In riichi Mahjong for example (I think it also applies to most of the Mahjong versions), if a player discards a tile, then the other players can call and claim this tile to form a "group" (part of the winning hand it wants to have). For certain group category like [Pon](https://en.wikipedia.org/wiki/Japanese_mahjong#Making_groups_by_calling_(melding)), I think it is the first one shouted claim the tile.

[...]

PR #1573, opened 2026-07-19
Add four-player Riichi Mahjong (single hand)

jaisinha77777:
Per #843 (see comment there flagging this before opening the PR, per #1511's guidance for larger contributions).

Adds `python_mahjong_riichi`: one hand of four-player Japanese Riichi Mahjong, Python-only.

**The calling problem.** Pon/Chi/Kan/Ron is a real-time interrupt mechanic with no direct mapping onto OpenSpiel's simultaneous-move or turn-based dynamics — there's an old, unresolved thread on exactly this at #979. This implements it as sequential, priority-ordered polling after each discard: every Ron-eligible player is asked first, in turn order (multiple simultaneous ron is supported); only if nobody rons is the single Pon/Kan-eligible player (if any) asked; only if nobody calls that is the single Chi-eligible player (the discarder's next player) asked. This reproduces standard priority (Ron > Pon/Kan > Chi) without requiring true real-time simultaneity.

**Scoring** (`mahjong_riichi_utils.py`, kept separate from game flow so it's independently unit-testable): standard hand-shape decomposition (4 sets + pair, chiitoitsu, kokushi), the standard ~25 named yaku, fu calculation (wait type, open/closed triplet-vs-kan distinction, shanpon-on-ron vs shanpon-on-tsumo, double-wind pairs), dora/aka-dora/ura-dora, and the mangan-through-yakuman point/payment tables. The common yakuman are covered (kokushi, suuankou, daisangen, shou/dai-suushii, tsuuiisou, chinroutou, ryuuiisou, suukantsu, chuuren poutou, tenhou/chiihou); a handful of rare local-rule yaku (e.g. renhou) are intentionally not included.

**Also implemented:** riichi declaration with furiten tracking (both permanent, from a player's own discards, and temporary, from declining a legal ron), chankan (robbing a kan), and the abortive draws (kyuushu kyuuhai, four-kan, four-riichi, four-wind-discard).

**Scope:** one hand per episode (deal through win or exhaustive draw), not a full multi-hand match with dealer rotation and score carryover across hands — same pattern as e.g. `leduc_poker` modeling one hand rather than a session. `returns()` is the point swing for the hand.

**Testing:**
- `mahjong_riichi_utils_test.py`: 27 tests on hand-shape decomposition, fu/yaku detection (including a regression test for a sanshoku-doujun bug caught along the way — it originally only matched runs starting at "1"), dora counting, and the payment tables, checked against hand-derived reference values.
- `mahjong_riichi_test.py`: 9 tests covering the deal structure, action-encoding consistency, a directed Tenhou-tsumo scenario, a directed ron + furiten scenario (confirming furiten blocks ron across a player's *entire* wait-set, not just the declined tile), riichi discard-locking, `pyspiel.random_sim_test` with serialization, and a zero-sum-returns check across many full random games.
- Playthrough generated and verified to round-trip via `generate_playthrough.replay`.
- Registered in `open_spiel/python/games/__init__.py`, `CMakeLists.txt`, and `pyspiel_test.py`'s expected-games list; added a row to `games.md`.

Happy to iterate on the calling-priority design in particular, since that's the part without existing precedent in the codebase.


jaisinha77777:
Closing this for now. Happy to revisit if there's interest later.

Issue #1575, opened 2026-07-23
AddGame "Atomic Chess"


SuryaSGit:
There are several existing chess implementations, but I'm excited to see the applications of RL for atomic chess. Basically atomic chess is chess but pieces explode in a 1 tile radius when captured (pawns don't explode). There's no forced win proven either (to my knowledge), so it should be an interesting pursuit. I'm excited to see what RL does with the explosion mechanic (go for high value piece captures, attack the king, or smth else entirely)! Would love your thoughts @lanctot 

https://en.wikipedia.org/wiki/Atomic_chess

lanctot:
Sounds great, would make a welcome contribution!

PR #1576, opened 2026-07-24
Add atomic chess implementation

SuryaSGit:
For #1575


SuryaSGit:
@lanctot can you trigger workflow again my bad

SuryaSGit:
am currently working on trying to use the framework to beat existing models like Fairy-Stockfish and MultiAra

PR #1579, opened 2026-07-25
add game abalone

cmarlin:
### Summary
Adds Abalone, a 2-player, deterministic, perfect-information, zero-sum board game: players push marbles off a hexagonal board, and the first to push six of the opponent's marbles off the board wins.

C++ implementation in open_spiel/games/abalone/, with Python bindings (python/pybind11/games_abalone.{h,cc}) and a docs/games.md entry.

### Parameters

> Parameter	Type	Default	Description
> marbles_to_win	int	6	Marbles to remove from the opponent to win
>board	string	"classic"	Initial layout: "classic" or "belgian-daisy"
>invert	bool	false	Swap player 1 / player 2 starting positions
>marble_advantage	bool	false	If the move limit is reached without a decisive win, the player with the most marbles wins instead of a draw
>marble_reward	double	0.10	Weight of the dense marble-advantage shaping reward

### Design notes
Dedicated alpha-beta solver. Abalone ships its own alpha-beta search (abalone_core_ab.{h,cc}) instead of the generic algorithms::minimax.h, since it operates directly on the lightweight core_state struct rather than the polymorphic State API — significantly faster given Abalone's branching factor.


cmarlin:
Relative to https://fr.wikipedia.org/wiki/Abalone_(jeu) : [Asmodée Éditions](https://fr.wikipedia.org/wiki/Asmod%C3%A9e_%C3%89ditions) is the editor

lanctot:
Thanks! 

@cmarlin there was an issue with the test (I guess a missing playthrough). Can you check out the "Adding a Game" page from the main site to see how to do it, and also update your branch from recent changes to master?

PR #1583, opened 2026-07-31
Add Game "Fox and Geese"

### LINKED ISSUE #1581: Add game "Fox and Geese"
### Game Overview

Fox and Geese is a classic asymmetric, two-player, zero-sum, perfect-information medieval board game (originating in Northern Europe) played on a cross-shaped board.

* **Player 0 (The Fox):** Moves or jumps over adjacent Geese into empty spaces to capture them. The Fox wins if it captures enough Geese so that they can no longer immobilize the Fox.
* **Player 1 (The Geese):** Geese move to adjacent empty spaces to surround and trap the Fox. Geese win if the Fox has no legal moves remaining.

### Scope

I will be focusing specifically on **Fox and Geese** as detailed in the section [Fox and geese](https://en.wikipedia.org/wiki/Fox_games#Fox_and_geese), with **one fox** and the three traditional goose counts: **13, 15, or 17**. Other variants listed in the broader article (such as *Fox and Hounds* or *Halatafl*) are out of scope for this initial implementation.

Stochastic-Batman:
Implements **Fox and Geese**, an asymmetric, zero-sum, perfect-information hunting game played on a 33-point cross-shaped board (7x7 grid with the four 2x2 corners removed). One fox tries to capture geese by jumping over them; the geese try to trap the fox so it has no legal move. The fox wins if it reduces the geese below the number needed to trap it (4); the geese win by immobilizing the fox; a player with no legal move loses; games reaching the move limit are declared a draw.

Here's the starting position (13 geese, the traditional default):
<div align="center">
image
</div>

Only the classic one-fox game is implemented, with the three traditional starting configurations (13/15/17 geese, selectable via the `num_geese` parameter). The two-fox Scandinavian variant (Halatafl/Rävspelet) and Asalto are out of scope - they have different win conditions and board rules, and are separate games rather than parameterizations of this one.

Captures are optional and may be chained (a jump keeps the turn if a further jump is available); geese in the 15/17-goose configurations are restricted to forward/sideways movement, per the historical rules that offset the extra material.

Closes #1581.

## Testing

- Full test suite passes: 293/293, including `fox_and_geese_test`, `pyspiel_test.py`, and  `api_test.py`.
- `fox_and_geese_test.cc` covers all three starting layouts, board geometry (diagonal vs. orthogonal points), optional and chained captures, goose movement restrictions, both win conditions, undo, and JSON struct round-trips (including mid-jump-chain states), plus `RandomSimTest`/`RandomSimTestWithUndo` across all configurations.
- Invalid `num_foxes`/`num_geese` values are rejected at construction with explanatory errors; verified via `pyspiel.load_game`.
- Verified interactively via `example` and `mcts_example`, including 1000-simulation MCTS self-play across all three configurations, which exercised multi-jump chains (up to 5 captures in a single turn) end to end.
- Playthrough file generated and passing via `playthrough_test.py`.


lanctot:
Thanks!

PR #1584, opened 2026-08-03
Enhance Python ISMCTSBot resampling with seeded randomness

tpemeja:
When you seed an `ISMCTSBot` with `random_state`, it should be deterministic. But the default resampler in `ismcts.py` was calling the unseeded `UniformProbabilitySampler` constructor, so it kept using wall-clock time instead of respecting the seed.

The C++ implementation was fine, Python just wasn't calling the seeded constructor.

Added a test to verify seeded bots now produce identical behavior across runs.


lanctot:
Thanks!

Issue #1358, opened 2025-08-07
Non-deterministic segfault in games_sim_test.py on Dou Dizhu


lanctot:
https://github.com/google-deepmind/open_spiel/actions/runs/16790943210

Seems to only happen when testing the wheels and only in Python 3.10. Probably a memory issue of some kind.

```
[ RUN      ] GamesSimTest.test_game_sim_dou_dizhu
  Fatal Python error: Segmentation fault
  
  Current thread 0x00007ff85e0b9a00 (most recent call first):
    File "/Users/runner/work/open_spiel/open_spiel/open_spiel/python/tests/games_sim_test.py", line 91 in apply_action_test_clone
    File "/Users/runner/work/open_spiel/open_spiel/open_spiel/python/tests/games_sim_test.py", line 174 in sim_game
    File "/Users/runner/work/open_spiel/open_spiel/open_spiel/python/tests/games_sim_test.py", line 210 in test_game_sim
    File "/private/var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/cibw-run-gc9ee5mi/cp310-macosx_x86_64/venv-test-x86_64/lib/python3.10/site-packages/absl/testing/parameterized.py", line 323 in bound_param_test
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/case.py", line 549 in _callTestMethod
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/case.py", line 591 in run
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/case.py", line 650 in __call__
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/suite.py", line 122 in run
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/suite.py", line 84 in __call__
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/suite.py", line 122 in run
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/suite.py", line 84 in __call__
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/runner.py", line 184 in run
[...]

PR #1602, opened 2026-08-31
Fix out-of-bounds write in dou_dizhu when nobody wins the auction

codewithfourtix:
If every player passes the auction no dizhu is chosen and dizhu_ stays kInvalidPlayer, but OriginalDeal() still does deal[dizhu_][cards_left_over_[i]]++, writing before the start of the array. ScoreUp() and WriteObservationTensor() already guard this case, this spot was missed.

It's reachable from ToString() on a terminal state — under -fsanitize=address,undefined that's a SEGV in FormatDeal() at dou_dizhu.cc:174, clean with the guard.

Also adds a regression test that plays out an all-pass auction and formats the terminal state. It crashes under sanitizers before the change, passes after.

Discussed in #1358. I couldn't reproduce that issue's segfault directly (macOS/py3.10 only), so I've left GamesWithKnownIssues() alone.

### DIFF
--- a/open_spiel/games/dou_dizhu/dou_dizhu.cc
+++ b/open_spiel/games/dou_dizhu/dou_dizhu.cc
@@ -160,8 +160,13 @@ DouDizhuState::OriginalDeal() const {
     deal[((i - 1 + first_player_) % kNumPlayers)]
         [CardToRank(history_[i].action - kDealingActionBase)]++;
 
-  for (int i = 0; i < kNumCardsLeftOver; ++i)
-    deal[dizhu_][cards_left_over_[i]]++;
+  // If everybody passed during the auction there is no dizhu, so the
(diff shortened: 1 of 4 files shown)


lanctot (changes requested):
Two things:

1. Can you remove `dou_dizhu` from the list returned by `GamesWithKnownIssues` in spiel.h ?  (return an empty list: `return {}`)
2. Can you take dou_dizhu out of the exemptions list(s) in `python/tests/games_sim_test.py`

sylvesterkaczmarek (commented):
Rechecked the current head after the maintainer follow-up. Guarding the leftover-card assignment on a valid dizhu fixes the OOB without changing normal deals, and the all-pass terminal path now has sanitizer-shaped regression coverage. The requested known-issue and sim-test exemptions are also removed. Looks clean to me.

lanctot (approved):

Issue #1596, opened 2026-08-29
Implement NashPG (Nash Policy Gradient): a Simple PPO-Based Nash Solver


easonyu0203:
## Summary

I'd like to contribute NashPG ([arXiv:2510.18183](https://arxiv.org/abs/2510.18183), Yu et al. 2026; accepted in TMLR 2026; I'm the lead author). NashPG is a policy-gradient method for finding Nash equilibria.

Algorithmically, it is an independent PPO learner plus a KL-regularization term toward a "magnet" policy that is periodically refreshed to the current policy. It has a convergence guarantee in two-player zero-sum (2p0s) matrix games and scales empirically on 2p0s imperfect-information games from Kuhn poker up to no-limit heads-up Texas hold'em.

## Relationship to R-NaD

NashPG targets a similar idea as R-NaD — iteratively refined regularization toward an anchor policy — but deliberately trades the machinery for simplicity:

* No V-trace, NeuRD, entropy schedule
* Just PPO + one KL term + a periodic magnet copy

Given R-NaD was removed as unmaintained (#1075, #1109, #1178), NashPG's simplicity may offer a much lower-maintenance way to keep a regularized-dynamics Nash solver in the repo. It also fits the independent-learner `AbstractAgent` model rather than a monolithic solver.

## Proposed Implementation

* `open_spiel/python/pytorch/nash_pg.py`: one `NashPG(rl_agent.AbstractAgent)` instance per player, so it runs in the standard self-play loop and works with `JointRLAgentPolicy` for exploitability, exactly like NFSP/DQN.
* Independent per-player networks and per-player magnets. (The paper shares parameters across players by relying on ego-centric observations; since OpenSpiel doesn't guarantee that, independent networks are both more general and closer to the "independent learners" framing).
* Magnet refresh:

  * With `auto_update_magnet=True` (default), each learner refreshes its magnet every `magnet_update_period` of its own updates.
  * An `update_magnet()` method is also exposed, so a user can set `auto_update_magnet=False` and drive a synchronized refresh from their training loop:

    ```python
    for agent in agents:
        agent.update_magnet()
    ```

    This allows the implementation to match the paper's synchronized procedure exactly.

## Scope

* Convergence guarantee: 2p0s matrix (normal-form) games.
* Empirically validated in the paper: 2p0s turn-based imperfect-information games.
* Runs, but with no guarantees: General-sum games; More than 2 players; simultaneous-move games.
[...]

PR #1608, opened 2026-09-02
pytorch: Implement Nash Policy Gradient (NashPG)

easonyu0203:
Implements NashPG, with the design discussed and approved in #1596. All deliverables from that issue are included: `nash_pg.py`, `nash_pg_pytorch_test.py` (registered in `python/CMakeLists.txt`, runs in < 3 s), `examples/nash_pg_example.py`, and a `docs/algorithms.md` row.

A few remarks:

- **Turn-based games with intermediate rewards.** The issue listed these as unsupported. They actually work, as long as the driver calls `step()` on every agent at every time step (not just the acting player). The example uses the standard "step only the acting player" loop; we add a commented three-line alternative for this case (search for `# For sequential game with dense rewards`).

- **Convergence.** The attached plot shows exploitability on Kuhn and Leduc poker with the shipped example defaults: Kuhn **0.041 ± 0.009**, Leduc **0.478 ± 0.066** (5 seeds). The plateau sits slightly above the paper's numbers, as expected since the paper uses a larger batch size and more updates.

image

Happy to revise or add tests. Thanks!


sylvesterkaczmarek (commented):
One edge case looks broken here: batch_size=1 is accepted, but advantages.std() uses PyTorch's default correction=1. With a single transition that returns NaN, so the normalization poisons the PPO update on the first batch. I'd either use std(correction=0)/unbiased=False here or reject batch sizes below 2. A tiny batch_size=1 regression would pin it.

PR #1616, opened 2026-09-13
Fix sweep_pi_vs_alpha crash when compute() fails early

rootkiller6788:
`sweep_pi_vs_alpha` crashes with `IndexError: index -2 is out of bounds for axis 1 with size 1` when `compute()` raises a ValueError before the sweep has recorded two stationary distributions.

That path is real: `compute()` throws when the transition matrix is reducible ("Expected 1 stationary distribution, but found 2") or blows up numerically at large alpha ("array must not contain infs or NaNs"). A warm-started sweep that gets one success then a failure used to fall into the `elif not np.allclose(pi_list[:, -1], pi_list[:, -2], ...)` branch with only one column collected, and the out-of-bounds read turned a recoverable error into a crash.

The fix tracks progress with `num_iters` and retries instead of indexing past what's been collected: no successes yet -> halve alpha and try again; fewer than two successes -> treat pi as not converged and shrink the step. Dropped the now-unused `alpharank_succeeded_once` flag.

Added two regression tests that mock `compute()` (first-call failure, and failure right after one success); both crashed before the fix and pass after. I also reproduced the failure with a concrete payoff table and a warm start at alpha=482.36.


sylvesterkaczmarek (commented):
The new `num_iters == 0` recovery has no retry bound because `num_iters` only increments after a successful `compute()`. If `compute()` keeps raising `ValueError` (for example for malformed/non-finite payoffs rather than an alpha that can be recovered by halving), this loop now halves `alpha` forever instead of surfacing the failure. Please bound failed attempts separately, or stop once alpha can no longer make progress, and add a regression for repeated failures.

rootkiller6788:
Thanks for the review — you're right, and I should have caught this myself.

I reproduced it: with a `compute()` that always raises `ValueError`, the sweep never returned and `alpha` underflowed all the way to `0.0`. The same thing happens on the other retry path (`num_iters >= 2`), where `alpha_mult_factor` converges to 1 so `alpha` effectively stops changing — I'd traded an `IndexError` for a hang in that case too.

Fixed by bounding failed attempts separately from `num_iters`. There's now a `num_failed_attempts` counter that is reset on every successful `compute()` and capped at 64; past that the original `ValueError` is surfaced as a `RuntimeError` (`raise ... from value_error`, so the cause is preserved). 64 consecutive halvings put `alpha` at `2^-64` of where it started, and on the `alpha_mult_factor` branch `alpha` has barely moved at that point, so in both cases running out of retries means the failure isn't an alpha problem.

Added two regression tests for this — one where `compute()` fails from the very first call, and one where it succeeds a couple of times and then fails permanently — both assert a `RuntimeError` and that the call count stays bounded. They hang against the previous revision and pass now.

Could you please take another look?

sylvesterkaczmarek (approved):
Rechecked b238a837. Failed compute() attempts are now bounded independently of successful iterations, reset after success, and the terminal RuntimeError preserves the original ValueError as its cause. The regressions cover permanent failure both before and after successful iterations, so the infinite-retry path I raised is closed. My concern is resolved.

